PR #22, opened 2022-04-18
Model update (pre Identity changes)

iamwillbar:


zvr:
@iamwillbar do you mind also adding the `.svg` (or `.xml` or ...) file, so that we have an editable version?

PR #27, opened 2022-08-16
[Build Profile] Add build element

lumjjb:
As part of the effort around Build Profiles, this PR introduces a "Build" element. A build is a representation of the process in which a piece of software or artifact is built. It encapsulates information related to a build process and provides an element from which relationships can be created to relate the build to the inputs/outputs, and entities related to the build (e.g. builders, identities, etc.).

The build will include relationships such as:

- `<Build> BUILT_BY[^1] <SPDX_Element>`
- `<Build> BUILD_INITIATED_BY <Subject/Actor>`
- `<Package> BUILD_INPUT_OF <Build>`
- `<Package> BUILD_OUTPUT_OF <Build>`

[^1]: For BUILT_BY:At the moment it can point to an artifact to signify a builder, but can point to a "Provider/Service" element in the future which can encapsulate more details of the builder if needed.

Signed-off-by: Brandon Lum <lumjjb@gmail.com>


model/Core/Classes/Build.md
```diff
@@ -0,0 +1,74 @@
...
+- name: Build
+- SubclassOf: Element
+- Instantiability: Concrete
+
+## Properties
+
+- buildType
+  - type: xsd:string
+  - minCount: 0
+- configSoureEntrypoint
```
iamwillbar, line 45:
```suggestion
- configSourceEntrypoint
```

model/Core/Classes/Build.md
```diff
@@ -0,0 +1,74 @@
...
+  - type: xsd:string
+  - minCount: 0
+- configSoureUri
```
iamwillbar, line 48:
```suggestion
- configSourceUri
```

iamwillbar:
```suggestion
- configSourceUri
```

iamwillbar, line 48: Did you consider representing the config source as a relationship to a package? `Build HAS_CONFIGURATION Package`. You could then, optionally, capture information about all of the files in the configuration and their hashes.

model/Core/Classes/Build.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Metadata
+
+- name: Build
+- SubclassOf: Element
+- Instantiability: Concrete
+
+## Properties
+
+- buildType
```
iamwillbar, line 42: Could you give some examples of what a `buildType` might be?

model/Core/Classes/Build.md
```diff
@@ -0,0 +1,74 @@
...
+- configSoureUri
+  - type: xsd:string
+  - minCount: 0
+- configSoureDigest
+  - type: xsd:string
+  - minCount: 0
+- parameters
+  - type: xsd:map<string>string
+  - minCount: 0
+- buildInvocationId
```
iamwillbar, line 57: Is this intended to be a unique identifier for the build instance? If so, since `Build` inherits from `Element` could this be stored in the `ExternalIdentifiers`?

model/Core/Classes/Build.md
```diff
@@ -0,0 +1,74 @@
...
+  - maxCount:1
+- startTime
+  - type: xsd:DateTime
+  - minCount:0
+  - maxCount:1
+- endTime
+  - type: xsd:DateTime
+  - minCount:0
+  - maxCount:1
+- logReference
```
iamwillbar, line 69: `Build` already has an `externalReferences` property that you could put a reference in, however, for logs I would probably lean towards this being a reference to a file, that allows you to capture the hashes so you can verify the log hasn't been tampered with.

zvr (commented):
Why is this being added to the `Core` profile, instead of the `Build` one ?

model/Core/Classes/Build.md
seabass-labrax: From my understanding of the discussions in recent Builds SIG meetings, you're spot on, @iamwillbar. I would also agree that `ExternalIdentifiers` is the right place to put this, given that the cardinality of the property is listed as 0-1 (that is, it's an optional field). However, this would mean that we would need to add new ExternalIdentifier types to support different build systems' ID schemes.

model/Core/Classes/Build.md
seabass-labrax: My gut reaction is to have a relationship (eg. `LOG_OF`) for this, since there might be multiple log files or multiple log formats. Again, you can communicate hashes in the relationship's target Element.

seabass-labrax (changes requested):

model/Core/Classes/Build.md
seabass-labrax: From my understanding of the discussions in recent Builds SIG meetings, you're spot on, @iamwillbar. I would also agree that ExternalIdentifiers is the right place to put this, given that the cardinality of the property is listed as 0-1 (that is, it's an optional field). However, this would mean that we would need to add new ExternalIdentifier types to support different build systems' ID schemes.

model/Core/Classes/Build.md
seabass-labrax: My gut reaction is to have a relationship (eg. LOG_OF) for this, since there might be multiple log files or multiple log formats. Again, you can communicate hashes in the relationship's target Element.

model/Core/Classes/Build.md
lumjjb: A URI indicating what type of build was performed. It determines the meaning of how to interpret the inputs of the build - e.g. how to handle the entry point or how to carry out the build

model/Core/Classes/Build.md
lumjjb: quick question: if we want to point a log URI, what would be the best way to represent that? Like if the build produces a pointer to a URI that the logs will be accessible at (but it may not be immediately accessible or only accessible to authorized individuals)

lumjjb:
Happy to move this wherever will be best, would that be creating a new directory called `Build`? or as part of `Software`?

model/Core/Classes/Build.md
lumjjb: hmm, I like that.. but have a question on how would we be able to relate the entrypoint to the particular config, that seems to be something that is expressed separately as builds.

lumjjb: fixed!

model/Core/Classes/Build.md
lumjjb: fixed!

zvr:
Yes, please create a new directory for the build profile -- `Build` is the appropriate name.
Inside the directory, have directories for `Classes`, `Properties` and `Vocabularies`, like the `Core` and `Software` profiles.

lumjjb (commented):
Added comments from build meeting on 22nd August 2022

lumjjb:
Made some changes based on the comments!

model/Build/Classes/Build.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Metadata
+
+- name: Build
+- SubclassOf: Core:Element
+- Instantiability: Concrete
+
+## Properties
+
+- buildType
```
lumjjb, line 56: comments on build type being a manual build, build type specifying how to map a set of inputs and configuration to execute a build.. 

Would the list of commands run be part of the configSource (expressed as a file)? or part of the build inputs?

model/Core/Classes/Build.md
lumjjb: we had a chat about this during today's build profile meeting. I think configSource can be a file or package (1:1 relationship), and then entrypoint would be a separate field still.

lumjjb (commented):
Add meeting discussion from 29 August 2022 meeting

lumjjb:
Closing in favor of #79

Issue #8, opened 2021-12-07
Punch List: Hash


iamwillbar:
This is a punch list of open questions from the 2021-12-07 Tech Team meeting. Please comment on this issue with any discussion, proposed answers, or additional questions you have:

- [ ] Do we have all the hash algorithms we want?
- [x] Should hash value be string?  (should serializations supporting binary data be optimized?)
- [x] Do we want to include interpretations in the strings in the logical model?  
- [x] What are the serialization rules? (keeping it distinct from serialization - byte level)

davaya:
Hash value is binary data.

Hash value can be serialized as a string (upper or lower case hex, with or without whitespace, base64, url-safe base64, etc.), all of which correspond to the same binary value.

This is equivalent to date/time being a single numeric value which can be serialized in many string formats, and IPv4 / IPv6 addresses which are 32 and 128-bit binary values that can be serialized in many text representations.

seabass-labrax:
I agree with @davaya. Further to his comments, I would suggest that the individual serialisation specifications should mandate a certain representation for hashes, which would facilitate consistent round-tripping for ingestion and export of these formats. For instance:

> In JSON format, SHA256 hashes should be serialised as a string of 64 characters, containing a lower-case base 32 representation of the binary hash value.
> Example: `"hashValue": "499346e3caca25705e1a0e358bc89fe411c5b6345a644c308ef2cf2c0ee8c230",`

There's a potential discussion point as to whether this would be too constraining, but on balance I think the round-trip advantage would outweigh any stylistic concerns.

davaya:
I agree.

The OpenC2 information model (https://docs.oasis-open.org/openc2/oc2ls/v1.0/cs02/oc2ls-v1.0-cs02.html#315-serialization) uses a "format" option to require a particular string representation.  By default binary values are required to be serialized as URL-safe base64, but using the "x" option (Binary /x) requires the string to be upper-case hex with no whitespace as specified in RFC 4648.  We're planning to update that so that upper-case /X requires RFC 4648 encoding and validation, while lower-case /x uses lower-case hex encoding and case-insensitive validation.

Any text-based serialization would use the same formats, and additional formats can be defined.  For example MAC address - Binary /eui - is required to be a 48 or 64 bit binary value represented as colon-separated hex bytes: 00:1B:44:11:3A:B7.  (This should also be relaxed - the IEEE spec allows no whitespace or hyphens between hex bytes.)

[...]

PR #28, opened 2022-08-23
Use binary for type of `hashValue` property

seabass-labrax:
This commit changes the type of the `hashValue` property of the non-element Hash class shown in the model diagram to read 'Binary' rather than 'String'. This is consistent with decisions made during the SPDX Tech Team meeting on 2022-01-04 and in the GitHub issue thread available at <https://github.com/spdx/spdx-3-model/issues/8>.

Signed-off-by: Sebastian Crane <seabass-labrax@gmx.com>


maxhbr:
LGTM ;)

iamwillbar:
I've made this change in the model I'll be uploading later today. Closing this to avoid conflicts.

PR #31, opened 2022-10-24
Drop version from ProfileIdentifier

maxhbr:
As decided in [`SPDX Tech Team meeting - 4 October, 2022`](https://spdx.swinslow.net/p/spdx-tech-minutes):
```
* Profile version discussion:
    * Allows for different profiles to maintain different versions - would allow for asynch release
    * Downside - adds complexity to validation
    * Immutable elements will be under a profile and once created, the version can't change
    * Strong concerns expressed with managing multiple versions in the tooling - similar to dependency hell in software development
    * Proposal 1 - Profiles will have versions in the logical model, but for the forseeable future, profile versions will be in sync with the spec version
    * Proposal 2 - Remove the version from profiles, profiles will always be released in lock step with the SPDX spec.  Implication, changing this would be a breaking change - so we are essentially deciding that profiles and specs will be release together until at least 4.0 (which allows breaking changes)
    * Decision: Proposal 2 - remove version from profile
```

This **drops the version** and slightly adopts the examples to be actually potentially extensible in the future:
```diff
-    "profile": ["core"],
+    "profile": [{ "name": "core"}],
```


maxhbr:
this is outdated and was changed independent of this PR

PR #45, opened 2022-11-08
Add glossary of terms

nishakm:
Signed-off-by: Nisha Kumar <nisha.kumar@oracle.com>

### DIFF
--- a/model/Glossary.md
+++ b/model/Glossary.md
@@ -1,0 +1,17 @@
+# Glossary of Terms
+
+| Term  | Description |
+| ------------- | ------------- |
+| Data type | |
+| Class | A blueprint for creating an object |
+| Property | A field within a class |
+| Domain | An area of concern. For example, licensing, security, usage, etc |
+| Namespace  | A portion of the model within which a class is defined |
+| Constraint | A statement in a profile, restricting the set of allowed values, e.g. for properties in the model. Examples could be the requirement of a different hash algorithm to be present. |
+| Core | The namespace which contains classes common to all domains |
+| Profile | A namespace which contains classes addressing a particular domain |
+| Validation | |
+| Shape | |
+| Canonicalization | A process that takes abstract model data and generates a deterministic (string) representation (Serialization), normalizing all choices like ordering, formatting and representation choices (like direction of relations and inlining). See also https://spdx.github.io/canonical-serialisation/ |
+| Serialization | A representation of data as a byte stream |
+


nishakm:
See https://github.com/nishakm/spdx-3-model/blob/glossary/model/Glossary.md for a markdown rendering

nishakm:
cc: @goneall @zvr @davaya
I don't know sbarnum's github username

nishakm:
Also @seabass-labrax :)

zvr:
> See https://github.com/nishakm/spdx-3-model/blob/glossary/model/Glossary.md for a markdown rendering

On this PR, one can go to the "Files changed" tab and click on the icon on the right-hand side above the code: the `<>` icon shows markdown code, the "page (sheet of paper)" icon shows rendered text.

zvr:
I propose that the file (`Glossary.md`) is on the top-level, not inside `model`.

davaya:
For the first item I suggest using the [UML](https://www.omg.org/spec/UML/2.5.1/PDF) definition verbatim:

> _DataTypes model Types whose instances are distinguished only by their value._

(Section 10.2.1).  But as discussed in the meeting we may need concrete examples, not just glossary definitions, to get to common understanding.  #44 has some suggested examples.

goneall:
As discussed in the call, below is a proposed definition of `Union`:

The term Union is defined as a [Set Union](https://en.wikipedia.org/wiki/Union_(set_theory)) where the union of a collection of Sets is the set of all elements in the collection[1].

As it applies to the recent SPDX discussion on properties and enum values, the sets being referred to are each set of properties or enums belonging to a class referenced in each profile and the class is defined in the core profile.

[1] Weisstein, Eric W. "Union". Wolfram's Mathworld. Archived from the original on 2009-02-07. Retrieved 2009-07-14.

goneall:
Based on 28 Mar 2023 tech call, Sean will review and comment on the terms

sbarnum:
I propose these new definitions for the outlined terms as well as several additional relevant terms.
I believe that these definitions are accurate and appropriate for the goals and approach of SPDX 3.0.
They also intentionally include terms and definitions aimed at reducing disconnect on a few areas of historical conflation/confusion including the difference between system design and language design (what we are actually doing) as well as the distinct meanings of namespaces and profiles in our approach.
I will drop a note to the spdx-tech email list to let people know that this note has been added.


# Glossary of Terms

| Term  | Description |
| ------------- | ------------- |
| Domain | A specific sphere of activity or knowledge. For example, cyber security, software, bills of material, licensing, software security, etc |
| Ontology | An explicit specification of concepts, terms, categories, relationships, etc. for a particular scoped domain forming a language or basis for communication and analysis of information within the scoped domain. SPDX 3.0 is an ontology targeted to support the scoped domain of bills of material.|
| Class | A represention of a scope/set of individual instances of a particular “concept” (e.g., File, Person, ExternalReference, etc.). Each individual instance of a class has an IRI ID and is asserted as a member of a particular class via a type statement. |
| Domain Class | A class representing a concept of primary focus within a given domain. In a graph or serialized set of instance content this would be the granularity of what would be a node in the graph or a top-level object in the serialization set. This is typically similar to the scoping seen in labeled property graphs. In SPDX 3.0, all domain classes are subclasses of Element.|
| Non-domain Class | A class representing a concept of secondary or tertiary focus within a given domain that is a characterization of some aspect of a domain class. When structuring the graph for a domain use case it is desirable and more practical for the domain graph nodes to only consist of domain class instances and object properties. Any non-domain classes and datatype properties would be considered internal adornments on the domain objects. In SPDX 3.0, all defined classes that are not subclasses of Element are non-domain classes.|
| Property | A representation of a relationship between a class (definition or instance) and some descriptive characterization of the class. |
| Datatype property | A representation of a relationship between a class (definition or instance) and some literal (string, integer, boolean, etc.) descriptive characterization of the class. |
| Object property | A representation of a relationship between a class (definition or instance) and some conceptual (class-based) descriptive characterization of the class. In other words, it is a relationship between one class and another class. |
| Domain | The class at the root of a property relationship. In other words, the class being characterized. |
| Range | The class or literal datatype of the value of a property relationship. |
| Namespace  | An explicit scoping within which concept classes and properties for a given domain are defined and referenced. Each namespace typically defines its own IRI prefix utilized in the IRI IDs of all concept classes and properties defined within it. In SPDX 3.0, the "core" namespace is the scoping within which all foundational BOM concept classes and properties are defined and managed. |
| Core | The namespace which contains definitions and constraints for all concpet classes and properties which are common to all other domains within the targeted scope of SPDX. |
| Constraint | An explicit conformance requirement specifying details of an SPDX-valid relationship between a property and a concept class including such details as the valid type of the property on that class, the cardinality (e.g., 0..1, 1, 1..*, etc.) of the property on that class, the value of the property on that class, etc. In SPDX 3.0 these constraints are expressed using the [W3C SHACL language](https://www.w3.org/TR/shacl/). One example could be the requirement of a specific hash algorithm to be present. |
| Shape | A formally defined set of constraints that explicitly specify the valid "shape" of particular data to support validation against SPDX conformance requirements. A [Property Shape](https://www.w3.org/TR/shacl/#property-shapes) defines the set of constraints for a given property when applied to a particular concept class. A [Node Shape](https://www.w3.org/TR/shacl/#node-shapes) defines the set of property shapes (and thus constraints) relevant for validation of instance data realization of a given concept class. In SPDX 3.0 these shapes are expressed using the [W3C SHACL language](https://www.w3.org/TR/shacl/).|
| Validation | The action of checking or proving that instance data content conforms to the explicit semantics (meaning) and syntax (structure of classes and properties) of the SPDX language standard. In SPDX 3.0, automated validation of RDF-based serialized content can be accomplished with various freely available SHACL validation engines such as pyshacl. Custom validation engines could also implement automated validation of non-RDF-based serializations against the defined SHACL shapes within SPDX 3.0.|
| Profile | An explicit scope of usage for SPDX targeting support for particular use cases and scenarios (e.g., core, software, licensing, build, etc.). A profile identifies which particular SPDX namespaces, concept classes and properties it leverages along with any custom specializations of shape constraints unique to its use. NOTE: A profile is NOT the same thing as a namespace. A namespace is a formal part of the language ontology and defines specific concept classes, properties and shape constraints unique for a particular domain scope. A profile represents usage and conformance documentation for the language ontology and references concept classes, properties and shape constraints defined within namespaces. For a given profile, there may exist a namespace closely aligned to the same usage scope but they are not the same thing and the profile will often leverage/reference concept classes, properties and shape constraints from more namespaces (e.g., core) than just the aligned namespace.|
| Serialization | The process of translating an abstract logical data structure into a format that can be stored or transmitted and reconstructed later. This series of resulting bits, processed according to its particular serialization format rules, can then be used to create a semantically identical clone of the original content through the process of deserialization. |
| Canonicalization | A process that takes data in any valid form (e.g., various serializations of SPDX like json-ld, yaml, xml, turtle, etc.) and transforms it into a single consistent normalized deterministic and reproducible form. Such a canonical form normalizes things like ordering, formatting, choices, etc. See also https://spdx.github.io/canonical-serialisation/ |

davaya:
The proposed definition of "Class" is incomplete.  A Class defines the set of data instances that have common structure and value constraints.  (This is a subset of object oriented classes, which have behavior/methods in addition to data constraints.)

Every item on the model diagram is a Class (or in UML terms, a "classifier") that determines whether a data instance is a member of the Class.
* A string instance is a member of the SemVer Class if it matches the described constraints
* A complex instance is  a member of the PositiveIntegerRange Class if it has start and end properties that are each members of the PositiveInteger Class (not currently shown)
* A string instance is a member of the SoftwarePurpose Class if it has one of the values listed in the enumeration.
* A complex instance is a member of a referenceable Class if its properties include a unique id (or in database terms, a primary key) that can be used as a reference (a foreign key).

Every Class is either a datatype (if it is identified by its value) or referenceable (if it has a unique key/id).  The SPDXv3 model has only one referenceable class (Element and the classes extended from it), but other models can have diverse referenceable classes (e.g., vehicles that are indexed by VIN number and have no IRI, or classes in an ontology that do not extend SPDXv3 Element).

davaya:
"Class" has two common meanings:
1) an object-oriented programming Class meaning a procedure that has attributes and methods
2) a generic class (or in UML terms, a "classifier") meaning characteristics that a set of instances have in common, for example "mammals" is a class of living things

The proposed Class definition is from object oriented programming.  But we have said that the SPDXv3 logical model is not an OOP model.

If Class has the generic meaning:
- A string instance is a member of the SemVer Class if it matches the described constraints
- A complex instance is a member of the PositiveIntegerRange Class if it has start and end properties that are each members of the PositiveInteger Class (not currently shown, should be NonNegativeInteger or WholeNumber if zero is intended to be included in the class)
- A string instance is a member of the SoftwarePurpose Class if it has one of the values listed in the enumeration.
- A complex instance is a member of a referenceable Class if its properties include a unique id (or in database terms, a primary key) that can be used as a reference (a foreign key).

Decision:  In the logical model, does Class have the OOP meaning (only Elements are classes) or the generic meaning (everything in the model (Elements, Simple data types, Complex data types, and Enumerations) are Classes)?

davaya:
> I’m OK with “Domain Class” always having an id.
> I’m hesitant to accept that “Domain Class” is always inheriting from Element.

I agree. I haven't heard the terms "Domain" and "Non-Domain" classes used before, but the common language meaning would be that if SPDX is the domain, then Element (with IRI id) and License (with LicenseId) would be domain classes and Vehicles (with VIN ids) would not be in the SPDX domain.

I won't object to defining Domain classes to mean those with an ID, but because to me "domain" implies "SPDX" or some other administrative boundary, I made up the term "referenceable classes" to mean those with an ID regardless of what domain they are defined in.  In that Venn diagram, all classes, regardless of where they are defined, are either referenceable (with an ID) or datatype (without any ID smaller than the entire value).

goneall:
> “Class” is the general term
> “Domain Class” is what we use as “Class” on the diagram (left side)
> “Non-Domain Class” is what we usually refer to as “complex data types” in the diagram (right side)

I'm OK with these definitions.  I strongly support having 3 different terms, one general, one that requires an IRI (left side), and one that can be a blank node / anonymous or an IRI (right side).

kestewart (approved):
Based on the comments in the thread, I think we have some degree of agreement.   Moving this into the repo, and we can iterate from there.

zvr:
@kestewart I don't think this should have been merged, since the general agreement was about what @sbarnum put in his comment, not what was in the PR...

zvr:
I've created #236 to add Sean's content.

PR #50, opened 2022-11-30
name case, metadata bullets

davaya:
Initial Cleanup
- Update filename case for consistency: types are capitalized, properties are lower case
- Annotation subject is 1..1
- BOM and SBOM classes are all-caps
- Add missing bullets to metadata
- externalID should be externalId


zvr:
Thanks for this, David.

-  I thought we'd have all the vocabularies with initial lowercase. (Classes are initial uppercase, properties are initial lowercase as you say)
- I wrote BOM etc. with initial uppercase only on purpose. Why should we have them all caps?
- The rest of the changes are completely correct.

davaya:
If we change BOM and SBOM on the diagram then Bom would be fine.  As long as the model and diagram agree.

Vocabularies are types - properties have vocabulary values.

model/Software/Classes/Sbom.md
```diff
@@ -1,15 +1,18 @@
 SPDX-License-Identifier: Community-Spec-1.0
 
-# File
+# SBOM
```
maxhbr, line 3: I thought we decided for `Sbom` as a class name

zvr:
I think I have incorporated all of these in #60.

model/Software/Classes/Sbom.md
maxhbr: Ah, the heading does not match the `name` of the class? Feels inconsistent and prone to introduce issues

PR #57, opened 2023-01-02
Serialization description and examples

davaya:
Adds a "serialization" folder to the model that describes the serialization process and contains JSON examples derived from the current model diagram.

See the full description viewable at https://github.com/davaya/spdx-3-model/tree/serialization/serialization.


nishakm (commented):
It's a little confusing for me to understand without documentation :). A roadmap of where things should be would be nice.

serialization/Elements/sbom.json
```diff
@@ -0,0 +1,30 @@
+{
+  "id": "urn:spdx.dev:null-sbom",
+  "type": {
+    "sbom": {
+      "element": [
+        "urn:spdx.dev:iamwillbar",
+        "urn:spdx.dev:spdx-tools-3.0.1",
+        "urn:spdx.dev:project",
+        "urn:spdx.dev:doc"
+      ],
```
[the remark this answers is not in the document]
davaya: Yes, externalReference assigns an SPDX ID to things for which there are no Elements.

serialization/Elements/sbom.json
```diff
@@ -0,0 +1,30 @@
...
+      "import": [
```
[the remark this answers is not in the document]
davaya: The model says Element has an externalReference property of type ExternalReference.  I have no clue where I got "import" from - thanks, I'll fix it.

serialization/Elements/sdoc.json
```diff
@@ -0,0 +1,21 @@
+{
```
maxhbr, line 1: for an SPDX document, the `name` is mandatory and is missing here

davaya: Good catch, thanks.

serialization/Payloads/sdoc-v1.0.json
```diff
@@ -0,0 +1,97 @@
...
+      "verifiedUsing": [
+        {
+          "hash": {
+            "sha256": "14a657a7118a333cc1fdc6af05071a59cda067fd11130d4ee5d6d47c26e7863f"
+          }
+        }
+      ],
+      "externalIdentifier": [
+        {"type": "purl", "identifier": ""},
+        {"type": "cpe_2.2", "identifier": ""}
```
armintaenzertng, line 70: I'm curious as to why you handle properties inherited from `Element` differently than the package specific properties (only the latter are grouped into the "type":"package" sub-dict). Is it because the schema is easier that way?

serialization/Payloads/sdoc-v1.0.json
```diff
@@ -0,0 +1,97 @@
+{
```
armintaenzertng, line 1: This payload example does not validate against the schema (even after adding the "payload" key on top). It would be nice to have an example that does.

Most of the problems are related to lists like "element" listing only the elements' Ids, while the schema expects the full elements.

serialization/Schemas/spdx-v3.json
```diff
@@ -0,0 +1,1061 @@
...
+  "description": "This information model defines the serialized structure of SPDXv3 logical model types",
+  "license": "CC0-1.0",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "payload": {
+      "$ref": "#/definitions/Payload"
+    },
+    "elements": {
+      "$ref": "#/definitions/Elements"
```
armintaenzertng, line 15: As payload already has a list of all elements, what is the idea of the "elements" list parallel to "payload"?

serialization/Schemas/spdx-v3.json
```diff
@@ -0,0 +1,1061 @@
...
+      "minItems": 1,
+      "maxItems": 100,
+      "items": {
+        "$ref": "#/definitions/Element"
+      }
+    },
+    "NamespaceMap": {
+      "title": "NamespaceMap",
+      "type": "object",
+      "additionalProperties": false,
```
armintaenzertng, line 103: `NamespaceMap` has the two properties `prefix` and `namespace` according to the model.png

serialization/Schemas/spdx-v3.json
```diff
@@ -0,0 +1,1061 @@
...
+        },
+        "type": {
+          "$ref": "#/definitions/ElementType",
+          "description": "Element type and type-specific properties"
+        },
+        "created": {
+          "$ref": "#/definitions/DateTime",
+          "description": "required: explicit or default"
+        },
+        "creator": {
```
armintaenzertng, line 129: this is called `createdBy` in the model

serialization/Schemas/spdx-v3.json
```diff
@@ -0,0 +1,1061 @@
...
+      ],
+      "maxProperties": 100,
+      "properties": {
+        "namespace": {
+          "$ref": "#/definitions/IRI",
+          "description": "Default namespace for Element ids in the SPDX file (rdf BASE)"
+        },
+        "prefixes": {
+          "$ref": "#/definitions/NamespaceMap",
+          "description": "Namespace abbreviations used in the SPDX file (rdf PREFIX)"
```
armintaenzertng, line 423: In the model, `namespace` is used for an object of type `NamespaceMap`, which here is called `prefixes`. I would prefer to keep the naming consistent with the model.

serialization/Schemas/spdx-v3.json
```diff
@@ -0,0 +1,1061 @@
...
+          "description": "IRI assignment to non-element hashable data entities",
+          "minItems": 1,
+          "items": {
+            "$ref": "#/definitions/ExternalMap",
+            "description": "IRI assignment to non-element hashable data entities"
+          }
+        }
+      }
+    },
+    "Bom": {
```
armintaenzertng, line 458: "Bom" and "Sbom" in the schema miss the properties `namespace` and `import` from the model.

"Bom", "Sbom" and "SpdxDocument" in the schema miss the property `context` from the model.

serialization/Schemas/spdx-v3.json
```diff
@@ -0,0 +1,1061 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "element"
+      ],
+      "maxProperties": 100,
+      "properties": {
+        "element": {
+          "type": "array",
+          "description": "<-Collection",
```
armintaenzertng, line 595: What do you mean by the "<-" before "Collection"? I assumed you are referring to the description of a Collection in the schema, but there is none.

armintaenzertng (commented):
Thanks for the effort you put in here! :)
While trying to implement your serialization suggestion I came across a few issues, see below.

serialization/Payloads/sdoc-v1.0.json
davaya: Grouping the package-specific properties in the "package" property keeps them together and guarantees that there are no property name collisions without having to prepend "package" to individual properties.   I'm not sure that there would be a "purpose" property and a "downloadLocation" property elsewhere in a Package element, but for some reason the logical model qualifies one and not the other.  

Grouping them also (imo) makes element data easier to read by keeping them together, and syntax editors can collapse "type" to get a compact overview of many elements.

As you say it also makes the schema easier to write manually - the Package type defines only package properties.  It's also possible for a schema generation tool to run the inheritance chain for each element type and generate a fully-populated Package type, but that means the human-created input schema is different from the inheritance-resolved output schema.

serialization/Payloads/sdoc-v1.0.json
davaya: Hmm - I'll investigate, I hadn't run into errors ...

The "element" property in all types (e.g. Sbom/element) must have ids because elements reference other elements.  The "element" property in Payload contains the full value for all elements in that payload.

serialization/Schemas/spdx-v3.json
davaya: Elements is an array of uncompressed element values.
Payload is the structure of context information plus the "element" property of compressed element values.

It would probably be clearer to have two different schemas.   Thanks.

serialization/Schemas/spdx-v3.json
davaya: That's a bug in my tooling - thanks for catching that!

The source in `serialization/Schemas/spdx-v3.jidl` is:
```
NamespaceMap = MapOf(Prefix, IRI){1..*}
```
which means that every property name must be a Prefix (without explicitly enumerating them) and every value is an IRI.  Now I'm thinking your validation errors mentioned earlier are also errors in my JSON Schema generation tool.

serialization/Schemas/spdx-v3.json
davaya: I was considering proposing a name change to the model, but will fix the schema to agree with the current model.

serialization/Schemas/spdx-v3.json
davaya: The model does not include the payload namespace.  XML supports both a document default namespace and prefixes to shorten references to other namespaces.  For example, from https://www.w3.org/TR/turtle/#sec-intro:
```
@base <http://example.org/> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
```
I can change `namespace` to either `base` or `baseNamespace`, but either way it needs to be added to the model.  Do you have a preference?

serialization/Schemas/spdx-v3.json
davaya: The model confuses Bom and Sbom (which are just elements like all other elements) with SpdxDocument (which is special in that it is information about the Payload).  This serialization is a proof-of-concept for demonstrating that the current model is unnecessarily complex: there is no need for an abstract ElementCollection or concrete Bundle, because Payload (which didn't exist when ElementCollection was created) replaces them both.

Only SpdxDocument needs `namespace`.  It could include an optional "context" string if there is still a need for that after simplifying the model.

serialization/Schemas/spdx-v3.json
davaya: The comments in ElementType show class inheritance.  Collection was a class in some previous version of the model, but with the proposed simplification SpdxDocument would be the only class inheriting from the abstract ElementCollection, or ElementCollection would disappear entirely if there are no use cases for something other than a Payload being a collection of Element values.

goneall:
From discussion on the 28 Mar 2023 tech call, we will review and merge this after the Serialization team finishes it's proposal and aligns on the content of this PR

PR #60, opened 2023-01-10
Change naming

zvr:
Changes names according to tech team decisions,
as documented in Tech Meeting 2022-12-06.


maxhbr (approved):
LGTM

kestewart (approved):
Approving this, so we can build off it, but Actor Type has been removed, and needs to be aligned with model that was checked in from 230110 meeting.

PR #63, opened 2023-01-11
Update description of Package to match SPDX 2.3

kestewart:
Added text from SPDX 2.3 - Section 5.2.2 to fill in Package Description.

-Signed-off-by:  Kate Stewart <kate.stewart@att.net>

### DIFF
--- a/model/Software/Classes/Package.md
+++ b/model/Software/Classes/Package.md
@@ -4,7 +4,24 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Description
 
-TODO
+If SPDX information is being used to describe packages, then one instance of the package information per package being described shall exist. It provides important meta information about the package as a whole. Packages are an abstract concept that can be used to refer to any distribution of software, typically consisting of one or more files and capable of containing sub-packages. Starting with SPDX 2.0, it is not necessary to have a package wrapping a set of files.
+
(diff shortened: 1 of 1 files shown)


model/Software/Classes/Package.md
```diff
@@ -4,7 +4,24 @@ SPDX-License-Identifier: Community-Spec-1.0
...
+Any of the following non-limiting examples may be (but are not required to be) represented in SPDX as a package:
+
+- a tarball, zip file or other archive
+- a directory or sub-directory
+- a separately distributed piece of software which another Package or File uses or depends upon (e.g., a Python package, a Go module, ...)
+- a container image, and/or each image layer within a container image
+- a collection of one or more sub-packages
+- a Git repository snapshot from a particular point in time
+
+Note that some of these could be represented in SPDX as a file as well.
```
goneall, line 20: Suggest adding something like "If the intention of the above examples is to be distributed, including package metadata is strongly encouraged".

goneall (changes requested):
Overall - looks good - I just had one comment to further reinforce when package metadata should be used.  I'm not really sure it is necessary, but as I was reading it I was thinking someone may miss-interpret the "can also be represented as a file" the the package metadata was more optional.

goneall:
Overridden by PR #122

PR #69, opened 2023-01-19
add SoftwareDependencyRelationship and LifecycleScopedRelationship

armintaenzertng:
also add the three software dependency vocabularies `DependencyLinkType`, `DependencyRequirement` and `DependencyScope` according to https://github.com/spdx/spdx-3-model/blob/main/model.png

### DIFF
--- a/model/Core/Classes/LifecycleScopedRelationship.md
+++ b/model/Core/Classes/LifecycleScopedRelationship.md
@@ -1,0 +1,21 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# LifecycleScopedRelationship
+
+## Description
+
+TODO
+
+## Metadata
+
+- name: LifecycleScopedRelationship
+- SubclassOf: Relationship
+- Instantiability: Concrete
+
+## Properties
+
+- scope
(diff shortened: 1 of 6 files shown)


maxhbr:
Kate wants to give feedback to the SoftwareDependencyRelationship concept as it is expressed in the drawio diagram and this is blocking this PR.

kestewart:
@iamwillbar - can you please plan on making next Tuesday's call, as I think it will take both of us to work through this.
I think there are some use cases the revised model will have trouble figuring out how to specify, so need to talk this out.

kestewart (changes requested):
Some concerns about the underlying model.   Needs further discussion.

armintaenzertng:
@kestewart, I updated the PR according to yesterday's tech team discussion. I'm not sure whether `DependencyConditionalityType` belongs in `Software` or if it is a general enough concept that it would be reused in other profiles.

kestewart (approved):
We need to fill in the TODOs, but this relects my understanding of the decisions made in the tech team meeting.   Thanks!

PR #72, opened 2023-01-23
Add `comment` property in CreationInformation.md

armintaenzertng:
As seen in the model.png.
This field will be needed to convert the SPDX2 property `CreatorComment` to SPDX3.


armintaenzertng:
ping: @zvr

PR #73, opened 2023-01-23
add packageVersion property in Package.md

armintaenzertng:
As seen in the model.png.
This field will be needed to convert the SPDX2 property `PackageVersion` to SPDX3.


armintaenzertng:
Ah, of course, sorry I forgot about that!
Added now

goneall:

PR #78, opened 2023-02-08
Tech Meeting 2023-02-07

iamwillbar:
* Split `createdBy` into `createdBy`  (`Identity[1..*]`) and `createdUsing` (`Tool[0..*]`).
* Removed the `Entity` layer.
  * `Tool` and `Identity` now inherit directly from `Element`.
* Changed the type of `originatedBy` to `Identity[1..*]`.
* Added `name` explicitly to `File` and `Snippet` to override the cardinality inherited from `Element` to `1..1` (this was from last week).


goneall (approved):
Looks good - Just a note that the Identity is still under discussion.

seabass-labrax (approved):
+1 from me :smiley:

PR #91, opened 2023-02-28
Add Build profile

lumjjb:
Add build profile definitions to spdx-model-3


model/Build/Classes/Build.md
```diff
@@ -0,0 +1,57 @@
...
+  - minCount: 0
+- parameters
+  - type: xsd:map<string>string
+  - minCount: 0
+- startTime
+  - type: xsd:DateTime
+  - minCount:0
+  - maxCount:1
+- endTime
+  - type: xsd:DateTime
```
armintaenzertng, line 52:
```suggestion
  - type: xsd:dateTime
```

model/Build/Vocabularies/BuildTypeVocab.md
armintaenzertng, line 1: This Vocabulary is not used anywhere and the markdown file has several issues. Is this still intended to be used?
It looks like a collection of examples to me, maybe they could be moved over to the description of `buildType.md`?

model/Build/Classes/Build.md
armintaenzertng, line 1: @goneall: I'm not sure about this, but as you raised concerns over multiple properties with the same name, do you think it will be problematic if a class has the same name as a profile?

armintaenzertng (changes requested):
Hi, thanks for bringing in the build profile! :)
I have a few remarks that are mainly concerned with the ability to automatically parse the markdown files.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -56,4 +58,12 @@ TODO
 - testDependency: TODOdescription
 - testTool: TODOdescription
 - variant: TODOdescription
+- buildInputOf: Input to the build instance
+- buildOutputOf: Output of the build instance
+- buildConfigOf: Build configuration of the build instance
+- buildToolOf: Build tool used in the build instance
```
nishakm, line 64: This already exists in 2.3. So this is OK.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -56,4 +58,12 @@ TODO
...
+- buildInvokedBy: Agent that invoked the build
+- buildOnBehalfOf: Person for which buildInvokedBy is acting on behalf of
+- buildHostOf: Element in which the build instance runs on
+- buildChildOf: Build instance that was created from another build instance
```
nishakm, line 68: Use ancestor relationship. One ancestor with lots of decendents.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -56,4 +58,12 @@ TODO
 - testDependency: TODOdescription
 - testTool: TODOdescription
 - variant: TODOdescription
+- buildInputOf: Input to the build instance
+- buildOutputOf: Output of the build instance
+- buildConfigOf: Build configuration of the build instance
+- buildToolOf: Build tool used in the build instance
+- buildInvokedBy: Agent that invoked the build
+- buildOnBehalfOf: Person for which buildInvokedBy is acting on behalf of
```
nishakm, line 66: Clarify that this is "Actor". Also need to clarify constraints on each of the relationships.

nishakm: Why not properties for buildInvokedBy and buildOnBehalfOf?

model/Build/Vocabularies/BuildTypeVocab.md
```diff
@@ -0,0 +1,17 @@
...
+
+# BuildTypeVocab
+
+## Description
+
+This is about the vocabulary of build types.
+
+## Metadata
+
+- Type: URI
```
zvr, line 11: xsd:anyURI

nishakm (commented):
- Commit suggestions
- Use builtDate instead of endTime. Or buildStart and buildEnd

model/Build/Classes/Build.md
```diff
@@ -0,0 +1,57 @@
...
+  - type: xsd:string
+  - minCount: 0
+- configSourceUri
+  - type: anyURI
+  - minCount: 0
+- configSourceDigest
+  - type: Hash
+  - minCount: 0
+- parameters
+  - type: xsd:map<string>string
```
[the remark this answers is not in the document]
nishakm: Perhaps define a class in Core called "Dictionary" and Parameters can be of type Dictionary.

nishakm:
I think this is ready to merge @goneall

@armintaenzertng I'm deferring the question about whether it's OK to give the class and the profile the same name "Build" and what DataType to use for key-value pairs. From this week's tech call, these are going to be further discussed.

goneall:
Thanks @nishakm - I'll merge these in.

@armintaenzertng - for the remaining questions/issue please open a new issue or PR to track.

PR #96, opened 2023-03-07
Create energyConsumption.md

rgopikrishnan91:
Creating the property file for energyConsumption. @zvr @kestewart @lumjjb

### DIFF
--- a/model/AI/Properties/energyConsumption.md
+++ b/model/AI/Properties/energyConsumption.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# energyConsumption
+
+## Summary
+
+Field captures the amount of energy consumed to build this AI package 
+
+## Description
+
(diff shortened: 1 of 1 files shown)


kestewart (approved):
TO DO:   we need to refine the Description to give guidance about how to fill this field in.  For example, what units are we looking for.  But this is good enough to start with.

Thank you!

bennetkl:
Here's more about energy profiling for an AI application during it's execution.  1).  Profiling tools can be used to monitor the energy consumption of the system, including the CPU, GPU, and memory usage.
or Cloud providers often provide tools to monitor the energy consumption of AI applications running on their platforms. 
Energy modelling uses mathematical models to estimate the energy consumption of an AI application. This technique can be used to predict the energy consumption of a system before it is deployed or to compare the energy efficiency of different algorithms or architectures.  Unit of metrics, I was thinking, was Joule (J)  or Watts.

    Benchmarking: Benchmarking involves running standardized tests to measure the energy consumption of AI applications. This technique can be used to compare the energy efficiency of different systems and algorithms.

PR #97, opened 2023-03-07
Create standardsCompliance.md

rgopikrishnan91:
Creating the property file for standardsCompliance. @zvr @kestewart @lumjjb


kestewart (changes requested):
Update Summary to match description.

kestewart (approved):
Thanks for fixing.   This looks better.

bennetkl:
Corrected a few typos ..Text field that captures the list of standards that the AI software including data and models complies with. This could include both published and unpublished standards  from ISO, IEEE, ETSI standards or others.  These standards could (but not necessarily) be used to satisfy a legal or regulatory requirement. Each standard is listed on a separate line or in a single line with each standard being comma separated.

PR #99, opened 2023-03-07
Create safetyRiskAssessmentType.md

rgopikrishnan91:
Creating the property file for safetyRiskAssessmentType. @zvr @kestewart @lumjjb


kestewart (approved):
Looks good as a starting point. 

We may want to add in a link to the definitions of risk.

kestewart (approved):
Thanks for adding the link to the Article.

bennetkl:
Confirmed link, and the explanation looks good.  One comment though, why is 's" in safety sometimes a 'capital' - only one occurrence, but I would have thought all should 'S'.

bennetkl (commented):
Review the use or capital of safety and Safety for Range and correct if needed.  This is not a blocker

PR #100, opened 2023-03-07
Create typeOfModel.md

rgopikrishnan91:
Creating the property file for typeOfModel. @zvr @kestewart @lumjjb


kestewart (approved):

bennetkl:
We might want to add in LLM as a type of model, but otherwise it looks good

bennetkl (commented):
Looks good, but it would be good to add LLM in the list of models, but not a blocker

PR #105, opened 2023-03-07
Create informationAboutTraining.md

rgopikrishnan91:
create properties file for informationAboutTraining @zvr @kestewart @lumjjb


bennetkl (approved):
I'm not familiar with active training, you might want to put an example.  This is not a blocker.

PR #107, opened 2023-03-07
Create metricsDecisionThresholds.md

rgopikrishnan91:
Create a properties file for metricsDecisionThresholds @zvr @kestewart @lumjjb


kestewart (changes requested):
suggest rewrite summary to"
Field that captures the thresholds being used for computation of those metrics described in the Metrics field."   The "Should match... "  should be moved down to the Description part.

bennetkl (approved):
Include Kate's proposed change and then it's good to go.

kestewart (approved):

PR #109, opened 2023-03-07
Create domain.md

rgopikrishnan91:
create properties file for domain @zvr @kestewart @lumjjb

### DIFF
--- a/model/AI/domain.md
+++ b/model/AI/domain.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


model/AI/domain.md
kestewart, line 1: "Field captures theDescribe" - doesn't make sense to me.   Clarify, and add proper punctuation to the Summary.

kestewart (changes requested):

bennetkl (approved):
Suggest changing Natural Language Processing (NLP).  This is not a blocker

kestewart (approved):

PR #115, opened 2023-03-09
add Relationship related descriptions

armintaenzertng:

### DIFF
--- a/model/Core/Properties/completeness.md
+++ b/model/Core/Properties/completeness.md
@@ -9,7 +9,7 @@ Provides information about the completeness of relationships.
 ## Description
 
 Completeness gives information about whether the provided relationships are
-complete, incomplete or if no assertion is made either way.
+complete, known to be incomplete or if no assertion is made either way.
 
 ## Metadata
 
--- a/model/Core/Properties/completeness.md
+++ b/model/Core/Properties/completeness.md
@@ -4,11 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Provides information about the completeness of relationships.
 
 ## Description
 
-A completeness is TODO
+Completeness gives information about whether the provided relationships are
+complete, incomplete or if no assertion is made either way.
 
 ## Metadata
 
--- a/model/Core/Properties/relationshipType.md
+++ b/model/Core/Properties/relationshipType.md
@@ -4,11 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Provides a reasonable estimation of the type of relation
+between two identified Elements from a developer perspective.
 
 ## Description
 
-A relationshipType is TODO
+RelationshipType provides a reasonable estimation of the type of relation
+between two identified Elements from a developer perspective.
 
 ## Metadata
 
--- a/model/Core/Properties/relationshipType.md
+++ b/model/Core/Properties/relationshipType.md
@@ -4,13 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provides a reasonable estimation of the type of relation
-between two identified Elements from a developer perspective.
+Information about the relationship between two Elements.
 
 ## Description
 
-RelationshipType provides a reasonable estimation of the type of relation
-between two identified Elements from a developer perspective.
+This field provides information about the relationship between two Elements.
+For example, you can represent a relationship between two different Files,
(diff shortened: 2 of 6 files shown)


model/Core/Properties/completeness.md
```diff
@@ -4,11 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 ## Summary
 
-TODO
+Provides information about the completeness of relationships.
 
 ## Description
 
-A completeness is TODO
+Completeness gives information about whether the provided relationships are
+complete, incomplete or if no assertion is made either way.
```
goneall, line 12: Suggest replacing `complete, incomplete ...` with `complete, known to be incomplete ...`

model/Core/Properties/relationshipType.md
```diff
@@ -4,11 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Provides a reasonable estimation of the type of relation
+between two identified Elements from a developer perspective.
```
goneall, line 8: I think the above 2 lines can be copied from the [similar section in the 2.3 spec](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/chapters/relationships-between-SPDX-elements.md#1111-description):

`This field provides information about the relationship between two SPDX elements. For example, you can represent a relationship between two different Files, between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument.`

goneall: Actually, this would be more appropriate for the description.  For the summary, it could be shorter: `Information about the relationship between two SPDX elements.`

model/Core/Properties/relationshipType.md
```diff
@@ -4,11 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Description
 
-A relationshipType is TODO
+RelationshipType provides a reasonable estimation of the type of relation
+between two identified Elements from a developer perspective.
```
goneall, line 13: See comment above

model/Core/Vocabularies/RelationshipCompleteness.md
```diff
@@ -4,19 +4,21 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Indicates whether a relationship is complete or incomplete or if there
```
goneall, line 7: similar to comment above - `incomplete` -> `known to be incomplete`

model/Core/Vocabularies/RelationshipCompleteness.md
```diff
@@ -4,19 +4,21 @@ SPDX-License-Identifier: Community-Spec-1.0
...
+can be made no assertion either way.
 
 ## Description
 
-TODO
+RelationshipCompleteness indicates whether a relationship is complete or incomplete or if there
```
goneall, line 12: Similar comment `incomplete` -> `known to be incomplete`

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+A reasonable estimation of the relation between two identified Elements from a developer perspective.
```
goneall, line 7: Same comments as above - prefer to use the 2.3 spec wording

goneall (changes requested):
Thanks @armintaenzertng for taking this on.  Two general comments outlined in detail below:
- would prefer to copy over the 2.3 descriptions if the semantics are the same
- In the descriptions, would like to point out that incomplete means known to be incomplete to distinguish it from no assertion

model/Core/Properties/completeness.md
armintaenzertng: done

model/Core/Properties/relationshipType.md
armintaenzertng: done

model/Core/Properties/relationshipType.md
armintaenzertng: done

model/Core/Vocabularies/RelationshipCompleteness.md
armintaenzertng: done

model/Core/Vocabularies/RelationshipType.md
armintaenzertng: done

goneall (approved):
LGTM

goneall:

PR #116, opened 2023-03-09
add IntegrityMethod related descriptions

armintaenzertng:

### DIFF
--- a/model/Core/Classes/IntegrityMethod.md
+++ b/model/Core/Classes/IntegrityMethod.md
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Provides an independently reproducible mechanism that permits identification of a specific Element.
 
 ## Description
 
-TODO
+An IntegrityMethod provides an independently reproducible mechanism that permits identification
+of a specific Element that correlates to the data in this SPDX document. This identifier enables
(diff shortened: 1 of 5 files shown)


model/Core/Classes/IntegrityMethod.md
```diff
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 ## Summary
 
-TODO
+Provides an independently reproducible mechanism that permits identification of a specific Element.
 
 ## Description
 
-TODO
+An IntegrityMethod provides an independently reproducible mechanism that permits identification
+of a specific Element that correlates to the data in this SPDX document. This identifier enables
```
goneall, line 12: not really an identifier, suggest changing `identifier` to `value`

model/Core/Properties/algorithm.md
```diff
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Specifies the algorithm used for calculating the hash value.
```
goneall, line 7: If we change the property `hash` to `checksum` we may want to use `checksum` rather than hash - same comment applies to the use of hash in other descriptions and files

goneall (changes requested):
Just a couple minor suggestions and a reference to an open issue on changing hash back to checksum

goneall:

PR #119, opened 2023-03-09
add Annotation related descriptions

armintaenzertng:

### DIFF
--- a/model/Core/Vocabularies/AnnotationType.md
+++ b/model/Core/Vocabularies/AnnotationType.md
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Specifies the type of an annotation.
 
 ## Description
 
-TODO
+AnnotationType specifies the type of an annotation.
 
 ## Metadata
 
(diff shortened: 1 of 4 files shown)


model/Core/Vocabularies/AnnotationType.md
```diff
@@ -4,18 +4,17 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 ## Metadata
 
 - name: AnnotationType
 
 ## Entries
 
-- other: TODOdescription
-- review: TODOdescription
-
+- other: Used to store extra information about an Element during creation.
```
goneall, line 19: Although this generally follows the 2.3 documentation, the other type may also be used after the creation.  Suggest softening this a bit like: 

`Used to store extra information about an Element which is not part of a Review (e.g. extra information provided during the creation of the Element).`

goneall (commented):
One minor suggestion

model/Core/Vocabularies/AnnotationType.md
armintaenzertng: done

goneall (approved):
LGTM

goneall:
This PR only adds missing text, merging per the process discussed on 28 Mar 2023 tech call

PR #121, opened 2023-03-09
add Collection related descriptions

armintaenzertng:

### DIFF
--- a/model/Core/Classes/NamespaceMap.md
+++ b/model/Core/Classes/NamespaceMap.md
@@ -9,7 +9,8 @@ A mapping between prefixes and namespace URIs.
 ## Description
 
 A namespace map allows the creator of a collection of Elements to use
-memorable identifiers ("prefixes") instead of URIs by defining a mapping between them.
+shorter identifiers ("prefixes") instead of URIs to provide a more
+human-readable and smaller serialized representation of the Elements.
 
 ## Metadata
 
--- a/model/Core/Classes/NamespaceMap.md
+++ b/model/Core/Classes/NamespaceMap.md
@@ -4,11 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+A mapping between prefixes and namespace URIs.
 
 ## Description
 
-TODO
+A namespace map allows the creator of a collection of Elements to use
+memorable identifiers ("prefixes") instead of URIs by defining a mapping between them.
 
 ## Metadata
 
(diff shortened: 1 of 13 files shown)


model/Core/Classes/NamespaceMap.md
```diff
@@ -4,11 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 ## Summary
 
-TODO
+A mapping between prefixes and namespace URIs.
 
 ## Description
 
-TODO
+A namespace map allows the creator of a collection of Elements to use
+memorable identifiers ("prefixes") instead of URIs by defining a mapping between them.
```
goneall, line 12: Suggest - replace above with `shorter identifiers ("prefixes") instead of URIs to provide a more human readable and smaller serialized representation of the Element.

model/Core/Classes/SpdxDocument.md
```diff
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+Assembles a collection of Elements under a common string, the name of the document.
 
 ## Description
 
-TODO
+An SpdxDocument assembles a collection of Elements under a common string, the name of the document.
```
goneall, line 11: Suggest adding - `Commonly used when representing a unit of transfer of SPDX Elements.

model/Core/Properties/prefix.md
```diff
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-TODO
+A memorable substitute for a URI.
```
goneall, line 7: I would replace `memorable` with a `shorthand` or just remove `memorable.

model/Core/Properties/prefix.md
```diff
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Description
 
-A prefix is TODO
+A prefix is a memorable substitute for a URI.
```
goneall, line 11: Same comment as above

goneall (changes requested):
Just a cursory review with a couple suggestions.

model/Core/Classes/NamespaceMap.md
armintaenzertng: done

model/Core/Properties/prefix.md
armintaenzertng: done

model/Core/Properties/prefix.md
armintaenzertng: done

goneall (approved):
LGTM

goneall:
Merging - only changes are update to TODO's or missing changes

PR #136, opened 2023-03-24
Add `xsd:` namespace to `anyURI`

armintaenzertng:
anyURI is part of the xml schema: https://www.w3.org/TR/xmlschema11-2/#anyURI
Therefore we should prepend it with the `xsd:` namespace.


maxhbr (approved):
LGTM

PR #139, opened 2023-03-26
Add CI job to run spec-parser and upload its output

maxhbr:
- this clones and runs https://github.com/spdx/spec-parser/
- the generated files are uploaded as artifacts to the CI job

### DIFF
--- a/.github/workflows/spec-parser.yml
+++ b/.github/workflows/spec-parser.yml
@@ -15,7 +15,7 @@ jobs:
           python-version: "3.11"
       - name: get spec-parser
         run: |
-          git clone https://github.com/spdx/spec-parser/
+          git clone --depth=1 --single-branch --branch=main https://github.com/spdx/spec-parser/
           cd spec-parser
           python3 -m pip install -r requirements.txt
           python main.py -h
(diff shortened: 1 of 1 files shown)


maxhbr:
An example CI run can be found here: https://github.com/maxhbr/spdx-3-model/actions/runs/4525612756. There you can also download the zip containing all artifacts

zvr:
Thanks, @maxhbr, this is great.

Can we have the generated output appearing in a `generated`branch?

maxhbr:
sure, but I would leave that for a follow-up PR

.github/workflows/spec-parser.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: get spec-parser
+        run: |
+          git clone https://github.com/spdx/spec-parser/
```
zvr, line 18: `--depth=1 --single-branch --branch=main` is what I'm usually adding (not that it matters much in this case)

PR #153, opened 2023-03-30
Create Dataset.md

rgopikrishnan91:
Hi @kestewart @bennetkl please review


kestewart (approved):
Will need to refine this and layer in other mandatory fields by shapes, but this should start it off.

PR #154, opened 2023-03-30
Create anonymizationMethodUsed.md

rgopikrishnan91:
Hi @kestewart @bennetkl Please review

### DIFF
--- a/model/Dataset/Properties/anonymizationMethodUsed.md
+++ b/model/Dataset/Properties/anonymizationMethodUsed.md
@@ -14,4 +14,4 @@ The field describes the anonymization methods used. If multiple anonymization me
 
 - name: sensors
 - Nature: DataProperty
-- Range: xsd: string
+- Range: xsd:string
--- a/model/Dataset/Properties/anonymizationMethodUsed.md
+++ b/model/Dataset/Properties/anonymizationMethodUsed.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# anonymizationMethodUsed
+
+## Summary
+
+The field describes the methods used to anonymyze the dataset (of fields in the dataset).
+
+## Description
+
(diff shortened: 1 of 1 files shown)


kestewart (changes requested):
Fix typo "anonymyze" in summary. 
In metadata name shouldn't be sensors.

model/Dataset/Properties/anonymizationMethodUsed.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Description
+
+The field describes the anonymization methods used. If multiple anonymization methods are used, each anonymization method is specified in a new line. 
+
+## Metadata
+
+- name: sensors
+- Nature: DataProperty
+- Range: xsd: string
```
armintaenzertng, line 17:
```suggestion
- Range: xsd:string
```

model/Dataset/Properties/anonymizationMethodUsed.md
```diff
@@ -0,0 +1,17 @@
...
+
+# anonymizationMethodUsed
+
+## Summary
+
+The field describes the methods used to anonymyze the dataset (of fields in the dataset).
+
+## Description
+
+The field describes the anonymization methods used. If multiple anonymization methods are used, each anonymization method is specified in a new line.
```
armintaenzertng, line 11: Same comment as [here](https://github.com/spdx/spdx-3-model/pull/163#discussion_r1156985033).
That is, `anonymizationMethodUsed` has arbitrary cardinality when used in the [Dataset class](https://github.com/spdx/spdx-3-model/blob/dataset-profile/model/Dataset/Classes/Dataset.md), so we should not discuss multiple values in this property.

bennetkl (commented):
Typically with AI applications, data anonymization is done by creating a mirror image of the dataset and then  implementing alteration strategies, such as character shuffling, encryption, term, or character substitution.     This field is meant to capture all the strategies used to 'change' the dataset

goneall:
Superseded by PR #185 which includes feedback form this PR

PR #163, opened 2023-03-30
Create sensors.md

rgopikrishnan91:
Hi @kestewart  and @bennetkl please review

### DIFF
--- a/model/Dataset/Properties/sensors.md
+++ b/model/Dataset/Properties/sensors.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# sensors
+
+## Summary
+
+The field describes what sensors where used for collecting the data (if any sensors were used) 
+
+## Description
+
+Each sensor and their callibration value in form Sensor:calibration is noted. If multiple sensors are used, each sensor and thier callibration value is noted ina new line
+
+## Metadata
+
+- name: sensors
+- Nature: DataProperty
+- Range: xsd:string


kestewart (changes requested):
Fix typo "ina" possibly should be " on a"

kestewart:
Also, fix typo: "where"  should be "were" in Summary

model/Dataset/Properties/sensors.md
```diff
@@ -0,0 +1,17 @@
...
+
+# sensors
+
+## Summary
+
+The field describes what sensors where used for collecting the data (if any sensors were used) 
+
+## Description
+
+Each sensor and their callibration value in form Sensor:calibration is noted. If multiple sensors are used, each sensor and thier callibration value is noted ina new line
```
armintaenzertng, line 11: `sensors` has arbitrary cardinality when used in the [`Dataset` class](https://github.com/spdx/spdx-3-model/blob/dataset-profile/model/Dataset/Classes/Dataset.md), so we should not discuss multiple values in this property.

bennetkl (commented):
Seems like we should use map, we can discuss in today's AI sbom meeting and I can provide description in Gopi's absense

kestewart:
Discussed in call, and we'll apply mapping.   This should not be a blocker though for merging.     After a class is identified, we'll apply in a single pass.

goneall:
Superseded by PR #187 which includes feedback with the exception of the map which is referenced in Issue #186

PR #170, opened 2023-04-02
External properties restrictions

zvr:
Adds restrictions to external properties

The way to specify restrictions to external properties is:

- refer to the external property as a complete path of /{profile}/{class}/{property}
- add the shape defintion, which overwrites the original one.

Realistically, the restrictions will mostly be optional properties becoming mandatory,
i.e., minCount changing from 0 to 1.

### DIFF
--- a/model/Core/Classes/SpdxDocument.md
+++ b/model/Core/Classes/SpdxDocument.md
@@ -24,3 +24,8 @@ Commonly used when representing a unit of transfer of SPDX Elements.
   - minCount: 1
   - maxCount: 1
 
+## External properties restrictions
+
+- /Core/Element/name
+  - minCount: 1
+
--- a/model/Software/Classes/File.md
+++ b/model/Software/Classes/File.md
@@ -13,7 +13,7 @@ TODO This is about the File class.
 ## Metadata
 
 - name: File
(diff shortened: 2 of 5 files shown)


goneall (approved):
LGTM

model/Core/Classes/SpdxDocument.md
```diff
@@ -24,3 +24,8 @@ Commonly used when representing a unit of transfer of SPDX Elements.
   - minCount: 1
   - maxCount: 1
 
+## External properties restrictions
+
+- /Core/Element/name
```
maxhbr, line 29: you use `/` as seperator and in other places we use `:`:

https://github.com/spdx/spdx-3-model/blob/01360996d37c1d913f3d15e2d5911ace182ed200/model/Software/Classes/File.md?plain=1#L16

Can we make this consistent?

maxhbr (changes requested):

model/Core/Classes/SpdxDocument.md
maxhbr: It would probably be best, to align with the way rdf is referencing the properties.

goneall (approved):
LGTM

PR #173, opened 2023-04-03
Add Actor identifier

davaya:
As discussed at the 3/21 tech meeting, add an "identifier" property to the Actor class to replace the use of external identifier.

**Rationale:** An Agent needs an individual instance identifier, as opposed to the class identifier types used by ExternalIdentifierType (cpe_2.2, cpe_2.3, purl, uri-scheme, gitoid, other).  Email was added to ExternalIdentifierType to support agent IDs, but that mixes apples and oranges - artifacts do not have agent identifiers and vice versa.

**Semantics:** An email identifier has the syntax of an email address ("username@domain"), but like a URN or User Principal Name (UPN) or Decentralized Identifier ("subject@controller") or blockchain wallet id ("1N52wHoVR79PMDishab2XmRHsbekCdGquK@Bittrex.com") it is not necessarily a resource locator.  It's essential feature is that it has two parts: a local id and an assigner.

An earlier Person example had an "email" type and "account" type, where the first was contact info and the second a unique name.  Agent could have separate "contact" and "identifier" attributes but the distinction seems unnecessary in the way that URIs are either URLs or URNs.  Thus a single "identifier" attribute is proposed.

This PR highlights four changes to the model diagram:
1) Add "identifier" property to Agent of type Email
2) Add Email simple datatype that is a string of form "local-id@domain"
3) Replace Person externalIdentifier example iwth identifier
4) Remove "email" from ExternalIdentifierType, since Artifacts do not have instance identifiers.

(Highlighting is for discussion purposes, not for the updated diagram. Model tables will be updated later if the diagram is changed.)

### DIFF
--- a/model/Software/Classes/Package.md
+++ b/model/Software/Classes/Package.md
@@ -51,6 +51,10 @@ Note that some of these could be represented in SPDX as a file as well.
   - type: xsd:anyURI
   - minCount: 0
   - maxCount: 1
+- sourceInfo
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
 
 ## External properties restrictions
 
--- a/model/Software/Properties/sourceInfo.md
+++ b/model/Software/Properties/sourceInfo.md
@@ -1,0 +1,23 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# sourceInfo
+
+## Summary
+
+Records any relevant background information or additional comments
+about the origin of the package.
+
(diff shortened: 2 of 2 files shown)


maxhbr:
you add a random arrow under Bundle:


Can you please also touch markdown files to keep them in sync. We should not diverge.

davaya:
Replaced by PR #202 as a result of branch sync issues.

PR #175, opened 2023-04-04
Rename class CreationInformation to CreationInfo

### LINKED ISSUE #134: Rename class CreationInformation to CreationInfo
SPDX 2.3 uses the name CreationInfo for name of the class containing creation information.  The current SPDX 3.0 model uses the name CreationInformation.

I don't recall discussing a rationale for the change, so it may just be an oversight.

Suggest changing the name to be compatible with the SPDX 2.3 spec.

goneall:
Fixes #134


zvr (approved):

kestewart (approved):
Looks good.

PR #176, opened 2023-04-05
Move datasetSize.md

armintaenzertng:
Was in `properties`, is now in `Properties`.
Also fixed a few typos and names.


model/Dataset/Properties/datasetSize.md
```diff
@@ -0,0 +1,19 @@
...
+# datasetSize
+
+## Summary
+
+The field describes the size of the dataset.
+
+## Description
+
+The field describes how a large a dataset is. 
+The size can be specified using any unit (e.g., Bytes, MegaBits).
```
armintaenzertng, line 12: I'm opposed to the idea of allowing "any unit", especially in light of mixing bits and bytes.
I'd propose to use a more restricted set of possible units (e.g. K, B, M, G, T) with accurate definitions to ensure that this field can be universally understood correctly.

goneall: @armintaenzertng Discussed with @kestewart and agreed (provisional) that bytes are OK for the model as long as any serializations that are intended to be human readable specify formats that allow for less typing (e.g. GB, MB, TB) and the corresponding less precision.

Please update the PR to reflect bytes.

PR #177, opened 2023-04-05
Add datasetUpdateMechanism.md

armintaenzertng:
I noticed that the `datasetUpdateMechanism.md` was missing from the Properties folder.
As I'm not entirely sure about the intention of this property, my initial summary/description is likely to miss a few details, please have a look @kestewart! :)


armintaenzertng:
Yes, I would like to give the descriptions another pass as soon as the actual content is all laid down.

PR #178, opened 2023-04-05
Software profile: add sourceInfo

### LINKED ISSUE #86: SPDX 2->3 conversion of Package SourceInfo
Part of #74:

There does not seem to be an SPDX3 analogue of SPDX2's `sourceInfo`. Should this be part of the `comment` property of `Element`?

This has been discussed in the tech team call on 2023-02-28. It seems like an oversight and should be added as a property to `Package`.

armintaenzertng:
fixes #86


goneall (approved):

kestewart (approved):
Looks good to me.

Issue #167, opened 2023-03-31
How to treat properties with the same name in different profiles?


armintaenzertng:
The [Security profile currently defines](https://github.com/spdx/spdx-3-model/blob/85735497db226e0afb8d96081ec1232390e3bdf7/security-profile/security-profile.png) a `locator` property in the class `ExploitCatalogVulnerabilityCategorization`.
 `locator` is also a property of [`Core:ExternalReference`](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/ExternalReference.md).

This made me wonder how we would treat properties with the same name but in different profiles. As they live in different namespaces there will probably not be a clash as long as the full URIs are being used.
But maybe there are other instances where the same name can cause problems nevertheless?

goneall:
I would really like to avoid using the same property name in different profiles which have different semantics.  Having the same property name, even in different namespaces, will be very confusing and may trip up some serializations where namespaces are not natively supported.

goneall:
After the discussion on the defects, some of the property names are defined in other standards with different semantics.  I would propose different namespaces.  In the case of a different standard we use the namespace from that standard.  If such a namespace doesn't exist or if we're creating a duplicate property name within a separate profile, we use the namespace of the profile to make it unique.

In all cases, we would have different namespaces for properties with the same name but different semantics.

goneall:
@zvr Can we express different namespaces for the different properties in the markdown?

zvr:
Yes. @goneall. By definition if they are in `somenamespace/Properties` they are in that specific namespace.

For example, `filePurpose` is defined in the Software profile.

swinslow:
Just leaving a note here regarding @armintaenzertng's comment at https://github.com/spdx/spdx-3-model/pull/179#issue-1655679452, regarding the "name" and "comment" properties currently in the licensing profile.

I'm going to go ahead and merge #179 as-is with the present names, but noting this here so that we can track and address this if those property names need to be changed.

kestewart:
@zvr, @swinslow  - can we close thisnow?

swinslow:
From the Licensing profile perspective, I believe this is now done since I believe we've renamed the overlapping properties. If that's incorrect, or if this is still relevant for other profiles, please feel free to re-open. Thanks!

PR #179, opened 2023-04-05
Add License properties

armintaenzertng:
This adds the skeleton markdown files for the properties of the `License` class.

I noticed the following issue, already discussed in #167:
`name` and `comment` are already defined in Core. I suggest renaming them to `licenseName` and `licenseComment`. This in turn clashes with the already existing property `licenseComment`. The latter is not used (yet), and seems to only refer to concluded licenses, so I suggest renaming it to `concludedLicenseComment`.

### DIFF
--- a/model/Licensing/Properties/comment.md
+++ b/model/Licensing/Properties/comment.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# comment
+
+## Summary
+
+Comment on the license.
+
+## Description
+
+This field provides commentary on the license.
+
+## Metadata
+
+- name: comment
+- Nature: DataProperty
+- Range: xsd:string
--- a/model/Licensing/Properties/comment.md
+++ b/model/Licensing/Properties/comment.md
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Comment on the license.
+TODO
 
 ## Description
 
-This field provides commentary on the license.
+TODO
 
 ## Metadata
 
--- a/model/Licensing/Properties/deprecatedVersion.md
+++ b/model/Licensing/Properties/deprecatedVersion.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# deprecatedVersion
+
+## Summary
+
+TODO
+
+## Description
+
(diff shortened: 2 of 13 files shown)


swinslow:
Hi @armintaenzertng, just to understand, would this then replace the ‘type’ field under each of the properties in the License class definition file at https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Licensing/Classes/License.md?

armintaenzertng:
No, the properties just get their own definitions in a separate folder, the class will not be touched by this. See for example the [Element class](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Core/Classes/Element.md) which has its properties defined in the separate [Properties folder](https://github.com/spdx/spdx-3-model/tree/licensing-profile/model/Core/Properties).

swinslow:
Thanks @armintaenzertng. 

Is it redundant to list the `type` multiple times for each property, e.g. "type: xsd:string" for the `name` in the [Class definition](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Core/Classes/Element.md) and "Range: xsd:string" in the name [Property definition](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Core/Properties/name.md)? That seems redundant and potentially a place where errors would be introduced. I'm fine to go with what the team's doing in either case, but it seems like perhaps it would be preferable to only have it defined in one place, unless I'm missing something!

zvr:
It _is_ providing the type in two different locations, but otherwise the class file would not show much information and people would have to go to each property file to understand what this is about... We decided to have this compromise when we first started experimenting with this format.

maxhbr:
(off topic)

> ... but otherwise the class file would not show much information and people would have to go to each property file to understand what this is about...

this is a useful property for the generated end-user documentation and spec. But not for the defining input that is used for this generation.

swinslow (approved):
Thank you @armintaenzertng! I reset a few of the summaries and descriptions to TODO as a reminder to myself to fill those in with more details.

I'll go ahead and merge this, and will circle back subsequently with filled-in summaries and descriptions for these.

To your comment about potentially changing the names of `name` and `comment`, I left a note at https://github.com/spdx/spdx-3-model/issues/167#issuecomment-1509835948 for tracking purposes so that it can be revised if needed based on what's decided in #167.

PR #182, opened 2023-04-07
Add Dictionary class

goneall:
This commit adds a generic dictionary class as discussed on the SPDX tech call on April 4, 2023


maxhbr:
The name "dictionary" implies some rules which are not true or not enforced here. E.g. that a key can not be duplicate. 

I think that "A dictionary consists of a collection of key-value pairs." is not complete. Do we need to define handling of duplicate keys?

model/Core/Classes/Dictionary.md
```diff
@@ -0,0 +1,26 @@
...
+
+# DictionaryEntry
+
+## Summary
+
+A key with an associated value.
+
+## Description
+
+The class used for implementing a generic dictionary in SPDX.  A dictionary consists of a collection of key-value pairs. Each key-value pair maps the key to its associated value.  The DictionaryEntry is the type used for the generic key-value pair.  Both the key and the value are strings.
```
zvr, line 11: I suggest changing "implementing a generic dictionary" to "implementing a generic string mapping"

zvr: We should also mention (copying from wikipedia):

> associative arrays, and a data structure more commonly known today as a dictionary or hash table.

zvr (changes requested):

model/Core/Classes/Dictionary.md
goneall: I re-wrote the description to address the above comments as well as @maxhbr comment below.

I'm not very happy with the rewrite - but I can't come up with clearer wording that incorporates the feedback.

Please feel free to propose and specific language changes.

model/Core/Classes/Dictionary.md
```diff
@@ -0,0 +1,26 @@
...
+
+# DictionaryEntry
+
+## Summary
+
+A key with an associated value.
+
+## Description
+
+The class used for implementing a generic string mapping (also known as associative arrays, dictionaries, and hash maps) in SPDX.  To implement a dictionary, this class would be used in a collection with unique keys. Each DictionaryEntry would contain a key-value pair which maps the key to its associated value.  Both the key and the value are strings.
```
armintaenzertng, line 11:
```suggestion
The class used for implementing a generic string mapping (also known as associative arrays, dictionaries, and hash maps) in SPDX.  Each DictionaryEntry contains a key-value pair which maps the key to its associated value.  Both the key and the value are strings.  To implement a dictionary, this class is to be used in a collection with unique keys.
```
I don't think the subjunctive makes sense here.
I also suggest the above sentence reordering: In the current version, the particular concept of `key` in the DictionaryEntry is only first introduced in the sentence after it is referred to.

armintaenzertng (commented):
Thanks for adding this, Gary! I have a few minor remarks and a major one in the form of the file name :)

model/Core/Properties/value.md
```diff
@@ -0,0 +1,18 @@
...
+# value
+
+## Summary
+
+A value used in a generic key-value pair.
+
+## Description
+
+A key-value pair can be used to implement a dictionary which associates a key with a value.  
+The generic value is of type string.
```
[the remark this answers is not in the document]
goneall: I'll remove this from both the value.md and the DictionaryEntry.md file since it is covered in the metadata.

model/Core/Properties/key.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# key
+
+## Summary
+
+A key used in a key-value pair.
```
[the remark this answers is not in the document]
goneall: I'll take your suggestion for key.md and also for the value.md file.

model/Core/Classes/Dictionary.md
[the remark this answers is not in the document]
goneall: great catch @armintaenzertng!  I renamed the file.

model/Core/Properties/key.md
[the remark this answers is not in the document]
goneall: Updated to use generic

model/Core/Classes/Dictionary.md
```diff
@@ -0,0 +1,26 @@
...
+## Metadata
+
+- name: DictionaryEntry
+- Instantiability: Concrete
+
+## Properties
+
+- key
+  - type: xsd:string
+  - maxCount: 1
```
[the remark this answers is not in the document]
goneall: Another good catch - key should be required, although I think it is fine for value to be optional.

model/Core/Properties/value.md
```diff
@@ -0,0 +1,18 @@
...
+A value used in a generic key-value pair.
+
+## Description
+
+A value used in a generic key-value pair.
+A key-value pair can be used to implement a dictionary which associates a key with a value.  
+
+## Metadata
+
+- name: key
```
armintaenzertng, line 16:
```suggestion
- name: value
```
I seem to have missed that one in my last pass.

model/Core/Classes/DictionaryEntry.md
```diff
@@ -0,0 +1,27 @@
...
+
+# DictionaryEntry
+
+## Summary
+
+A key with an associated value.
+
+## Description
+
+The class used for implementing a generic string mapping (also known as associative arrays, dictionaries, and hash maps) in SPDX.  Each DictionaryEntry contains a key-value pair which maps the key to its associated value.  To implement a dictionary, this class is to be used in a collection with unique keys.
```
armintaenzertng, line 11:
```suggestion
The class used for implementing a generic string mapping (also known as associative array, dictionary, or hash map) in SPDX.  Each DictionaryEntry contains a key-value pair which maps the key to its associated value.  To implement a dictionary, this class is to be used in a collection with unique keys.
```
As "a generic string mapping" is singular, the following enumeration should be, too. The final "and" only makes sense to me if we know the list to be exhaustive. If that is the case, then ignore my "or" suggestion.

armintaenzertng, line 11: As an aside: I noticed you include two spaces after each period. Is there a reason for that?

armintaenzertng (commented):
Thanks for addressing my issues, Gary! :)
I have found another one I missed the last time, sorry...

model/Core/Classes/DictionaryEntry.md
goneall: Habit.  Before [2020 when Microsoft declared 2 spaces](https://www.theverge.com/2020/4/24/21234170/microsoft-word-two-spaces-period-error-correction-great-space-debate) as incorrect, both single and double spaces were allowed.  In school, I (and millions of others) were taught that 2 spaces was correct.

armintaenzertng: Oh wow, I didn't know about that! :O

goneall:
@armintaenzertng @maxhbr @zvr - This should now be ready to merge.  If you could give it a final review.

armintaenzertng (approved):
Looks good to me now, thanks :)

maxhbr:
I am still conflicted on the use of "dictionary", but I am not blocking the merge.

(we also use "set" in the license profile but there is no enforcement of uniqueness of members)

maxhbr:
Maybe we can call the object `KeyValue` instead of `DictionaryEntry`?

goneall:
> Maybe we can call the class KeyValue instead of DictionaryEntry?

I'm OK with either.  Maybe KeyValuePair.  We did make a rather hasty decision at the end of last week's tech call to use the term DictionaryEntry.  If we want to change the name, we should bring this up at the next tech call and see if there are any objections.

@maxhbr Do you think the name is important enough to (re)raise on the next tech call?

maxhbr:
maybe merge, and we can create a PR to rename it and discuss that

zvr:
Alternative names might include AVPair, AttributeValuePair, KVPair, etc.

But realistically, do we even have a case where these are used for something that can have multiple entries with the same key?

zvr (approved):

maxhbr (approved):
we should discuss naming again

goneall:
> But realistically, do we even have a case where these are used for something that can have multiple entries with the same key?

@kestewart Do you know if any of the AI useages would have 1 key with 2 different values?

If we don't know of any cases, I would propose we use DictionaryEntry as the name and change the summary and description to indicate that it only applies to the case where the keys are unique.  @maxhbr let me know if this solves your issue with naming.

If, and when, a need comes up for a generic key/value pair not used in a dictionary, we can create a new type/class that still use one key and one value, but has different summaries and descriptions.

maxhbr:
As said: merge this and discuss later. The discussion should not block the progress here.

bennetkl:
The only example I can think of is if synthetic (starting from the same source ) and  non-synthetic data are used in the same AI application.  I haven't seen it in any production application as yet.   Maybe others have seen an example

    On Wednesday, April 12, 2023 at 07:24:25 p.m. EDT, goneall ***@***.***> wrote:  
 
 



But realistically, do we even have a case where these are used for something that can have multiple entries with the same key?


@kestewart Do you know if any of the AI useages would have 1 key with 2 different values?

If we don't know of any cases, I would propose we use DictionaryEntry as the name and change the summary and description to indicate that it only applies to the case where the keys are unique. @maxhbr let me know if this solves your issue with naming.

If, and when, a need comes up for a generic key/value pair not used in a dictionary, we can create a new type/class that still use one key and one value, but has different summaries and descriptions.

model/Core/Classes/DictionaryEntry.md
seabass-labrax: I'm a single-spacer myself[1], but the original reason for this  was that double-spaces emulated on the (fixed-width) typewriters the correct spacing that would be there with full movable type printing presses! Of course the only real solution in the digital era is to use TeX :grinning:

[1]: `M-: (setq sentence-end-double-space nil) RET`

PR #191, opened 2023-04-11
Usage profile

yoshi-i:
Model descriptions with drawio and supplemental pdf document of Usage Profile

Signed-off-by: Yoshiyuki Ito <yoshiyuki.ito.ub@renesas.com>


maxhbr (dismissed):

swinslow:
Hello,

It doesn't look like there is a way to comment inline in the PDF itself, so I'm replying with a couple of thoughts here.

For slide 6 "Terms of Use for these deliverables": if this is about a contractual limitation on use, wouldn't that be more appropriate to include a licensing profile section with a `LicenseRef-` ID, pointing to the text of the contract? I don't think a separate field to refer to a contract that imposes different limitations on use of software would be helpful, since it would require an SBOM recipient to look in two separate places to understand the license limitations.

For slide 7 "Expiration date and time OR Expiration event": similarly, it would be helpful to have more explanation to understand what is "expiring" in this situation. I assume that it wouldn't be either the SBOM itself, or the license to use the software. Is there something else that is "expiring"?

doc/UsageProfile.svg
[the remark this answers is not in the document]
yoshi-i: @maxhbr -san, It was overwritten by my operation mistake. I've committed correct one.

goneall (approved):

goneall (changes requested):
I just saw @maxhbr comment - indeed the draw.io file is HTML rather than the draw.io data which can be downloaded.

yoshi-i:
> I just saw @maxhbr comment - indeed the draw.io file is HTML rather than the draw.io data which can be downloaded.

Sorry, I've re-committed at #38f3040 for that drawio file.

goneall (approved):
LGTM - Thanks @yoshi-i

goneall:
@maxhbr - pls review and if OK, we can merge

doc/UsageProfile.svg
maxhbr, line 1: I do not understand the `USAGE OPERATOR`, are there some examples?

maxhbr: This uses the DESCRIBES relation, but that is (at least in 2.3) defined as "Is to be used when SPDXRef-DOCUMENT describes SPDXRef-A." with the example "An SPDX document WildFly.spdx describes package ‘WildFly’. Note this is a logical relationship to help organize related items within an SPDX document that is mandatory if more than one package or set of files (not in a package) is present.". Not sure if it is valid in this usecase.

maxhbr: This shows a specific way of how licensing information is expressed, which might not align with the actual future of the licensing profile. Maybe this could be made transparent

maxhbr: I think we always use singular: 
* `Deliverables` -> `Deliverable`
* `comments` -> `comment`

maxhbr: The `AND` and `OR` looks very similar to license expressions and might cause confusion.

doc/UsageProfile-20230328_en.pdf
maxhbr, line 1: Could you maybe instead provide some file that can be edited and further improved instead of a PDF?

maxhbr (changes requested):

goneall:
The operations profile is now picking up some of these fields - I'm going to close this for now - @yoshi-i - If you would like to reopen, please take a look at the operations profile branch first and see if we still need the Usage profile.

PR #192, opened 2023-04-11
Add Dataset profile 

kestewart:
Draft dataset model is in reasonable shape for wider review, prior to adding to SPDX 3.0


model/Dataset/Classes/Dataset.md
```diff
@@ -0,0 +1,60 @@
...
+Provides information about the fields in the Dataset profile.
+
+## Description
+
+Metadata information that can be added to a a dataset that maybe used in a software or to train/test an AI package
+
+## Metadata
+
+- name: Dataset
+- SubclassOf: Software:Package
```
maxhbr, line 16:
```suggestion
- SubclassOf: /Software/Package
```

model/Dataset/Classes/Dataset.md
```diff
@@ -0,0 +1,60 @@
...
+- dataCollectionProcess
+  - type: xsd:string
+  - minCount: 0
+  - maxCount:1
+- intendedUse
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
+- datasetSize
+  - type: xsd:string
```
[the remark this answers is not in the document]
armintaenzertng: Ah, that's my bad, I only changed it in the `datasetSitze.md`.

model/Dataset/Properties/intendedUse.md
```diff
@@ -0,0 +1,17 @@
...
+
+# intendedUse
+
+## Summary
+
+The field describes what the given dataset should be used for.
+
+## Description
+
+Some datasets are collected to be used only for particular purposes. For example, medical data collected from a specific demography might only be applicable for training machine learning models to make predictions for that demography. In such a case, the intendedUse field would capture this information. Similarly, if a dataset is collected for building a facial recognition model, the intendedUse field would specify that.
```
swinslow, line 11: Is this field meant to capture only the _intended_ use (e.g. "I think you should only use the dataset for X"), or to capture _restrictions_ on use (e.g. "You may not use the dataset for Y")?

If it's the latter, I think that would more appropriately be handled by representing a license with that restriction (most likely via a `LicenseRef-` license), rather than having a separate data field for what is essentially a licensing restriction somewhere outside the licensing profile.

model/Dataset/Vocabulary/ConfidentialityLevelType.md
```diff
@@ -0,0 +1,22 @@
...
+
+Describes the different confidentiality levels as given by the [Trafic Light Protocol](https://en.wikipedia.org/wiki/Traffic_Light_Protocol)
+
+## Metadata
+
+- name: ConfidentialityLevelType
+
+## Entries
+
+- Red: Data points in the dataset are highly confidential and only can be shared with named recipients.
```
swinslow, line 19: Similar to the comment above for intended use -- is this Confidentiality Level Type meant to capture e.g. license restrictions on sharing of the data? If so, wouldn't a `LicenseRef-` be the preferred way to capture that?

model/Dataset/Properties/intendedUse.md
bennetkl: With AI applications, builders typically 'document' their intended use, but it's typically made in the context of positive so not use a restricted license profile would work.  Hope that makes sense on why license information wasn't used.

swinslow: That makes sense -- thank you @bennetkl!

model/Dataset/Classes/Dataset.md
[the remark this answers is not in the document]
kestewart: Good catch.   Applying.

model/Dataset/Vocabulary/ConfidentialityLevelType.md
kestewart: This is a standard vocabulary from the industry, which is orthogonal to actual license on the data.   Please open issue if you disagree with me marking this resolved.

goneall (approved):

goneall:
Per review with @kestewart and @nicoschif I'll go ahead and merge.

PR #195, opened 2023-04-12
Add dataset Availability to dataset profile. 

nicoschif:
Adding the datasetAvailability property to the dataset profile

### DIFF
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -59,6 +59,6 @@ Metadata information that can be added to a a dataset that maybe used in a softw
   - minCount: 0
   - maxCount: 1
 - datasetAvailability
-  - type: xsd:string
+  - type: DatasetAvailabilityType
   - minCount: 0
   - maxCount: 1
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -58,3 +58,7 @@ Metadata information that can be added to a a dataset that maybe used in a softw
   - type: xsd:string
   - minCount: 0
   - maxCount: 1
+- datasetAvailability
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
--- a/model/Dataset/Properties/datasetAvailability.md
+++ b/model/Dataset/Properties/datasetAvailability.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# datasetAvailability
+
+## Summary
+
+The field describes the availability of a dataset.
+
+## Description
+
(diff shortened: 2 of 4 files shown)


model/Dataset/Vocabulary/DatasetAvailability.md
```diff
@@ -0,0 +1,24 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# DatasetAvailability
```
armintaenzertng, line 3: The name of the vocabulary is the same as that of the property, which will likely lead to confusion. It seems to be common to append "Type" at the end of the vocabulary in that case, i.e. `DatasetAvailabilityType`.

model/Dataset/Classes/Dataset.md
```diff
@@ -58,3 +58,7 @@ Metadata information that can be added to a a dataset that maybe used in a softw
   - type: xsd:string
   - minCount: 0
   - maxCount: 1
+- datasetAvailability
+  - type: xsd:string
```
armintaenzertng, line 62: I presume it is planned to use the newly introduced vocabulary here, not just a string?

model/Dataset/Properties/datasetAvailability.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Description
+
+Some datasets are publicly available and can be downloaded directly. Others are only accessible behind a clickthrough, or after filling a registration form. This field will describe the dataset availability from that perspective.
+
+## Metadata
+
+- name: datasetAvailability
+- Nature: DataProperty
+- Range: xsd:string
```
armintaenzertng, line 17: see above

model/Dataset/Vocabulary/DatasetAvailability.md
bennetkl: Also think renaming with append of "Type" will make it clearer.  Good idea

model/Dataset/Classes/Dataset.md
bennetkl: Yes, it's to use the new definitions.  What did we not do correctly?   Nic listed all the possible 'strings' that would be accepted for this field.   @nicoschif with  the new concept of 'models pretrained with data, should we add an additional  term to the list?

armintaenzertng: The field should then state 
`- type: datasetAvailabilityType`
(provided the vocabulary is renamed to end with "Type")

kestewart (approved):
Thanks,  looks good to me for including into dataset profile, and getting it merged.

If there are further tweaks, we can do it in main directly.

model/Dataset/Vocabulary/DatasetAvailability.md
kestewart: New version with Type added.

PR #196, opened 2023-04-14
add builtTime, releaseTime and validUntilTime to Artifact

armintaenzertng:
These entries not only ensure compatibility with 2.3 but can also be used by other profiles like dataset.

### DIFF
--- a/model/Core/Properties/validUntilTime.md
+++ b/model/Core/Properties/validUntilTime.md
@@ -1,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# validUntilTime
+
+## Summary
+
+Specifies until when the artifact can be used before its usage needs to be reassessed.
+
+## Description
+
+A validUntilTime specifies until when the artifact can be used before its usage needs to be reassessed.
+
+## Metadata
+
+- name: validUntilTime
+- Nature: DataProperty
+- Range: xsd:dateTime
+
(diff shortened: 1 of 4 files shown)


swinslow:
What does "validUntilTime" mean in this context, regarding how long an "artifact" is "valid"?

If it is meant to imply something about the _use_ of the artifact no longer being permitted after the specified time, I don't think that's appropriate. Presumably that would be handled via something like a contractual prohibition, which I'd assume should be more appropriately reflected in a `LicenseRef-` license.

Or if it is suggesting that the _SPDX Document's data_ is no longer permitted to be used after the specified time, I'm not sure I follow that either. SPDX data is licensed under CC0-1.0, which places no limitations on how it may be used.

Is there some other meaning of "valid" that's intended here? If so, I'd suggest that it should be more clearly spelled out in the spec's definition of the "valid" term.

goneall (approved):

model/Core/Properties/validUntilTime.md
```diff
@@ -0,0 +1,18 @@
...
+
+# validUntilTime
+
+## Summary
+
+Specifies the time until which an artifact is valid.
+
+## Description
+
+A validUntilTime specifies the time until which an artifact is valid.
```
kestewart, line 11: Replace this with something closer to the semantics in 2.3. 

Suggest: 
" A valdiUntilTime specifies when the artifact can be used until it's useage needs to be reassessed"

armintaenzertng: fixed

kestewart (approved):

PR #197, opened 2023-04-14
AI profile: use DictionaryEntry where appropriate

armintaenzertng:
The AI profile uses some key-value-like properties that can make use of the structure provided by the new `DictionaryEntry` class.

### DIFF
--- a/model/AI/Properties/metricsDecisionThresholds.md
+++ b/model/AI/Properties/metricsDecisionThresholds.md
@@ -17,4 +17,4 @@ Each decision threshold should match with the metrics field defined in the AI Pa
 
 - name: metricsDecisionThresholds
 - Nature: DataProperty
-- Range: xsd:string
+- Range: /Core/DictionaryEntry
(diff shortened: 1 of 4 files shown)


model/AI/Properties/metricsDecisionThresholds.md
```diff
@@ -13,4 +13,4 @@ Each metric might be computed based on a decision threshold (for instance precis
```
goneall, line 13: Same comment as above - suggest removing:

```
This field records the decision threshold in comma separated values. E.g., metric_1:decsion_threshold, metric_2:decision threshold. If there are multiple models each line represents each model’s metrics and decision threshold. Each decision threshold should match with the metrics field defined in the AI Package.
```

goneall (changes requested):
@armintaenzertng Can you remove some of the description text that referred to the old text formats?

armintaenzertng:
@goneall, I fully agree with your suggestions, but I did not include them in this PR as they were already handled in another one. I rebased this PR now so it includes the reworked descriptions,

goneall (approved):
Thanks @armintaenzertng

kestewart (approved):
Changes match discussion.

rgopikrishnan91 (commented):
Looks good to me. I approve the changes

PR #198, opened 2023-04-14
Dataset profile: use DictionaryEntry for sensors property

armintaenzertng:
The `sensors` property uses a key-value-style format, which is provided by the new `DictionaryEntry` class.


goneall (approved):

kestewart (approved):
This matches what we discussed.

Issue #94, opened 2023-03-05
Agent or Identity


davaya:
The newly-simplified creation portion of the model reflects tech team consensus on the following features:

* The single createdBy top-level class is an Element, not a string.
* The top-level class has three sub-classes: Person, Organization, SoftwareAgent.
* Tool is a separate Element type, not under the top-level class, used in the new optional createdUsing property

If software can create on its own and has an individual instance identifier, it is a SoftwareAgent.  If it cannot act on its own then it is a Tool.

![spdx-id1](https://user-images.githubusercontent.com/19152940/222727722-ffa685b8-a1ff-4bd0-9430-6b419150a744.jpg)

The only open issue is what name should be given to the top-level Element.  In defining the SPDX v3 model, naming is less important than functionality, but naming does matter for external communication.  Sean and David have expressed conflicting views on which name, "Agent" or "Identity", best promotes understanding the SPDX model by the wider cybersecurity community.

**David:**
* The grayed-out "Identity" class, representing future work, is incorrect.  The community understands Identity to include SoftwareAgent:
    * [The law](https://www.law.cornell.edu/wex/legal_person) uses "Legal Person" as the parent of Person and Organization -- "an individual, company, or other entity which has legal rights and is subject to obligations."  Its subclasses are "Natural Person" for human entities and "[Juridical Person](https://en.wikipedia.org/wiki/Legal_person)" for non-human entities.
    * The law does not use Identity to mean Legal Person, it has only other meanings ("detection, prevention, and mitigation of [identity theft](https://www.law.cornell.edu/cfr/text/17/162.30)", "[chemical identity](https://www.law.cornell.edu/cfr/text/40/350.5) of an extremely hazardous substance", etc.)
* The software community includes SoftwareAgent Identities:
    * Azure Active Directory [device identity](https://learn.microsoft.com/en-us/azure/active-directory/devices/overview)
    * [Decentralized Identifiers](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers): "Anything can be the subject of a DID: person, group, organization, thing, or concept"
    * The software [Actor Model](https://en.wikipedia.org/wiki/Actor_model) adopts the philosophy that everything is an actor whose identity is a "mailbox". In an actor model ([Clojure](https://clojure.org/about/state#actors)), state is encapsulated in an actor (identity) and can only be affected/seen via the passing of messages (values).
[...]

PR #202, opened 2023-04-14
Change Agent to Actor and add Actor identifier

davaya:
Replaces #173


model/Core/Classes/Email.md
```diff
@@ -0,0 +1,21 @@
...
+
+# Email
+
+## Summary
+
+An individual identifier in RFC 822 email address format
+
+## Description
+
+An Email identifier is a String constrained to be an addr-spec as defined in RFC 822 Section 6.2,
```
kestewart, line 11: Add link so folks can find RFC 822

kestewart (changes requested):
Thanks for pulling all this together, and getting the diagram updated.

For the email, can you add in the RFC 822 link to the description.  We've been doing it with other external standards we reference in other parts of the spec, and it would be good to be consistent here.

kestewart (approved):
Thanks for adding the reference.

model/Core/Classes/Email.md
```diff
@@ -0,0 +1,22 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Email
+
+## Summary
+
+An individual identifier in RFC 822 email address format
```
armintaenzertng, line 7:
```suggestion
An individual identifier in RFC 822 email address format.
```

armintaenzertng (commented):
I like periods at the end of lines. :)

model/Core/Classes/Actor.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Actor
+
+## Summary
+
+An individually-identified Person, Organization or SoftwareAgent
```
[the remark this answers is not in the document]
davaya: camelCase is used for properties, PascalCase is used for classes, and lower case text in the description refers to entities described by SPDX elements.

So it was intentional to refer to model classes in the Summary and real-world entities in the Description, but I won't claim my intentions make any sense :-).

Description: Actor is a grouping of identifying characteristics unique to an individual person, organization or software agent.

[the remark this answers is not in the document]
davaya: I try to use periods at the end of sentences and leave them off for lines that are not sentences (e.g. phrases with no verb).  But consistency with the rest of the documentation is more important - I'll look at all the classes for guidance.

goneall (approved):
LGTM - but we should have @iamwillbar review since he's been more involved in the identity discussions

model.png
armintaenzertng, line 1: While you are at it, you can also change the name of the Agent class to Actor.

model/Core/Classes/Actor.md
```diff
@@ -0,0 +1,20 @@
...
+Actor is a grouping of identifying characteristics unique to an individual person, organization or software agent.
+
+## Metadata
+
+- name: Actor
+- SubclassOf: Entity
+- Instantiability: Concrete
+
+## Properties
+
```
armintaenzertng, line 20: This misses the newly introduced property?
According to the model.png, this should have a property `identifier` of type `Email`.

Issue: `identifier` is already defined in [ExternalIdentifier.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/ExternalIdentifier.md) with type `xsd:string`.

Double issue: The `identifier` property is missing from the [Properties](https://github.com/spdx/spdx-3-model/tree/main/model/Core/Properties) folder.

armintaenzertng (changes requested):
I overlooked a critical issue in my last review, see below.

model.png
davaya: The agreed name prior to the PR is Agent.   I prefer Actor because Agent has two conflicting definitions (an entity that itself has agency, and an entity representing a different entity), but Sean and William can decide.

model/Core/Classes/Actor.md
davaya: The ExternalIdentifier property and the Agent/Actor property have different semantics and different syntax - Agent must be a local@domain identifier that uniquely identifies a single instance.  ExternalIdentifier can (and usually does) identify a class of things (such as a product) rather than an individual instance of that product.

It's not a problem for two Classes to have properties with the same name and different types (Email vs. unrestricted String).  But creating a Property file would require listing one type and ignoring the other.  Email, SemVer and NonNegativeInteger Classes don't have that problem.  Recommendation: do not create an `identifier` file in the Properties folder.

zvr (changes requested):

model/Core/Classes/Actor.md
davaya: [SemVer](https://github.com/spdx/spdx-3-model/tree/main/model/Core/Classes) is a Class.
The "specVersion" property of [CreationInfo](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/CreationInfo.md) is type SemVer.  Assume for the sake of argument that the property name is "version".

Now assume that you want [ExternalIdentifier](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/ExternalIdentifier.md) to have a "version" property, but that ExternalIdentifier versions are PURLs (or Integers, or anything - it doesn't have to make sense).  Add the version property to the ExternalIdentifier class and voila, it's represented.

zvr: In general, we do not follow an object-oriented model where we have "classes" that have "properties". We define the ontology which includes Classes and Properties. Instances of more than one classes may have the same properties. For example both an instance of an `Element` and of a `CreationInfo` have a `comment`. And it goes the other way: if _something_ has an `algorithm` and a `hashValue`, then this something also is of class `Hash`.

Therefore we can not "add the version property to the ExternalIdentifier class" in your example.

model/Core/Classes/Actor.md
```diff
@@ -18,3 +18,7 @@ Actor is a grouping of identifying characteristics unique to an individual perso
 
 ## Properties
 
+- identifier
+  - type: Email
```
zvr, line 22: The property `identifier` is existing, of type `xsd:string` and is being used in `ExternalIdentifier` class.

zvr (changes requested):

model/Core/Classes/Actor.md
davaya: > In general, we do not follow an object-oriented model where we have "classes" that have "properties".

We are not doing object-oriented programming based on the concept of "objects" which contain data and code.

We are designing structured data, and JSON uses the word "object" for data values that have named "members", and for mapping types that map "keys" to "values".  So you are saying we aren't following a design approach where data "objects" have "keys" with "values"?  That doesn't sound scalable.

In the global describe-everything-in-the-universe ontology, common property names like "version" and "identifier" are first-come-first-served?  If SPDX decides that the "version" property must be a SemVer string, we've:
1) done a copyright search to ensure that nobody else has ever defined a "version" property with a different type, and 
2) can be confident that everybody else will follow the stake that we've claimed?

I'm not arguing that the model cannot have a set of common "properties" with names and values.  I am arguing that those property files are a library of name-value pairs that MAY be used, and not a set of constraints that MUST be followed.  Where "identifier" and "version" could have different types in different classes, the model could either refrain from defining identifier and value properties entirely (my recommendation), or define a property with one type while not prohibiting other types.

kestewart:
@iamwillbar to take point on resolving through this, as there seems to be some evolution from discussions.

kestewart:
In discussion with William - the W3 specification uses the term Agent (rather than Actor).

goneall:
In discussing issue #94 on the tech call, the general consensus for the attendees was that this PR doesn't line up with the decision completely.

@nishakm will create a separate PR.

We'll leave this PR open for now and decide which to merge and which to close.

davaya:
Superseded by #340

PR #213, opened 2023-04-17
Core: add PositiveIntegerRange class

armintaenzertng:
I noticed that the `PositiveIntegerRange` class and its properties were not yet defined in markdown.
`PositiveIntegerRange` is currently only used in the Software namespace in the properties `byteRange` and `lineRange` of the `Snippet` class, but I think it is safer to include it in the Core namespace to enable future extensions to make use of it.

### DIFF
--- a/model/Core/Classes/PositiveIntegerRange.md
+++ b/model/Core/Classes/PositiveIntegerRange.md
@@ -1,0 +1,30 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PositiveIntegerRange
+
+## Summary
+
+A tuple of two positive integers that define a range.
+
+## Description
+
+PositiveIntegerRange is a tuple of two positive integers that define a range.
+"begin" must be less than or equal to "end".
+
+## Metadata
+
+- name: PositiveIntegerRange
+- SubclassOf: none
+- Instantiability: Concrete
+
+## Properties
+
+- begin
+  - type: xsd:positiveInteger
+  - minCount: 1
+  - maxCount: 1
+- end
+  - type: xsd:positiveInteger
+  - minCount: 1
+  - maxCount: 1
+
--- a/model/Core/Properties/begin.md
+++ b/model/Core/Properties/begin.md
@@ -1,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# begin
+
+## Summary
+
+Defines the beginning of a range.
+
+## Description
+
+begin is a positive integer that defines the beginning of a range.
+
+## Metadata
+
+- name: begin
+- Nature: DataProperty
+- Range: xsd:positiveInteger
+
--- a/model/Core/Properties/end.md
+++ b/model/Core/Properties/end.md
@@ -1,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# end
+
+## Summary
+
+Defines the end of a range.
+
+## Description
+
+end is a positive integer that defines the end of a range.
+
+## Metadata
+
+- name: end
+- Nature: DataProperty
+- Range: xsd:positiveInteger
+
--- a/model/Software/Classes/Snippet.md
+++ b/model/Software/Classes/Snippet.md
@@ -27,11 +27,11 @@ may have been originally created under another license or copied from a place wi
   - type: SoftwarePurpose
   - minCount: 0
 - byteRange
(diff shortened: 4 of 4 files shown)


goneall:
The PR looks good, but in reviewing this I realized we probably need to changes this to `nonNegativeInteger` and `nonNegativeIntegerRange` since a value of zero is allowed.  See https://www.w3.org/TR/xmlschema-2/#nonNegativeInteger.

davaya:
The NonNegativeInteger class should also be added along with the SemVer, MediaType, and Email classes.  This is particularly true since it may be useful in places other than the NonNegativeIntegerRange class.  Creating two or more "properties" with different names to refer to the same class is not a good design practice because it invites inconsistency (as well as wastes space).

We may also want to include an explanatory note in the Description that PositiveInteger is all integers greater than or equal to one, while NonNegative integer is all integers greater than or equal to zero.

goneall:
> The NonNegativeInteger class should also be added ...

There's actually an XSD data type nonNegativeInteger - so we can use that.  Valid point for the other classes, however.

armintaenzertng:
Sorry, I'm not sure now whether we need both `NonNegativeIntegerRange` and `PositiveIntegerRange` or only the former.
Line counts start with 1, so in that case we would still need `PositiveIntegerRange`?

davaya:
> There's actually an XSD data type nonNegativeInteger - so we can use that. Valid point for the other classes, however.

But the property files should not be added.  There is no benefit to creating "start" and "end" files in the Properties folder since the XSD type is used in the Range class.  And the harm is that it both:
* gives the idea that new Properties files would be created for all other uses of nonNegativeInteger such as "count" or "width"
* precludes the use of "start" and "end" property names in other classes where the values are not nonNegativeIntegers.

goneall:
> Line counts start with 1, so in that case we would still need PositiveIntegerRange?

If both `byteRange` and `lineRange` start at 1, I would agree we only need `PositiveIntegerRange`

zvr:
@davaya I disagree with your statement about not having property files.

We *do* want to have property files for each and every property, to provide description and other metadata.

davaya:
@zvr: We do want to provide description and other metadata.  But that is what Class files provide without the disadvantages I listed.

**Alternative 1:**
* Define the NonNegativeInteger class with all the description and metadata about non-negative integers, but no property names
* Define the NonNegativeIntegerRange class that has two properties "start" and "end" with NonNegativeInteger values

**Alternative 2:**
* Define "start" property that has description and metadata about non-negative integers
* Define "end" property that has the identical description and metadata about non-negative integers
* Define "count" property that has the same description and metadata about non-negative integers, (if a class with a count property exists)
* Define "quantity" property that has the same description and metadata about non-negative integers (if a shopping cart class calls the number of an item in the cart "quantity" instead of "count")

then

* Define the NonNegativeIntegerRange class that has two properties "start" and "end" with NonNegativeInteger values (same as alternative 1)

The same problem exists with the SemVer class - if you have a class with "productVersion" and "packageVersion" properties, you either define SemVer once as a class, or you duplicate the SemVer metadata in a new property file every time there is a class using a version with a different property name.  And if you create a "version" property file, then you are forever prohibited from creating new classes with a "version" property that isn't in SemVer format.

Binding property names to property values in multiple model files is not a scalable modeling approach.

goneall:
I think we can merge this as is and track a separate issue if we want to add additional files for the property descriptions.

goneall (approved):
Looks like these properties are 1 based, so PositiveIntegerRange is correct.

zvr:
@goneall the nomenclature is usually begin/end or start/stop.  The pair start/end seems strange (to me).

davaya:
> > Line counts start with 1, so in that case we would still need PositiveIntegerRange?
> 
> If both `byteRange` and `lineRange` start at 1, I would agree we only need `PositiveIntegerRange`

Wouldn't byteRange always start at zero?

goneall:
> the nomenclature is usually begin/end or start/stop. The pair start/end seems strange (to me).

No opinion on the name here.  @armintaenzertng - if you agree with @zvr please update, otherwise we can discuss.

armintaenzertng:
I have no strong feelings on this one. Changed "start" to "begin" now.

kestewart (approved):
Proposal looks reasonable.  Merging and we can discuss further issues as needed.

PR #219, opened 2023-04-20
Dataset: fix names and cardinalities

armintaenzertng:
I fixed two property names to match the ones in the Properties folder.
Also added `maxCount: 1` to both as it does not make sense to have multiple values for `sensitivePersonalInformation` (a boolean) or `confidentialityLevel`.


armintaenzertng:
Another thing: Is `anonymizationMethodUsed` meant to have no `maxCount`?

goneall (approved):

zvr (approved):

PR #229, opened 2023-04-25
Create sbomType.md

kestewart:
With the publication of https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf on April 21, 2023.  We have agreed on standard definitions of the types of SBOMs.   Adding this property to the SBOM class so this information can be shared between tools.


SBOMDocumentTypes.md
```diff
@@ -0,0 +1,30 @@
...
+## Metadata
+
+- name: SBOMDocumentTypes
+
+## Entries
+
+- design: SBOM of intended, planned software project or product with included components (some of which may not yet exist) for a new software artifact.
+- source: SBOM created directly from the development environment, source files, and included dependencies used to build an product artifact. 
+- build: SBOM generated as part of the process of building the software to create a releasable artifact (e.g., executable or package) from data such as source files, dependencies, built
+components, build process ephemeral data, and other SBOMs.
```
armintaenzertng, line 24:
```suggestion
- build: SBOM generated as part of the process of building the software to create a releasable artifact (e.g., executable or package) from data such as source files, dependencies, build
components, build process ephemeral data, and other SBOMs.
```
This is more of a question if this "t" is intentional or should be a "d".

SBOMDocumentTypes.md
```diff
@@ -0,0 +1,30 @@
...
+- deployed: SBOM provides an inventory of software that is present on a system. This may be an assembly of other SBOMs that combines analysis of configuration options, and examination
+of execution behavior in a (potentially simulated) deployment environment.
+- runtime: SBOM generated through instrumenting the system running the software, to capture only components present in the system, as well as external call-outs or dynamically loaded
+components. In some contexts, this may also be referred to as an “Instrumented” or “Dynamic” SBOM.
+- analyzed: SBOM generated through analysis of artifacts (e.g.,cexecutables, packages, containers, and virtual machine images) after its build. Such analysis generally requires a
```
armintaenzertng, line 29:
```suggestion
- analyzed: SBOM generated through analysis of artifacts (e.g., executables, packages, containers, and virtual machine images) after its build. Such analysis generally requires a
```

SBOMDocumentTypes.md
armintaenzertng, line 1: This should go into the `model/Software/Vocabularies` folder.
Also, the name should be singular (in the file name as well as all references to it).

armintaenzertng (commented):
Thanks for the addition! :)
I spotted two major issues, see below.

SBOMDocumentTypes.md
```diff
@@ -0,0 +1,30 @@
...
+
+# SBOMDocumentTypes
+
+## Summary
+
+Provides a set of values to be used to describe the common types of SBOMs that tools may create.
+
+## Description
+
+This field provides a list with agreed on definitions of SBOM types, as defined in [Types of Software Bill of Material (SBOM) Documents](https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf), published on April 21, 2023.
```
rnjudge, line 11: If the field name is singular per @armintaenzertng's comment then why is this field described as a list? Shouldn't it be only a single value (from the list defined in https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf?).

SBOMDocumentTypes.md
rnjudge: If the field name is singular then why is this field described as a list? Shouldn't it be only a single value (from the list defined in https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf)? If this field is meant to be a list of options to fill in the `sbomType` field then I would suggest the name to stay plural for clarity.

SBOMDocumentTypes.md
rnjudge: In either case, `build` or `built` doesn't make much sense as a single word in this sentence... "From data such as build?" Seems like it's missing a word (build logs?)

SBOMDocumentTypes.md
```diff
@@ -0,0 +1,30 @@
...
+# SBOMDocumentTypes
+
+## Summary
+
+Provides a set of values to be used to describe the common types of SBOMs that tools may create.
+
+## Description
+
+This field provides a list with agreed on definitions of SBOM types, as defined in [Types of Software Bill of Material (SBOM) Documents](https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf), published on April 21, 2023. 
+This field is a reasonable estimate of the most likely type of an SBOM from the producer and consumer perspective from which both parties can draw conclusions
```
rnjudge, line 12: The description "reasonable estimate of the most likely type of an SBOM from the **producer and consumer perspective**"  is inconsistent with the `sbomType` field below which says: "This field is a reasonable estimation of the type of SBOM created from a **creator perspective**"

SBOMDocumentTypes.md
kestewart: @zvr - Plural seems more natural to me for a list, but please weigh in on how you want it.

SBOMDocumentTypes.md
kestewart: @rnjudge - it was a bad line split.    Should be "built components"    Want to retain it as is, as that's what in the publication.

SBOMDocumentTypes.md
kestewart: @armintaenzertng - thanks for spotting the directory issue.   Was moving too fast yesterday it seems.   Fixed.    Pleases re-review.

armintaenzertng: For clarification: This is a vocabulary that lists all possible types of SBOM; but it is not a list in the sense of a property of a class having multiple values (the latter is specified by there being no `maxCount` on `sbomType`).
See for example https://github.com/spdx/spdx-3-model/tree/main/model/Core/Vocabularies where all Vocabularies are singular even though they list multiple entries each.

model/Software/Vocabularies/SBOMDocumentTypes.md
```diff
@@ -0,0 +1,25 @@
...
+
+# SBOMDocumentTypes
+
+## Summary
+
+Provides a set of values to be used to describe the common types of SBOMs that tools may create.
+
+## Description
+
+This list of SBOM types with definitions, as defined in [Types of Software Bill of Material (SBOM) Documents](https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf), published on April 21, 2023.
```
armintaenzertng, line 11: This sentence is not finished. Maybe a "The" at the beginning?
I like the word "set" used in the summary, that maybe clarifies the issue that this is not a list in the sense of properties having multiple values.

model/Software/Vocabularies/SBOMDocumentTypes.md
```diff
@@ -0,0 +1,25 @@
...
+# SBOMDocumentTypes
+
+## Summary
+
+Provides a set of values to be used to describe the common types of SBOMs that tools may create.
+
+## Description
+
+This list of SBOM types with definitions, as defined in [Types of Software Bill of Material (SBOM) Documents](https://www.cisa.gov/sites/default/files/2023-04/sbom-types-document-508c.pdf), published on April 21, 2023. 
+The SBOM types describe the most likely type of an SBOM from the producer perspective, so that consumers can draw conclusions about the data inside an SBOM.  A single SBOM can have multiple SBOM document types associated with it.
```
armintaenzertng, line 12:
```suggestion
An SBOM type describes the most likely type of an SBOM from the producer perspective, so that consumers can draw conclusions about the data inside an SBOM.  A single SBOM can have multiple SBOM document types associated with it.
```
This hopefully makes it even clearer.
Maybe even consider omitting the "most likely" as there are probably not several "most" likely types?

model/Software/Vocabularies/SBOMDocumentTypes.md
kestewart: @armintaenzertng - have adjusted the sentence,  please re-review

kestewart:
@armintaenzertng , @rnjudge  - please review and go ahead an merge if you 're good with this latest version.

maxhbr (approved):
LGTM

tsteenbe:
Commit history is super ugly are we planning to squash the commits?

model/Software/Vocabularies/SBOMDocumentType.md
```diff
@@ -0,0 +1,25 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SBOMDocumentType
```
iamwillbar, line 3: Recommend removing the word document because "bill of materials" implies document.

zvr:
Change SBOM-Document-Type to SBOM-Type in the vocabulary name

iamwillbar (changes requested):
Approved pending removal of word "document".

Issue #227, opened 2023-04-24
Dataset profile: Cardinalities


armintaenzertng:
The current [dataset profile](https://github.com/spdx/spdx-3-model/blob/main/model/Dataset/Classes/Dataset.md) seems inconsistent in its use of cardinalities.

`anonymizationMethodUsed` is the only property with no `maxCount`. @maxhbr thinks this is a good idea to allow for multiple methods listed here. Most of the properties are free form text, so in principal you don't need multiple entries. In examples like this, though, it might be better to split different methods into different entries of this property.

On the other hand, for example `knownBiases` and `dataPreprocessingSteps` both have `maxCount = 1`, even if their property names are plural (see also #226 for the question whether they should be converted to singular), but I'm not sure that they constitute different cases to the above concerning their cardinality.

I already opened PR #219 to fix cardinalities for properties where having more than one value does not make sense.

armintaenzertng:
This basically also applies to the AI profile. I'm not sure if all cardinalities there are set as intended.

kestewart:
@rgopikrishnan91 - can you please review/comment.

maxhbr:
I think https://github.com/spdx/spdx-3-model/pull/230 is closed, and that should fix this ticket, right?

rgopikrishnan91:
@maxhbr I think we can convert these properties to have singular names, however, I think @goneall suggestion of removing the maxcount should be used here. Also, in AI profile, modelExplainability and modelDataPreprocessing both having maxcount removed might also be a good idea (should I reopen[ #230](https://github.com/spdx/spdx-3-model/pull/230))?

rgopikrishnan91:
@maxhbr I think we can convert these properties to have singular names, however, I think @goneall suggestion of removing the maxcount should be used here. Also, in AI profile, modelExplainability and modelDataPreprocessing both having maxcount removed might also be a good idea (should I reopen[ #230](https://github.com/spdx/spdx-3-model/pull/230))?

maxhbr:
No, do not reopen this PR. Instead you can propose new changes in a new PR.

kestewart:
Updated cardinalities merged for Dataset and AI.  Closing this.

Issue #226, opened 2023-04-24
Property names: Singular vs Plural


armintaenzertng:
I'm not sure if this was discussed but I can't remember a decision:
Do we want to keep property names consistently to singular?

Some currently plural property names: [externalIdentifiers.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Properties/externalIdentifiers.md), [externalReferences.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Properties/externalReferences.md), [imports.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Properties/imports.md), various [properties in the Dataset profile](https://github.com/spdx/spdx-3-model/tree/main/model/Dataset/Properties).

Possible problem: [namespace.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Properties/namespace.md) vs [namespaces.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Properties/namespaces.md). Maybe find a better word for the latter?

goneall:
It was discussed and decided on singular in the April 4th tech meeting - see the serialization updates in the [notes](https://github.com/spdx/meetings/blob/main/tech/2023-04-04.md#notes).

kestewart:
All tagged references are merged.   Closing.

bact:
> It was discussed and decided on singular in the April 4th tech meeting - see the serialization updates in the [notes](https://github.com/spdx/meetings/blob/main/tech/2023-04-04.md#notes).

Update the tech meeting link: https://github.com/spdx/meetings/blob/main/tech/2023/2023-04-04.md

PR #230, opened 2023-04-25
AI Profile: Use singular for properties

armintaenzertng:
According to #226, property names should be kept singular. I changed this for the AI profile.

`modelDataPreprocessingSteps` and `modelExplainabilityMechanisms` made this approach slightly awkward, as they both have `maxCount = 1` but can list multiple "values" (see also #227). I have a proposal to circumvent this issue by removing the "Steps" and "Mechanisms", respectively. Feel free to disagree.


goneall (approved):
LGTM - Thanks @armintaenzertng

kestewart:
@rgopikrishnan91 - please comment.  I'm concerned this is significantly changing the semantics.

rgopikrishnan91:
I think data preprocessing in particular can be multiple steps, we need that property element to be able to represent and denote that. For instance, the data can first have outliers removed, normalized and then embedded. So even though it can be written in one line of text, these are three distinct steps.

goneall:
@rgopikrishnan91 Could we have each step be a separate string and remove the maxCount?

maxhbr:
> I think data preprocessing in particular can be multiple steps, we need that property element to be able to represent and denote that. For instance, the data can first have outliers removed, normalized and then embedded. So even though it can be written in one line of text, these are three distinct steps.

I think this could be a follow-up PR.

maxhbr:
@armintaenzertng : I rebased this PR

maxhbr (approved):
LGTM

rgopikrishnan91:
@goneall I think I am okay with it being multiple strings and each step (or any two logically) being in different strings. Same for the model explainability mechanism

rgopikrishnan91:
@kestewart what do you think?

PR #231, opened 2023-04-25
Core: Use singular for properties

armintaenzertng:
According to https://github.com/spdx/spdx-3-model/issues/226, property names should be kept singular. I changed this for Core.

Two issues remain:
- There are both `namespace` and `namespaces`. Maybe we can find a better word for the latter as it is actually a `namespaceMap`?
- `imports` would become `import`, which would be a keyword in many programming languages, for example in Python. That is not a good idea. As this is an external map, maybe `importMap` might be an idea?

### DIFF
--- a/model/Core/Classes/Element.md
+++ b/model/Core/Classes/Element.md
@@ -45,10 +45,10 @@ and inter-relatable content objects.
   - maxCount: 1
 - verifiedUsing
   - type: IntegrityMethod
-- externalReferences
+- externalReference
   - type: ExternalReference
-- externalIdentifiers
+- externalIdentifier
   - type: ExternalIdentifier
-- extensions
+- extension
   - type: Extension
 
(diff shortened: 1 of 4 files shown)


model/Core/Classes/Element.md
```diff
@@ -45,10 +45,10 @@ and inter-relatable content objects.
   - maxCount: 1
 - verifiedUsing
   - type: IntegrityMethod
-- externalReferences
+- externalReference
```
goneall, line 48: This is called `externalRef` in SPDX 2.X.  Although there has been an argument that `externalReference` is clearer, I don't think that justification is enough to cause migration effort.  Note that the choice of `externalRef` as the name in 2.X was chosen after a discussion and comparison to other terms such as `externalReference`.  I added issue #234 to track renaming this back to the 2.3 value.

armintaenzertng: I'd rather wait until the issue is resolved before going ahead and already changing it prior to consensus.

goneall (approved):
LGTM

kestewart (approved):
LGTM

PR #232, opened 2023-04-25
Dataset profile: Use singular for properties

armintaenzertng:
According to https://github.com/spdx/spdx-3-model/issues/226, property names should be kept singular. I changed this for the Dataset profile.

As in #230, I removed the "Steps" from `dataPreprocessingSteps` to circumvent the issue.


kestewart (approved):
LGTM - Thanks Armin

PR #233, opened 2023-04-25
Build profile: use slash notation to reference classes from Core

armintaenzertng:
As is used for example [here](https://github.com/spdx/spdx-3-model/blob/main/model/Software/Classes/Package.md).


kestewart (approved):
Thanks for the fix!

PR #235, opened 2023-04-26
Fix CI issues

armintaenzertng:
The CI is finally green again! :)

In the future, we should not ignore failing runs anymore, as they can actually provide helpful insight into what might be wrong. See for example the following code snippet taken from the run [here](https://github.com/spdx/spdx-3-model/actions/runs/4804376621/jobs/8549800261) (under the `run spec-parser` step):
```
 ERROR   : ../model/AI/Classes/AIPackage.md:Ln 17,Col 2: Error: Invalid metadata key `Instatiability`.
	 Line no 17: '- Instatiability: Concrete\n\n'
ERROR   : ../model/AI/Classes/AIPackage.md:Ln 31,Col 2: Syntax Error: Expected `<key>: <values>`
	 Line no 31: '  - maxCount:1\n'
ERROR   : Unable to parse `Class` markdown: '../model/AI/Classes/AIPackage.md'
ERROR   : ../model/AI/Properties/modelExplainabilityMechanisms.md:Ln 1,Col 1: empty
	 Error: Missing `SPDX-License-Identifier: <value>`
ERROR   : Unable to parse `Property` markdown: '../model/AI/Properties/modelExplainabilityMechanisms.md'
```
These issues have been fixed, and I also changed the output name `tst.ttl` to `model.ttl`, a change that has been recently introduced into the `spec-parser`.


.github/workflows/spec-parser.yml
```diff
@@ -1,6 +1,11 @@
 name: run spec-parser
 
-on: [push]
+on:
+  push:
+    branches:
+      - main
```
maxhbr, line 6: since we are developing on feature branches. It might also be helpful to also run on pushes to them. But we can decide that later and should merge this first.

maxhbr (approved):
LGTM

zvr (approved):

PR #236, opened 2023-04-26
Glossary

zvr:
- Adds Glossary of Terms


Glossary.md
```diff
@@ -0,0 +1,74 @@
...
+
+An explicit specification of concepts, terms, categories, relationships, etc. for a particular scoped domain forming a language or basis for communication and analysis of information within the scoped domain. SPDX 3.0 is an ontology targeted to support the scoped domain of bills of material.
+
+## Class
+
+A represention of a scope/set of individual instances of a particular “concept” (e.g., File, Person, ExternalReference, etc.). Each individual instance of a class has an IRI ID and is asserted as a member of a particular class via a type statement.
+
+## Domain Class
+
+A class representing a concept of primary focus within a given domain. In a graph or serialized set of instance content this would be the granularity of what would be a node in the graph or a top-level object in the serialization set. This is typically similar to the scoping seen in labeled property graphs. In SPDX 3.0, all domain classes are subclasses of Element.
```
goneall, line 17: > In SPDX 3.0, all domain classes are subclasses of Element.

In the licensing profile, License will be a domain class at as currently proposed will not be a subclass of Element.  Suggest removing this sentence.

Glossary.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Non-domain Class
+
+A class representing a concept of secondary or tertiary focus within a given domain that is a characterization of some aspect of a domain class. When structuring the graph for a domain use case it is desirable and more practical for the domain graph nodes to only consist of domain class instances and object properties. Any non-domain classes and datatype properties would be considered internal adornments on the domain objects. In SPDX 3.0, all defined classes that are not subclasses of Element are non-domain classes.
```
goneall, line 21: Same comment as above, we either need to make License a subclass of Element OR remove the statement that all domain classes are subclasses of Element.

goneall (changes requested):
In reviewing the pull request, I realized we have an inconsistency between the glossary and the proposed classes in the licensing profile.  I would recommend removing the inconsistent statements from the glossary as are more statements of policy than definition of terms.  We can decide the policies and document them separately.

Otherwise, it all looks good and I really appreciate the work that went into the precise definitions I think it will really help in understanding and adoption.

Glossary.md
zvr: @goneall  I completely agree (and had made the remark before).

But maybe we can say that in the Core profile this holds?

goneall: @zvr I guess so - my preference is to remove the statement, but it would not be inconsistent to qualify the statement `In the SPDX 3.0 core profile, all domain classes are subclasses of Element.`

goneall (approved):
Thanks @zvr

Glossary.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Datatype property
+
+A representation of a relationship between a class (definition or instance) and some literal (string, integer, boolean, etc.) descriptive characterization of the class.
+
+## Object property
+
+A representation of a relationship between a class (definition or instance) and some conceptual (class-based) descriptive characterization of the class. In other words, it is a relationship between one class and another class.
+
+## Domain
```
kestewart, line 35: Thiere are two definitions of "Domain"

kestewart (changes requested):
Can this be sorted alphabettically, and make sure we have no different  definitions on the same term (see "Domain"

Glossary.md
zvr: that is correct! The second one is definitely an established term ("domain" and "range" of a property). The first one, I assume, is @sbarnum's attempt to talk about "area of interest" or something.
I agree that we should not have both.

zvr:
Actually I think that the order is OK as it is, if people read the whole list:
"Class", there are two types, "Domain Class", then "Non-Domain Class".

Essentially, the definitions build on previous ones. If we want to have them dictionary-style, in alphabetic order, we have to make sure that each one works standalone.

Glossary.md
kestewart: Can you please update the PR to remove it then.

kestewart (changes requested):

kestewart:
@zvr - Each item in the Glossary should be standalone.  Can we rename the file to be guidance or something similar, until each entry is standalone.

kestewart:
@zvr - to add in property for domain and range - to deconflict.

zvr:
Change the second Domain and Range to "Property Domain" and "Property Range"

zvr:
Opened a new issue #253 for re-wording and alphabetizing the definitions. Not blocking for release.

kestewart (approved):
Thanks for clarifying.  LGTM.

PR #238, opened 2023-04-27
serialization

zvr:
- Adds initial files about serializarion


goneall (approved):
Looks like a good structure for collaboration

PR #241, opened 2023-04-28
Adds DateTime data type in Core

### LINKED ISSUE #215: Adding DateTime class in core
I propose we add a class named `DateTime` or `Timestamp` to Core and change all properties that currently have range  `xsd:dateTime` to have this new class as range.

The new class will be a subclass of `xsd:string` (like `SemVer`, for example), but will also have a shape constraint that the format is `yyyymmddThhmmssZ`.

This will allow us to keep the same format as used in SPDXv2.

If we don't do this:

- we either have to add the format shape restriction to every property (and there are a number of them in the different profiles)
- or we don't restrict the format, which forces all implementations to have a full parser for all variants of ISO-8601 that are allowed by `xsd:dateTime`.

If there is agreement, I can submit the PR to implement this.

zvr:
The type might as well be called `Timestamp`,
but `DateTime` is already well-known from xsd.

Closes #215

Signed-off-by: Alexios Zavras (zvr) <github@zvr.gr>

### DIFF
--- a/model/Core/Classes/DateTime.md
+++ b/model/Core/Classes/DateTime.md
@@ -1,0 +1,25 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# DateTime
+
+## Summary
+
+A string representing a specific date and time.
+
+## Description
+
+A Datetime is a string representation of a specific date and time.
+It has resolution of seconds and is always expressed in UTC timezone.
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

kestewart (approved):

PR #244, opened 2023-04-28
Fix cardinality of relationshipType

meretp:
According to the png relationshipType is mandatory.


zvr (approved):
Ooops!

kestewart (approved):
Good catch Meret!   Thanks

PR #240, opened 2023-04-28
Adds new properties in Core/ExternalIentifier

zvr:
As decided in the 2023-04-25 Tech Team call.

Signed-off-by: Alexios Zavras (zvr) <github@zvr.gr>

PR #245, opened 2023-04-28
Add identifier property

meretp:
The markdown file for the identifier property was still missing in core.


zvr (approved):

zvr:
Just saw that and wanted to add it to #240, but since you were faster...

kestewart (approved):
Looks good to me.

Issue #56, opened 2023-01-01
Serialization: lists of constants


davaya:
The element "profile" property is a list of profile names, e.g., "core", "software", etc.

The conventional way of serializing a list of string values is as a list:
```
  "profile": ["core", "software"]
```
It has been suggested to instead require all list items to be serialized as objects with repeated constant property names:
```
  "profile": [
    {
      "name": "core"
    },
    {
      "name": "software"
    }
  ]
```
**Proposal:** Serialize lists with their actual values (e.g., IRIs, enumerations such as profile name, SoftwarePurpose, etc. rather than injecting an artificial object wrapper around those values.

Rationale: an object wrapper makes serialized data both more verbose and more difficult to read.  But more importantly, the extra data provides no information - the receiver knows nothing more after seeing it than it does without seeing it.  Proof: a receiver, after receiving the native list, has everything it needs to locally transform it into the list of objects if that format is needed for some reason.

**Decision:** ?

maxhbr:
Although I would prefer the "conventional way", profiles are expected to get versioned, which is possible with objects without a breaking change:
```
  "profile": [
    {
      "name": "core",
      "version": "3.0.1"
    }
  ]
```

davaya:
Every Element has its own specVersion; I'm not sure what applications would be expected to do with specVersion 3.2.0, profile core 3.1.2, software 3.0.5.

Of course lists are a general case, not just versioning.  But if there were a use case for separately-versioned profiles they could be  serialized as:
```json
  "profile": {
    "core": "3.1.2",
    "software:": "3.0.5"
  }
```

The general case is described in #55 - do you serialize email headers with lists of repeated [ {name: x1, value: y1}, {name: x2, value: y2}, {...} ] pairs or with normal object properties {x1: y1, x2: y2, ...}?

maxhbr:
> Of course lists are a general case, not just versioning. But if there were a use case for separately-versioned profiles they could be serialized as:

yes, that would be possible. But that would still be a breaking change to your initial proposal. So, this does not help to avoid the breaking change in the "there might or might not be a future with individual versions"-Problem.

maxhbr:
> Every Element has its own specVersion; I'm not sure what applications would be expected to do with specVersion 3.2.0, profile core 3.1.2, software 3.0.5.

+1 from me, I also think that this is not a future to hope for. Maybe we can decide that it will not come. This would solve this thread.

davaya:
[...]

PR #249, opened 2023-04-30
Create ProfileIdentifierType.md

kestewart:
An arbitrary string for profile identifier, is going to cause too many snags, with arbitrary matching and capitalization.     Recommend restricting this down to the agreed on set of formal names for profiles.


kestewart:
Note:  this should address https://github.com/spdx/spdx-3-model/issues/56

zvr (approved):

maxhbr (approved):
LGTM

Issue #47, opened 2022-11-14
Proposal: move Extensions to their own profile


maxhbr:
I would like to move the `extensions` property of elements to its own `Extensions` object in a new profile.

![20221114_07h59m28s_grim](https://user-images.githubusercontent.com/1187050/201598129-3f259bb1-b54d-4dd6-b7d3-e77c1da6e680.png)

This has several advantages:
* a consumer can on a profile level decide, if he wants to support extensions
* a extension can have more fields, which for example point to documentation or can have comments, ...
* the extensions do not loose any functionality (one can still add extensions to arbitrary elements), but still can be separated from core discussions

tsteenbe:
+1 on exploring moving extensions to its own profile - consuming SPDX is already had enough for tools and know of certain user that won't want non-specified fields in their SBOMs.

maxhbr:
ping @iamwillbar

jeff-schutt:
+1 This seems to allow the flexibility desired while reducing potential for errors in tools. I like that this is similar to how other classes and properties will be interpreted by tools: by first interpreting what profiles are in place via the CreationInformation.

goneall:
+1 here too - same reasons as @jeff-schutt

davaya:
+1 for putting extensions in their own namespace.  This allows various conformance targets to specify one or more namespaces as optional or required.

kestewart:
There seems to be agreement in this thread on having extensions in their own namespace.   @davaya - can you work up a pull requiest.   @iamwillbar - please let us know if you disagree.

maxhbr:
I think this could be done in two PRs. A first one, that just drops extensions (and thus is removing it from the critical path) and another one, that adds them back in.

kestewart:
@maxhbr - if you want to work up the PR, and then tag @davaya for review,  that might be faster.   Thanks!

kestewart:
Given the discussion with Sean on the May 2nd call - and the wider optics.  Hold off on drop and re-add.  Instead focus on getting definition from Sean, and updates.

kestewart:
Moving this to be part of the 3.0 release as it's going to take more discusion.

maxhbr:
I think that one of the reasons for extensions was, that profiles might use or need them. Until now, no profile thought of using them. I think that at least that reason can be crossed of the list.

goneall:
@sbarnum will attempt to document this as a separate extension profile - if it ends up being difficult to implement, we will revisit.

[...]

PR #252, opened 2023-05-02
Drop extensions from Core (see #47)

maxhbr:
this is the first part of #47, as agreed in https://github.com/spdx/spdx-3-model/issues/47#issuecomment-1529039618


kestewart:
Closing this for now,   will reopen after we have circled back from Sean.

Issue #251, opened 2023-05-02
Adding more types to ExternalReference


tsteenbe:
Whilst working to improve the SPDX SBOM generation in [ORT](https://github.com/oss-review-toolkit/ort) we had a discussion why we generate multiple packages in SPDX versus a single one in our CycloneDX. The reason was that in SPDX 2.3 to encode in SPDX that a Java package has a binary artifact, sources artifact and source code repository requires 3 SPDX package and several relationship such as GENERATES and DISTRIBUTION_ARTIFACT. This needlessly bloats the SPDX file if you just want to exchange information about include JAR but also want to convey additional information the SBOM tool collected.

Currently the ExternalReferenceType in 3.0 model is defined as

- altDownloadLocation: A reference to an alternative download location.
- altWebPage: A reference to an alternative web page.
- other: Used when the type doesn't match any of the other options.
- securityAdvisory: A reference to the published security advisory (where advisory as defined per ISO 29147:2018). It may contain an impact statement whether a package (e.g. a product) is or is not affected by vulnerabilities.
- securityFix: A reference to the source code with a fix for the vulnerability (e.g., a GitHub commit).
- securityOther: Used when the reference is security related but doesn't match any of the other types.

Proposal is to extend ExternalReferenceType with the below types.
Doing so will enable:
1.  SPDX creators or tools to easily include more data they know about an element (without having to any clean-up).  This data can then be amended by other tools.
2.  better back and forward conversion with CycloneDX

-  **api-advisories**: A reference to API service where up-to-date security advisory statements related to artifacts can be retrieved
-  **api-vex**: A reference to API service where up-to-date VEX statements related to artifacts can be retrieved
-  **bom**: A reference to the bill of materials information related to an artifact / / Use case Eclipse foundation publishing SPDX SBOMs for all their projects in a central location
-  **binary-artifact**:  A reference to binary artifacts related to a package
-  **build-meta**: A reference build metadata related to a published package
-  **build-system**: A reference build system used to create or publish the package
-  **source-artifact**:  A reference to an artifact containing the source for a package
-  **chat**: A reference to the instant messaging system used by the maintainer for a package
-  **documentation**: A reference to the documentation for a package
-  **funding**: A reference to funding informaton related to a package
-  **issue-tracker**: A reference to the issue tracker for a package
-  **mailing-list**:  A reference to the mailing list used by the maintainer for a package
-  **metrics**:  A reference to metrics related to package such as OpenSSF scorecards.
[...]

PR #255, opened 2023-05-02
Add new types to ExternalReferenceType

tsteenbe:
Provide SPDX creators an option to easily share any (raw) info collected which may be useful or can be clean-up by downstream users.

Relates-to: #251


kestewart (changes requested):
These look like reasonable starting points.    Can the syntax be changed to be camel case, rather than dashes to match the others.

maxhbr (commented):
I changed it to camelCase

maxhbr (approved):
Changes look good to me and are a good next step

PR #248, opened 2023-04-29
Security Profile: Add Relationship Types

puerco:
This PR adds the new relationship types required by the security profile.

The first commit adds the VEX relationship types required by `VulnerabilityVexAssessmentRelationship`:

 - affected
 - notAffected
 - fixed
 - underInvestigation

**HOLD** until the final Security profile is in 

/cc @tsteenbe @rnjudge @jeff-schutt @karsten-klein @goneall

PR #258, opened 2023-05-03
Update VEX relationships to even their directions and add prepositions

puerco:
This commit modifies the proposed VEX relationships to make them all work in the same direction: from vulnerability to (VEX) product addressing [feedback from the tech call on May 2 2023](https://github.com/spdx/spdx-3-model/pull/248#issuecomment-1531761847).

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -66,7 +66,7 @@ Build Profile specific RelationshipType descriptions can be found [here](https:/
 - buildInvokedBy: Agent that invoked the build
 - buildOnBehalfOf: Actor for which buildInvokedBy is acting on behalf of
 - buildHostOf: Element in which the build instance runs on
-- affected: (Security/VEX) Designates an elemented as affected by a vulnerability
-- notAffected: (Security/VEX) Specifies a vulnerability has no impact on an element
-- fixed: (Security/VEX) A vulnerability has been fixed in an element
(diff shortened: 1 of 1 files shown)


model/Core/Vocabularies/RelationshipType.md
```diff
@@ -66,7 +66,7 @@ Build Profile specific RelationshipType descriptions can be found [here](https:/
 - buildInvokedBy: Agent that invoked the build
 - buildOnBehalfOf: Actor for which buildInvokedBy is acting on behalf of
 - buildHostOf: Element in which the build instance runs on
-- affected: (Security/VEX) Designates an elemented as affected by a vulnerability
-- notAffected: (Security/VEX) Specifies a vulnerability has no impact on an element
-- fixed: (Security/VEX) A vulnerability has been fixed in an element
-- underInvestigation: (Security/VEX) The impact of a vulnerability is being investigated
+- affects: (Security/VEX) Designates an elemented as affected by a vulnerability
```
maxhbr, line 69: This direction can also be used, to describe multiple affected versions or artifacts at the same time.

puerco: Thanks Max good catch :pray: 
I've updated the descriptions to call for one or more elements.

kestewart (approved):
Thanks.  This matches what was discussed in the May 2nd meeting, and subsequent comments.

PR #260, opened 2023-05-03
Add "Agent" entry to Glossary.md

puerco:
This commit adds a definition of _agent_ to the glossary (trying my best to capture Sean's definition)

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>


kestewart (approved):
This matches my understanding.   If Sean disagrees, we can adjust further.

zvr (approved):

PR #262, opened 2023-05-04
Create datasetType

rgopikrishnan91:
Adding a missing required field

@kestewart please review


kestewart (approved):
Matches what was discuss.

maxhbr (approved):
Looks good, and a follow-up PR should also use that new property

PR #264, opened 2023-05-04
add missing properties and fix some references

meretp:
I noticed some missing properties and wrong references while generating the ttl file with the `spec-parser`.


kestewart (approved):
Thanks for getting the parsing working again.  

Next step is we'll need to fill in the descriptions here.

PR #267, opened 2023-05-04
Licensing: Rename LicenseExpression to AnyLicenseInfo

### LINKED ISSUE #222: Rename LicenseExpression to AnyLicenseInfo
AnyLicenseInfo is used in the [SPDX 2.3 UML class diagram](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/model/SPDX-UML-Class-Diagram.jpg).  It is also currently used in some of the tooling.

This is an abstract class, so it doesn't impact the serializations.

swinslow:
Fixes #222 


kestewart (approved):
Understand the background.    Change looks fine for now.

PR #270, opened 2023-05-04
Add start and end time to Relationships

tsteenbe:
As was agreed in SPDX Technical Team Meeting of 2023-04-25.

### DIFF
--- a/model/Core/Classes/Relationship.md
+++ b/model/Core/Classes/Relationship.md
@@ -36,4 +36,11 @@ that one Element is related to one or more other Elements in some way.
   - type: RelationshipCompleteness
   - minCount: 0
   - maxCount: 1
-
+- startTime
+  - type: /Core/DateTime
+  - minCount: 0
+  - maxCount: 1
+- endTime
+  - type: /Core/DateTime
+  - minCount: 0
+  - maxCount: 1
--- a/model/Core/Classes/Relationship.md
+++ b/model/Core/Classes/Relationship.md
@@ -37,10 +37,10 @@ that one Element is related to one or more other Elements in some way.
   - minCount: 0
   - maxCount: 1
 - startTime
-  - type: /Core/DateTime
+  - type: DateTime
   - minCount: 0
   - maxCount: 1
 - endTime
-  - type: /Core/DateTime
+  - type: DateTime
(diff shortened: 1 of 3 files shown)


zvr (changes requested):
Agree, but all types should be `DateTime` (already in Core)

tsteenbe:
@zvr  I fixed up the `xsd:dataTime` to `xsd:dateTime` in Relationship.md. Hope I get you right that `type: xsd:dateTime` should be `type: DateTime`?

I just copied what was used elsewhere, if everything need to be `type: DateTime` now I can fill a pull request to fix this across the entire specification.

tsteenbe:
Slightly related, should the `created` property in Core not be renamed to `createdTime` so consistently across the spec all properties related to time will end in `Time`?

tsteenbe:
Updated this PR to be aligned with @zvr  https://github.com/spdx/spdx-3-model/pull/278.

model/Core/Properties/endTime.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# startTime
```
maxhbr, line 3:
```suggestion
# endTime
```

maxhbr (changes requested):

model/Core/Classes/Relationship.md
```diff
@@ -36,4 +36,11 @@ that one Element is related to one or more other Elements in some way.
   - type: RelationshipCompleteness
   - minCount: 0
   - maxCount: 1
-
+- startTime
+  - type: /Core/DateTime
```
zvr, line 40: No need for "/Core" here, this is in Core anyway.

zvr (changes requested):

maxhbr (approved):
It does what it tries to do in a correct way. But I still think that not every relationship should have this feature. Would be good to fix that as a follow up.

I think that most generated documents describe some snapshot information and such time stamps add an orthogonal dimension that will create a lot of problems, inconsistencies and surprises.

kestewart (approved):
This may require further discussion but merging this for now, so we can consider wholistically.

PR #273, opened 2023-05-05
Fixup various Element properties

tsteenbe:
Made CreationInfo optional to prevent SBOM bloating see for example https://gist.github.com/meretp/561e4ea963122f5811a147ea056e4d84#file-example-jsonld-L29-L75. CreationInfo should only be mandatory on certain elements such as SPDXDocument, Bundle, etc. 

IMO SPDX should enable linked data format such as JSON-LD and RDF but it must not make it mandatory for JSON and YAML formats. Often heard complaint about SPDX 2.x was that it was too complex/bloated, we should not make the same mistake in SPDX 3.0 but providing something simple for those that just want to exchange data and enable linked data SBOM for parties that want to graph reasoning.

Ping @maxhbr  @jeff-schutt  @puerco - please add your opinion from our discussions here as well.

### DIFF
--- a/model/Core/Classes/Element.md
+++ b/model/Core/Classes/Element.md
@@ -47,6 +47,7 @@ and inter-relatable content objects.
   - type: IntegrityMethod
 - externalReference
   - type: ExternalReference
+  - minCount: 0
 - externalIdentifier
   - type: ExternalIdentifier
 - extension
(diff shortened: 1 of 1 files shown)


tsteenbe:
> `minCount: 0` is the default for everything.
> 
> Why do you think there can only be a single externalReference, a single ExternalIdentifier, and a single Extension? This does not match the model as discussed till now...

@zvr  You right, made a copy-paste mistake only minCount should be there an no maxCounts, fixed it.

kestewart (approved):

PR #274, opened 2023-05-05
Create suppliedBy

rgopikrishnan91:
@kestewart I am creating this PR as we discussed yesterday on the call

### DIFF
--- a/model/Core/Properties/suppliedBy
+++ b/model/Core/Properties/suppliedBy
@@ -1,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# suppliedBy
+
+## Summary
+
+Identifies who or what supplied the Element.
+
+## Description
+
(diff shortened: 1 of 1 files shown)


model/Core/Properties/suppliedBy
maxhbr, line 1: Where is this used? can you add the property to the relevant class so that it is reviewable and understandable?

maxhbr (changes requested):

model/Core/Properties/suppliedBy
kestewart: This is a mandatory field in the AI & Dataset profiles. 

Discussed this on the meeting with Gary, Thomas, Rose, & Adolfo right now, and referencing the decisions made in the meetings, we're adding back a supplier as a property in the core.     Gopi will add external property into both profiles.

kestewart (approved):
Thanks for putting this together

PR #275, opened 2023-05-05
Update AIPackage.md

rgopikrishnan91:
Adds external references and fixes applied that addresses the comments in Issue#227 

@kestewart Please review

### DIFF
--- a/model/AI/Classes/AIPackage.md
+++ b/model/AI/Classes/AIPackage.md
@@ -46,11 +46,9 @@ Metadata information that can be added to a package to describe an AI applicatio
 - modelDataPreprocessing
   - type: xsd:string
   - minCount: 0
-  - maxCount: 1
 - modelExplainability
   - type: xsd:string
   - minCount: 0
-  - maxCount: 1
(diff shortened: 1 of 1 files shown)


model/AI/Classes/AIPackage.md
```diff
@@ -72,3 +70,34 @@ Metadata information that can be added to a package to describe an AI applicatio
   - type: SafetyRiskAssessmentType
   - minCount: 0
   - maxCount: 1
+
+## External properties restrictions
+
+- /Core/Properties/createdBy
```
kestewart, line 76: This should be referring to use in a Class, rather than Properties.     Consistent info that needs to be resolved in the rest of these.

kestewart (changes requested):

kestewart (approved):
Thanks for fixing up the external properties, Max.

model/AI/Classes/AIPackage.md
kestewart: Thanks @maxhbr for fixing this up.

PR #276, opened 2023-05-05
Update Dataset.md

rgopikrishnan91:
Fixes what's discussed in #227 and the adds the external references

@kestewart Please review

### DIFF
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -89,7 +89,7 @@ Metadata information that can be added to a dataset that may be used in a softwa
   - minCount: 1
 - /Licensing/Properties/declaredLicense
   - minCount: 1
-- /Core/Properties/releaseTime
+- /Core/Artifact/releaseTime
   - minCount: 1
 - /Core/Properties/builtTime
   - minCount: 1
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -87,7 +87,7 @@ Metadata information that can be added to a dataset that may be used in a softwa
   - minCount: 0
 - /Licensing/Properties/concludedLicense
   - minCount: 1
-- /Licensing/Properties/declaredLicense
+- /Software/SoftwareArtifact/declaredLicense
   - minCount: 1
 - /Core/Artifact/releaseTime
   - minCount: 1
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -69,25 +69,13 @@ Metadata information that can be added to a dataset that may be used in a softwa
 
 ## External properties restrictions
 
-- /Core/Element/name
-  - minCount: 1
 - /Core/Artifact/originatedBy
   - minCount: 1
-- /Core/Artifact/suppliedBy
-  - minCount: 0
-- /Core/Hash
-  - minCount: 0
 - /Software/Package/downloadLocation
   - minCount: 1
-- /Software/Package/packageVersion
-  - minCount: 0
 - /Software/SoftwareArtifact/purpose
   - minCount: 1
-- /Core/Element/comment
-  - minCount: 0
 - /Core/Artifact/releaseTime
   - minCount: 1
 - /Core/Artifact/builtTime
   - minCount: 1
-- /Core/Artifact/validUntilTime
-  - minCount: 0
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -71,9 +71,9 @@ Metadata information that can be added to a dataset that may be used in a softwa
 
 - /Core/Element/name
   - minCount: 1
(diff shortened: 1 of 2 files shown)


model/Dataset/Classes/Dataset.md
```diff
@@ -68,3 +66,32 @@ Metadata information that can be added to a dataset that may be used in a softwa
...
   - maxCount: 1
+
+## External properties restrictions
+
+- /Core/Element/name
+  - minCount: 1
+- /Core/Properties/originatedBy
+  - minCount: 1
+- /Core/Properties/suppliedBy
+  - minCount: 0
```
maxhbr, line 77: this line is a no-op? Every entry with `- minCount: 0` can be dropped.

model/Dataset/Classes/Dataset.md
```diff
@@ -68,3 +66,32 @@ Metadata information that can be added to a dataset that may be used in a softwa
   - type: DatasetAvailabilityType
   - minCount: 0
   - maxCount: 1
+
+## External properties restrictions
+
+- /Core/Element/name
+  - minCount: 1
```
maxhbr, line 73: This is already set for `/Software/Package` and does not add a new restriction here. Can be removed.

model/Dataset/Classes/Dataset.md
```diff
@@ -68,3 +66,32 @@ Metadata information that can be added to a dataset that may be used in a softwa
...
+- /Software/Properties/purpose
+  - minCount: 1
+- /Core/Element/comment
+  - minCount: 0
+- /Licensing/Properties/concludedLicense
+  - minCount: 1
+- /Licensing/Properties/declaredLicense
+  - minCount: 1
+- /Core/Properties/releaseTime
+  - minCount: 1
```
maxhbr, line 93: So, only released Datasets can be described?

model/Dataset/Classes/Dataset.md
```diff
@@ -68,3 +66,32 @@ Metadata information that can be added to a dataset that may be used in a softwa
...
+  - minCount: 0
+- /Software/Package/downloadLocation
+  - minCount: 1
+- /Software/Package/packageVersion
+  - minCount: 0
+- /Software/Properties/purpose
+  - minCount: 1
+- /Core/Element/comment
+  - minCount: 0
+- /Licensing/Properties/concludedLicense
```
maxhbr, line 88: ConcludedLicense and licensing is not yet in core and this can not yet be merged. This also might be expressed as a relationship, so this approach might not work.

model/Dataset/Classes/Dataset.md
```diff
@@ -68,3 +66,32 @@ Metadata information that can be added to a dataset that may be used in a softwa
...
+  - minCount: 1
+- /Licensing/Properties/declaredLicense
```
maxhbr, line 90: same as ConcludedLicense

model/Dataset/Classes/Dataset.md
```diff
@@ -68,3 +66,32 @@ Metadata information that can be added to a dataset that may be used in a softwa
...
+  - minCount: 1
+- /Core/Properties/releaseTime
```
maxhbr, line 92: This references a property, but instead should reference a field of a parent class

maxhbr (changes requested):

model/Dataset/Classes/Dataset.md
kestewart: It helps make it clearer which properties, are mandatory for Dataset profile.   I don't see issue with leaving it hear for now until we have the auto generation of fields per profile working.

maxhbr: It might not come at 0 cost and might introduce more complexity into generated documents like the ttl and the SHACL. And it makes it harder to find the actual changes.

model/Dataset/Classes/Dataset.md
kestewart: As specified now,  we may need to revisit it being mandatory for streaming data.   Best to put in an issue for this so we don't overlook.

model/Dataset/Classes/Dataset.md
maxhbr: The issue is that it references a property.

model/Dataset/Classes/Dataset.md
maxhbr: The issue is that it references a property. And even with licensing profile merged, one can not add a constraint to a field that does not yet exist

model/Dataset/Classes/Dataset.md
maxhbr: The issue is that it references a property. And even with licensing profile merged, one can not add a constraint to a field that does not yet exist

model/Dataset/Classes/Dataset.md
kestewart: suggest the fix to remove it, and we'll revisit after the merge.

maxhbr (changes requested):

Issue #210, opened 2023-04-16
Licensing: Naming for License properties defined in other standards


swinslow:
This question follows on from @armintaenzertng's comment at https://github.com/spdx/spdx-3-model/pull/179#issue-1655679452. It relates to the thread at #167 and in particular @goneall's comment at https://github.com/spdx/spdx-3-model/issues/167#issuecomment-1497960149.

The [`comment` property for Licenses](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Licensing/Classes/License.md) is, I believe, one of those with semantics defined in other standards, like you described above. See https://www.w3.org/2000/01/rdf-schema#comment

Although its name is `comment`, if I understand correctly I believe that ties to the "Notes" field for licenses on the SPDX License List. I'm not certain that we are actually using the "Notes" field in a manner that ties into the RDF-defined meaning of "A description of the subject resource." Perhaps this should be changed to `licenseNotes`?

A similar situation applies for `seeAlso`, which is described as "Other web pages for this license" on the HTML License List pages and with the `<crossRef>` tags in the license list XML syntax. For this one, our usage on the License List does appear to be in line with the RDF-defined meaning of "Further information about the subject resource", but I note the open question about #167 as to overlapping property names across profiles.

Finally, `name` for Licenses is perhaps better to change to `licenseName` to avoid similar conflict issues, though I don't think there's the same question about terms defined in other standards.

@goneall in particular -- do you have thoughts on this? Should we keep `comment` and `seeAlso` with the current names in the SPDX 3.0 model in order to maintain the tie to the RDF terms, or would these be appropriate to change to `licenseNotes` and `licenseCrossRefs`?

goneall:
@swinslow You make some very good points.  We have tried to (re)use previous standards even when the semantics were close, but not exact.  In RDF, there is nice feature we can use which is [sub-property](https://www.w3.org/TR/rdf-schema/#ch_subpropertyof).

For the `comment` property of licenses, it doesn't map exactly to the `notes`.  The comments is related to the license comments in the spec.  BTW - there is an [issue in the SPDX spec](https://github.com/spdx/spdx-spec/issues/46) I added in 2017 to document properties of the license list itself.  If we were to act on that issue, we could add the notes in as suggested.

For `seeAlso`, that property pre-dates the introduction of `crossRef`.  We added `crossRef` for some of the reasons you mentioned.

[...]

PR #277, opened 2023-05-05
Licensing: Merge Licensing profile into main

swinslow:
cc @jlovejoy @kestewart @zvr 


swinslow:
@zvr @maxhbr @armintaenzertng Probably going to need some help here  :)

zvr:
I propose we wait to have the license-as-relationship stuff in the profile before we merge.  Look out for PR (towards the `licenssing-profile` branch) later today.

kestewart (approved):
Changes look good.  Not spotting any blockers.  We may want to revisit some of the cardinalities, but this should be a good starting point.   Nice work.

kestewart:
@zvr - there are some properties here, that other profiles need.   So rather see this merged now, rather than wait until the relationship approach is proposed.

maxhbr (changes requested):

model/Licensing/Classes/NoneText.md
```diff
@@ -0,0 +1,21 @@
+SPDX-License-Identifier: Community-Spec-1.0
```
tsteenbe, line 1: Proposed to rename NoneText to NoneCopyrightText for consistency within the profile and prevent future naming conflicts.

model/Licensing/Properties/additionId.md
```diff
@@ -0,0 +1,29 @@
...
+
+# additionId
+
+## Summary
+
+Provides a short, unique identifier to refer to a LicenseAddition.
+
+## Description
+
+An additionId contains a human-readable, short-form identifier for a
```
maxhbr, line 11: (non-blocking) this should probably be encoded as a [Format->pattern](https://github.com/spdx/spdx-3-model/blob/ac23ca73c879f37b2a1b17867db56cf306aa4097/model/Core/Classes/DateTime.md?plain=1#L24)

swinslow:
@maxhbr Thanks for 0d8ff9e and 24b38fb, I am fine with these.

For the rest of the comments, now that this is passing, can we merge this PR and consider future edits as part of the RC evaluation? Or at least post-merging into main?

swinslow:
> I see no benefit of having namespace-specific properties `comment` and `name`, since they are already in Core.

Issues pending for this at #167 and #210. I'm fine with either of the following:

1. just deleting the name and comment properties from the `Licensing/Properties/` directory, if that's appropriate; or
2. changing these to licenseName/licenseComment and additionName/additionComment if that's preferable.

model/Licensing/Classes/ListedLicense.md
```diff
@@ -0,0 +1,30 @@
...
+- SubclassOf: License
+- Instantiability: Concrete
+
+## Properties
+
+- listVersionAdded
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
+- deprecatedVersion
```
maxhbr, line 26: Maybe use a more consistend name, like:

```suggestion
- listVersionDeprecated
```

model/Licensing/Classes/ListedLicense.md
```diff
@@ -0,0 +1,30 @@
...
+
+- name: ListedLicense
+- SubclassOf: License
+- Instantiability: Concrete
+
+## Properties
+
+- listVersionAdded
+  - type: xsd:string
+  - minCount: 0
```
maxhbr, line 24: why not mandatory?

swinslow: Because not every license on the license list has a value for this field, it was added later than the creation of the license list.

model/Licensing/Classes/ListedLicense.md
swinslow: The existing names tie to the current names in the license-list-XML schema:

https://github.com/spdx/license-list-XML/blob/e9849fde827b339e0e29988e17f167b4be4f7653/schema/ListedLicense.xsd#L52-L53

model/Licensing/Properties/obsoletedBy.md
```diff
@@ -0,0 +1,25 @@
...
+
+The License's or LicenseAddition's comment value may include more information
+about the reason why the licenseId specified in the obsoletedBy value is
+preferred.
+
+## Metadata
+
+- name: obsoletedBy
+- Nature: DataProperty
+- Range: xsd:string
```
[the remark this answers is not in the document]
swinslow: It's possible it could be a license expression rather than a single license. In the [license-list-XML schema](https://github.com/spdx/license-list-XML/blob/main/schema/ListedLicense.xsd) I believe it's essentially a string, so I'd be inclined to leave it as that for the moment.

swinslow:
> > I see no benefit of having namespace-specific properties `comment` and `name`, since they are already in Core.
> 
> Issues pending for this at #167 and #210. I'm fine with either of the following:
> 
> 1. just deleting the name and comment properties from the `Licensing/Properties/` directory, if that's appropriate; or
> 2. changing these to licenseName/licenseComment and additionName/additionComment if that's preferable.

I've gone with option 2 for the time being so that we can get this merged. Can discuss whether to switch to name/comment in Core down the road, since that would involve Licenses and LicenseAdditions being Elements.

kestewart:
@zvr - the properties in core are related to element, and it's not clear that we should be applying this to license.    In meeting discussion just now - we'll go with licenseName, licenseComment for now, and look at merging when we have more time.

model/Licensing/Classes/NoneText.md
kestewart: Considering this non-blocking on merge for now.   Please open issue, so we discuss this after merge.

swinslow:
Thanks all! I'll merge this now.

PR #280, opened 2023-05-05
change # to / and add root .md for dataset profile

maxhbr:
This changes the ids in the root md files to the convention used in the generated spec and adds a `Dataset.md`.


kestewart (approved):
Thanks for fixing.

PR #281, opened 2023-05-05
Licensing: change to license-specific versions for name and comment

swinslow:
Addresses overlap with `name` and `comment` in Core by switching to license- and addition-specific properties, since Licenses and Additions are not currently Elements.

Also fixes a missing `.md` extension for subjectAddition.


model/Licensing/Properties/licenseName.md
```diff
@@ -1,14 +1,14 @@
 SPDX-License-Identifier: Community-Spec-1.0
 
-# name
+# LicenseName
```
kestewart, line 3: licenseName rather than LicenseName

kestewart (changes requested):

kestewart (approved):

PR #284, opened 2023-05-05
Add security assessment relationship types

tsteenbe:


kestewart (approved):
I think we should create a vocabulary for the Assessments, and consolidate to a single relationship, but not going to block for now.

PR #285, opened 2023-05-05
Add Time to build time properties for consistency

tsteenbe:
Most time related fields in the spec end with Time.


kestewart (approved):
LGTM

PR #291, opened 2023-05-05
Software properies (licensing-related)

zvr:
- Moves properties of SoftwareArtifact (licensing-relates) to Software


swinslow (approved):
Works for me if this is where these ought to be located, thank you @zvr!

PR #297, opened 2023-05-06
Rework licensing profile

zvr:
Here's a re-work of the classes in the Licensing profile.

Among other stuff:
- move common stuff from licenses and additions to a common parent
- separate the classes with information on licenses (e.g. licenses and additions) from the "values of licensing info about software"
- everything becoming an Element so it can be referenced and used in a dataset
- simplify naming; since everything is in the "/Licensing" path, we can shorten names

This PR includes a graphical representation of the current classes and a textual representation of the new structure.

The `None` and `NoAssertion` for license and copyrightText should most probably be individuals.

BTW, this re-work is independent to whether properties or relationships will be used to denote the connection between SoftwareArtifacts and licensing information.


zvr:
OK, I guess no interest in me pushing these changes now... We can revisit it in the next months.

goneall:
> OK, I guess no interest in me pushing these changes now... We can revisit it in the next months.

Wish we had more time before RC1 - perhaps revisit in the next week or so?

A couple initial thoughts:
> everything becoming an Element so it can be referenced and used in a dataset

I can see the advantage, but it would also bring in a number of additional properties that may or may not make sense for licenses (e.g. relationships) - I'll need to think about this some more.  Perhaps we create another class in Core which subclasses to both Element and License?

> simplify naming; since everything is in the "/Licensing" path, we can shorten names

I don't mind simplifying the new names, but I'd like to keep the same naming as 2.X unless we have feedback that the existing names cause some issues.

model/Licensing/Classes/AttributionText.md
goneall, line 1: Do we need a new class to represent a string type?  The property description may be sufficient.

model/Licensing/Classes/CopyrightText.md
goneall, line 1: Same comment as above - may not need its own class

model/Licensing/Classes/ConjunctiveSet.md
goneall, line 1: Prefer to keep the current name for compatibility

model/Licensing/Classes/DisjunctiveLicenseSet.md
goneall, line 1: Prefer to keep the current name for compatibility

model/Licensing/Classes/License.md
```diff
@@ -14,30 +14,11 @@ A License represents a license text, whether listed on the SPDX License List
 ## Metadata
 
 - name: License
-- SubclassOf: AnyLicenseInfo
+- SubclassOf: LicensingText
```
goneall, line 17: In SPDX 2.X this is called "SimpleLicensingInfo" - a bit long, but I would prefer the existing name.

model/Licensing/Classes/LicenseExpression.md
goneall, line 1: I'm not quite sure how this fits into the model - I think of LicenseExpression as a text representation of AnyLicenseInfo which is parsed.  I prefer having separate names to represent the class of Text with a specific format vs. that actual license model of the fully parsed tree.

model/Licensing/Classes/LicensingText.md
goneall, line 1: In SPDX 2.X this is called "SimpleLicensingInfo" - a bit long, but I would prefer the existing name.

model/Licensing/Classes/LicensingText.md
```diff
@@ -0,0 +1,39 @@
...
+  - type: xsd:string
+  - minCount: 1
+  - maxCount: 1
+- seeAlso
+  - type: xsd:anyURI
+- standardeTemplate
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
+- isDeprecated
```
goneall, line 31: Deprecated properties only apply to listed licenses IMO.

model/Licensing/Classes/ListedException.md
goneall, line 1: Prefer to keep original name

model/Licensing/Classes/ListedLicense.md
```diff
@@ -20,11 +20,11 @@ at https://spdx.org/licenses.
 ## Properties
 
 - listVersionAdded
-  - type: xsd:string
+  - type: LicenseListVersion
```
goneall, line 23: I like this additional type with the constraints - nice addition

model/Licensing/Classes/OrLaterOperator.md
```diff
@@ -4,14 +4,16 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Portion of an AnyLicenseInfo representing this version, or any later version,
+A license expression representing this version, or any later version,
```
goneall, line 7: Same comment on the use of "license expression"

model/Licensing/Classes/SingleLicense.md
goneall, line 1: I'll need to review the model more carefully - I was thinking this may be SimpleLicenseInfo - probably need to discuss the model real-time

model/Licensing/Individuals/NoAssertionCopyrightText.md
goneall, line 1: I see you created an Individuals directory :)  I like the idea of creating text individuals with definitions as you have done here.

goneall (changes requested):
I'll probably need to do a real-time review of the proposal, but I made a few comments

zvr:
@goneall as a general comment, I do not consider "keep SPDXv2 terms" a strict restriction, given that the whole model is very different -- but more importantly given that all this license expansion stuff was not part of the spec, only of the ontology used (that only few people like us actually used).

Looking forward to a live discussion at some point.

goneall:
> as a general comment, I do not consider "keep SPDXv2 terms" a strict restriction, given that the whole model is very different -- but more importantly given that all this license expansion stuff was not part of the spec, only of the ontology used (that only few people like us actually used).

I agree - it should not be a strict restriction.  At the same time, I only want to change the names if there is a sufficiently good reason since it will cause some effort in migration.

model/Licensing/Classes/OrLaterOperator.md
```diff
@@ -23,7 +25,7 @@ data will need to determine for themselves what meaning to attribute to a
 ## Metadata
 
 - name: OrLaterOperator
-- SubclassOf: AnyLicenseInfo
+- SubclassOf: LicenseExpression
```
meretp, line 28: In the Tooling Mini Summit one use case for the licensing profile was the following license expression "GPL-3.0-or-later WITH AdditionRef-autoconf-gpl3". This implies that we need to be able to use the `OrLaterOperator` as `license` in the `WithOperator`. One option would be to make the `OrLaterOperator` a subclass of `License` but as this would then allow some kind of "nested" `OrLaterOperator` which is not valid on the other hand, we probably need to rethink the structure here. As this is also a bug with the current model, I'll open an issue for this.

zvr: Please note that this specific example is not using the OrLater operator; it's a single license named `GPL-3.0-or-later`.

I agree that we should clarify whether OrLater can be used with With.

meretp: Ah sorry, bad example, but "Apache-1.0+ WITH AdditionRef-autoconf-gpl3" would be an example for this use case, right? I will update the example in the issue.

maxhbr: legacy but still a valid SPDX expression that can be used as example: `GPL-3.0+ WITH AdditionRef-autoconf-gpl3`

kestewart:
Moved this to be considered for -rc2.  @swinslow - can you weigh in on this one, before we look into merging.

sbarnum:
I would pretty strongly disagree with the abstraction of class and property names removing "License".
While it is true that separate namespaces keep the ontology level formality in check, this sort of abstraction will make it highly difficult if not untenable to provide a full rigor rdf serailization such as json-ld in a way that looks like simple json for players that don't care. For this sort of thing to work the json-ld context needs to be able to abstract away the formality details but it cannot do that if properties or classes of the same name exist in multiple namespaces. We really want to avoid that unless absolutely necessary.
I would argue that the name abstractions proposed here are not absolutely necessary. For example, "licenseText" does not need to be reduced to "text".

model/Licensing/Classes/AttributionText.md
```diff
@@ -0,0 +1,20 @@
...
+Class representing attribution text.
+
+## Description
+
+An AttributionText is used to indicate the attribution text of a software artifact.
+
+## Metadata
+
+- name: AttributionText
+- SubclassOf: xsd:string
```
sbarnum, line 16: Classes cannot be subclassed from datatypes

zvr: Yes, @sbarnum ; in RDF it won't be a subclass, but a derived datatype.

goneall:
Here's a partial list of classes and properties renamed in this PR.  I would prefer to keep the original names for compatibility and for the same reasons we came up with the names during the 2.0 modeling exercise.

```
AnyLicenseInfo changed to LicenseExpression
ConjunctiveLicenseSet changed to ConjunctiveSet
CustomLicenseAddition changed to CustomAddition
DisjunctiveLicenseSet changed to DisjunctiveSet
SimpleLicensingInfo changed to LicenseText
ListedLicenseException changed to LicenseException
WithAdditionOperator -> WithOperator
```

There is one exception the above - I agree with changing `isDeprecatedLicenseId` to `isDeprecatedId` so that we can use it in the `LicenseAddition` as well as `ListedLicense`.

kestewart:
Closing this as it's marked as wontfix at this point, and lack critical participants to reassess.

PR #298, opened 2023-05-06
Change copyrightText to be xsd:string

goneall:
Removes the class hierarchy for expressing NONE and NOASSERTION for the copyrightText.  This simplifies the model.  The NONE and NOASSERTION values are already described in the copyrightText property description.  The text values can be parsed and interpreted as text.


zvr:
My approach (outline in #297) was to have a `CopyrightText` class (inheriting from `xsd:string`) and two individuals, `NoneCopyrightText` and `NoAssertionCopyrightText`.

zvr:
Yes, the individuals would have values "NONE" and "NOASSERTION" to make serialization a no-op.

kestewart (approved):

kestewart:
Ok, given Steve's approved this, I'll merge it in now.   Thanks.

swinslow:
+1 on these changes. I'm going to take a quick look through after merging this and see if anything else is needed.

@zvr Definitely interested in looking at #297 after RC / before actual 3.0.

PR #300, opened 2023-05-06
Licensing: Clarify as strings for NONE and NOASSERTION

swinslow:
Follow-on from #298 


goneall (approved):
LGTM

PR #301, opened 2023-05-06
Remove relationshipType from subclasses

goneall:
The property relationshipType is defined in the ancestor class core/Relationship and does not need to be repeated in the sub-classes.


kestewart (approved):
Not essential, but good practice.

tsteenbe (approved):

PR #302, opened 2023-05-06
Change namespace for homepage

### LINKED ISSUE #132: Change homePage property back to doap:homepage
SPDX 2.3 uses a pre-existing RDF vocabulary [DOAP](https://en.wikipedia.org/wiki/DOAP) for the package home page property.

Using DOAP has the advantage of correlating similar information from different sources without having to do explicit mappings.

Using DOAP will also maintain compatibility with SPDX 2.3.

goneall:
Fixes #132


goneall:
@zvr - I need to change the namespace for a property.  This is probably the wrong approach, but let me know if you have an idea how specify a namespace outside of SPDX.

zvr:
Simply use `doap:homepage`.

We'll add `doap` to the standard set of namespaces, like `xsd`, no worries.

goneall:
Going to close this PR and try again...

PR #304, opened 2023-05-06
Change homePage to doap:homepage

goneall:
The property has a namespace of doap - http://usefulinc.com/ns/doap#

In doap, there is no capital P in Page.

### DIFF
--- a/model/Software/Classes/Package.md
+++ b/model/Software/Classes/Package.md
@@ -40,7 +40,7 @@ Note that some of these could be represented in SPDX as a file as well.
   - type: xsd:anyURI
   - minCount: 0
   - maxCount: 1
-- homePage
+- doap:homepage
   - type: xsd:anyURI
   - minCount: 0
   - maxCount: 1
--- a/model/Software/Properties/homePage.md
+++ b/model/Software/Properties/homePage.md
@@ -1,6 +1,6 @@
 SPDX-License-Identifier: Community-Spec-1.0
 
-# homePage
+# homepage
 
 ## Summary
 
@@ -8,7 +8,7 @@ A place for the SPDX document creator to record a website that serves as the pac
 
 ## Description
 
-HomePage is a place for the SPDX document creator to record a website that serves as the package's home page.
+Homepage is a place for the SPDX document creator to record a website that serves as the package's home page.
 This saves the recipient of the SPDX document who is looking for more info from
 having to search for and verify a match between the package and the associated project home page.
 This link can also be used to reference further information about the package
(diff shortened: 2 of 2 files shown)


kestewart (changes requested):
given the property is "homepage" should the filename be "homepage.md" as well?   Rather than "homePage.md"

goneall:
Good catch @kestewart - file is now renamed

kestewart (approved):
Thanks Gary.  Looks good to me now.

model/Software/Classes/Package.md
```diff
@@ -40,7 +40,7 @@ Note that some of these could be represented in SPDX as a file as well.
   - type: xsd:anyURI
   - minCount: 0
   - maxCount: 1
-- homePage
+- doap:homepage
```
maxhbr, line 43: do we have a place, where we fix what `doap` expands to?

goneall: yes

goneall: Yes - there is a file defined where all other namespaces (e.g. XSD) - we will add DOAP to this

zvr:
If we're going to be using `doap:homepage` it's not an SPDX property any more, and we should delete the corresponding property file.

goneall:
> If we're going to be using doap:homepage it's not an SPDX property any more, and we should delete the corresponding property file.

@zvr If we remove it, do we loose the description from our documentation?

maxhbr:
If we use `doap:homepage` we depend on the doap description and definition? We can just describe properties within our own scope. It is also unclear, where one would put it.

zvr:
Yes, using an external property means we're losing our definition/description and depend on the original one.

We used to use properties like pointer(?) from external namespaces and we ended up defining our own `PositiveIntegerRange`.

I propose we keep it in our model instead of using the external one.

model/Software/Properties/homepage.md
```diff
@@ -1,22 +1,22 @@
...
+Homepage is a place for the SPDX document creator to record a website that serves as the package's home page.
 This saves the recipient of the SPDX document who is looking for more info from
 having to search for and verify a match between the package and the associated project home page.
 This link can also be used to reference further information about the package
 referenced by the SPDX document creator.
 
 ## Metadata
 
-- name: homePage
+- name: doap:homepage
```
maxhbr, line 19: we also have no file for `xsd:string` and  the filename is not `daop:hompepage.md`. This is some inconsistency that I can not resolve. I do not see that this is matching the expectations of the spec-parser.


I think that using `doap:homepage` implies removing our file `model/Software/Properties/homepage.md`

maxhbr (changes requested):
for using `doap:homepage`, our file `Software/Properties/homepage.md` should be dropped

PR #305, opened 2023-05-06
Add comments for relationships

goneall:
Replaced all TODO's with comments.

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -20,46 +20,46 @@ Build Profile specific RelationshipType descriptions can be found [here](https:/
 
 ## Entries
 
-- amends: TODOdescription
-- ancestor: TODOdescription
+- amends: The `to` element(s) amends the `from` element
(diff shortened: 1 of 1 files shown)


kestewart (approved):
Thanks Gary.  That's definitely better than it was.   It could use some more wordsmithing, and there are some redundant relationships in the set,  but that can be subsequent PRs.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -20,46 +20,46 @@ Build Profile specific RelationshipType descriptions can be found [here](https:/
 
 ## Entries
 
-- amends: TODOdescription
-- ancestor: TODOdescription
+- amends: The `to` element(s) amends the `from` element
```
maxhbr, line 23: maybe easier to read could be:
```suggestion
- amends: Every `to` element amends the `from` element
```

goneall: Agree - I just updated all the descriptions.

tsteenbe (approved):

PR #307, opened 2023-05-07
Fix cardinality of (primary)Purpose

### LINKED ISSUE #133: Rename packagePurpose to primaryPackagePurpose
The current model show `packagePurpose` as the property name in the SPDX 3 model diagram.  Late in the 2.3 review cycle, we renamed this property to `primaryPackagePurpose`.  We likely just missed updating the SPDX 3 model diagram.

goneall:
Fixes #133

This PR just renames purpose to primaryPurpose and changes the cardinality to 0..1 to be consistent with SPDX 2.3 and support the CycloneDX translation.

### DIFF
--- a/model/Software/Classes/SoftwareArtifact.md
+++ b/model/Software/Classes/SoftwareArtifact.md
@@ -23,9 +23,10 @@ such as a package, a file, or a snippet.
   - type: xsd:anyURI
   - minCount: 0
   - maxCount: 1
-- purpose
+- primaryPurpose
   - type: SoftwarePurpose
   - minCount: 0
+  - maxCount: 1
 - concludedLicense
   - type: /Licensing/LicenseField
   - minCount: 0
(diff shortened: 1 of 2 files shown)


kestewart (approved):
I think we still need a Purpose field with the original cardinality.   Specifically there are use cases when you have a something that is "primary", but others may want to find "secondary" elements as well.   Example - primary of "evidence",  secondary of "data".    primary of "operatingSystem", secondary of "executable",   etc.

kestewart (changes requested):
Would like to see this with a "purpose" and if you want to add in a "primarypurpose" it should be a separate field.

goneall:
Multiple purposes is planned to be used in the Safety profile, therefore we should add back purpose with multiple cardinalities.

goneall:
I added the field `additionalPurpose` to capture additional purposes.  The original name `purpose` may be confused with the `primaryPurpose` field.

goneall:
@kestewart Please review

kestewart (approved):
Reviewed in call.   Going with additionalpurpose, as well as primary purpose.

PR #308, opened 2023-05-07
Add summary and desc. to File

goneall:
I tried to use language consistent with the property names.


tsteenbe (approved):

kestewart:
Updated words look like a good starting point.   Would have approved this if Thomas hadn't merged it already.  ;-)

PR #318, opened 2023-05-10
delete unnecessary external property restriction

meretp:
While working on the implementation in the Python tools I noticed this unnecessary external property restriction.


kestewart (approved):

goneall (approved):
It looks like this could be removed

PR #326, opened 2023-05-11
For NamespaceMap should be both fields mandatory

maxhbr:


goneall (approved):
Agree - good catch @maxhbr

armintaenzertng:
Actually, can't we just make use of the `DictionaryEntry` class as the type of `namespaceMap`?
Then we would also get rid of the double naming `namespaceMap` vs `NamespaceMap`.

kestewart (approved):
Reviewed during call.   No issues raised after walking through diagram.

PR #329, opened 2023-05-12
Security: Fix markdown and add external property restriction for "to"

meretp:
Looking at the generated documentation fro the model, I noticed that the property `security:withdrawn` isn't assigned correctly. This was caused by the wrong title in the markdown file for `withdrawnTime`.

The `to` property for `Relationship` is optional since #293. From the descriptions in the security relationships I think the `to` property should be mandatory for all security relationships.


puerco (approved):
Thanks for catching this @meretp.

I am surprised and not very happy about the minCount:0 cardinality of `to`. Having it optional challenges the idea of a relationship itself :/    At least this makes sure we don't get an incomplete assessment.

PR #333, opened 2023-05-15
fix references to LicenseField

meretp:
Working with the `spec-parser` and the generated model.ttl file I noticed these incorrect references.


swinslow (approved):
LGTM, thanks @meretp!

PR #334, opened 2023-05-15
Security profile: add markdown files for properties

meretp:
This PR adds markdown files for some properties in the security profile which were missing. I didn't know how to fill description and summary, so there are some open TODOs, suggestions welcome.  

Also I updated the properties with `xsd:datetime` to use the introduced `/Core/DateTime` class.


puerco (approved):
Thanks for the correction @meretp , I can take care of the TODOs in a follow-up

tsteenbe (approved):

PR #340, opened 2023-05-16
core: Replace Entity with Agent

nishakm:
- Replace concrete class Entity with concrete class Agent
- Delete class Identity
- Add new class SoftwareAgent
- Make Person and Organization inherit from Agent

Fixes: #94


nishakm:
@sbarnum please review

kestewart (approved):
Thanks Nisha!   This matches what I've understood from the discussions.

maxhbr (changes requested):
`Identity` is also used in:
```
Core/Classes/Artifact.md
24:  - type: Identity
27:  - type: Identity

Core/Properties/originatedBy.md
17:- Range: Identity

Security/Classes/VulnAssessmentRelationship.md
31:  - type: /Core/Identity
```

Entity is also used in:
```
Core/Classes/CreationInfo.md
31:  - type: Entity

Core/Properties/createdBy.md
19:- Range: Entity

Core/Classes/Tool.md
16:- SubclassOf: Entity
```

These findings need to be migrated.


(Also: the spec-parser should probably fail and find these inconsistencies)

model/Core/Classes/Agent.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Agent
+
+## Summary
+
+Agent represents anything acting on a system.
```
armintaenzertng, line 7:
```suggestion
Represents anything acting on a system.
```
Actually, a tool can also act on a system and is not a subclass of Agent...?

model/Core/Classes/SoftwareAgent.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SoftwareAgent
+
+## Summary
+
+A software agent.
```
armintaenzertng, line 7:
```suggestion
A software representation of a user.
```
I think this gives a bit more information in the summary than just restating the class name.

model/Core/Classes/Agent.md
sbarnum, line 7: Suggest changing to "An entity that has the potential to act."

This would align it with the the basic definitions of agent in the W3C PROVO ontology, FIBO ontology, FOAF ontology, and dbpedia ontology.
It also leaves it minimal and flexible for 3.0 as was agreed to.

model/Core/Classes/SoftwareAgent.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Description
+
+A SoftwareAgent is a software representation of a user that could be assigned authorization to perform tasks on behalf of a person or organization.
```
sbarnum, line 11: Suggest changing to: "A SoftwareAgent is a software program with the potential to act on behalf of a user (person or organization) when it possesses appropriate authority."

The word representation has potential confusion of being interpreted as a facsimile. Really the software agent is acting "on behalf of" the user.

iamwillbar (approved):
This matches what is currently in the diagram.

model/Core/Classes/SoftwareAgent.md
kestewart: From discussion in meeting: 

A SoftwareAgent is a software program that is given the authority (similar to a user's authority) to perform an operation.

model/Core/Classes/SoftwareAgent.md
kestewart: A software program given authority to perform an operation.

davaya: Agent is "anything that has the potential to act.. This could be a person, organization, software agent, etc.
Agent is not limited to acting on a system but rather acting in any way or context."

So software agent has the potential to act, and the reason it has that potential is that it has authority, either by a system or anything else that recognizes that authority.

model/Core/Properties/originatedBy.md
```diff
@@ -14,5 +14,5 @@ OriginatedBy identifies from where or whom the Element originally came.
 
 - name: originatedBy
 - Nature: ObjectProperty
-- Range: Identity
+- Range: Entity
```
kestewart: SHould be Agent.

kestewart (approved):
Based on discussion,  there's a subsequent PR needed for security profile to align to agent, but we should be good to merge this, and get coherence.

maxhbr (approved):
LGTM

PR #342, opened 2023-05-16
Update description of licenseId and additionId

### LINKED ISSUE #339: Update description of licenseId and additionId
In ```model/Licensing/Properties/additionId.md``` and ```model/Licensing/Properties/licenseId.md``` we have:

```
It may only include letters, numbers, period (".") and hyphen ("-")
```

First proposed modification: replace ```numbers``` by ```digits```

Rationale: a number can contain a comma, which is not expected.

Second proposed modification: replace ```period``` by ```periods``` and ```hyphen``` by ```hyphens```

Rationale: we can have more than one period and more than one hyphen.

Third proposed modification: define what a letter is.

Rationale: it is not clear what a letter is, is it only letter from A to Z without diacritics, or are PèreNoël and λογισμικού valid identifiers?

vargenau:
Fixes #339

### DIFF
--- a/model/Licensing/Properties/additionId.md
+++ b/model/Licensing/Properties/additionId.md
@@ -9,8 +9,9 @@ Provides a short, unique identifier to refer to a LicenseAddition.
 ## Description
 
 An additionId contains a human-readable, short-form identifier for a
-LicenseAddition. It may only include letters, numbers, period (".") and
-hyphen ("-") characters.
+LicenseAddition. It may only include letters, digits, periods (“.”) and
+hyphens (“-”). A letter is an ASCII character from “A” to “Z”, upper
+and lower case.
 
 For a ListedLicenseException, the licenseId will be as specified on the
 [SPDX Exceptions List](https://spdx.org/licenses/exceptions-index.html) for the
--- a/model/Licensing/Properties/licenseId.md
+++ b/model/Licensing/Properties/licenseId.md
@@ -9,8 +9,8 @@ Provides a short, unique identifier to refer to a License.
 ## Description
 
 A licenseId contains a human-readable, short-form license identifier for a
-License. It may only include letters, numbers, period (".") and hyphen ("-")
-characters.
+License. It may only include letters, digits, periods (“.”) and hyphens (“-”).
+A letter is an ASCII character from “A” to “Z”, upper and lower case.
 
 For a ListedLicense, the licenseId will be as specified on the
 [SPDX License List](https://spdx.org/licenses) for the particular license.


model/Licensing/Properties/additionId.md
```diff
@@ -9,8 +9,9 @@ Provides a short, unique identifier to refer to a LicenseAddition.
 ## Description
 
 An additionId contains a human-readable, short-form identifier for a
-LicenseAddition. It may only include letters, numbers, period (".") and
-hyphen ("-") characters.
+LicenseAddition. It may only include letters, digits, periods (“.”) and
```
maxhbr, line 12: Can you maybe avoid non-ascii characters, as they might confuse the parsing. 
```suggestion
LicenseAddition. It may only include letters, digits, periods (".") and
```

maxhbr (dismissed):

model/Licensing/Properties/additionId.md
maxhbr: It is touching undefined behavior that was never discussed. We have no idea if every documentation that we want to generate supports utf-8 in descriptions. Therefore I was suggesting to stay conservative, to have less issues down the road.

Our parsing and generation tooling is right now not strict enough so that it could catch such potential problems. And it is not keeping up.

vargenau: The following files currently are in UTF-8:

- ./Glossary.md: Unicode text, UTF-8 text, with very long lines (993)
- ./License.md: Unicode text, UTF-8 text, with very long lines (1977)
- ./README.md: Unicode text, UTF-8 text
- ./model/Software/Vocabularies/SBOMType.md: Unicode text, UTF-8 text
- ./model/Software/Properties/attributionText.md: Unicode text, UTF-8 text
- ./model/Software/Properties/declaredLicense.md: Unicode text, UTF-8 text
- ./model/Core/Properties/dataLicense.md: Unicode text, UTF-8 text

zvr: The ones in the spec (inside `/model`) have to be fixed, then. We should have consistency, at the very least.

kestewart:
Waiting to merge until after updates to licensing profile merged.

zvr:
This should be closed (without merging).  We no longer have licenseId and additionId properties, they inherit id from `Element`.

goneall (approved):
LGTM

goneall:
@vargenau Can you update per Max's suggestion on non-ascii characters?  Once changed, I'll merge.

vargenau:
Hi @goneall,
I am a bit confused.
My proposed commit has two things:
- clarify what a letter is (only ASCII)
- replacement of some ASCII quotes "" by smart quotes “”
From the comments, I understand that none of these two modifications should be done.
So what do you want me to update?

goneall:
@vargenau - Never mind, I see the previous comment that these are now inherited fields.  I'll close this issue.

Issue #306, opened 2023-05-06
How to require CreationInfo without inflating the size of serialized data


goneall:
The `creationInfo` is a property of `Element` with a current cardinality of `0..1`.  We should require `creationInfo` at least once per serialization for validation and integrity purposes.  However, changing the cardinality to `1..1` would significantly increase the size of the serialized SPDX data.

zvr:
Not really; in the serialization meetings we've talked about re-using a single `CreationInfo` structure (if it's the same for a number of Elements).

maxhbr:
I think there are currently several options discussed:
1. making it mandatory for some collections, like SpdxDocument or Sbom
1. making it mandatory for all elements and using the `@id` of the creation info for deduplication. E.g. by using the line `"creationInfo": "my_namespace_prefix:110856b8-bd44-4eed-ad03-16c7dbc2cbe3"` in every Element
1. making it mandatory for all elements and using some magic which implicitly applies the creation info from the parent collection to contained elements.
1. adding something to the top level of the serialization that contains the default / fallback creationInfo


### Problems with 1.
One might just serialize a single bare element, and there the requirement would not be enforced.

### Problems with 2.
This adds complexity and noise which needs to be handled by tools. It is also hard to read.

### Problems with 3.
This is magic that is unclear. @sbarnum mentioned that this should be possible.

Elements might be contained in multiple collections.

### Problems with 4.
Might not be easy to add to RDF / JSON-LD. Adds an layer of custom de-serialization logic around already complex RDF.

maxhbr:
related issue: https://github.com/spdx/spdx-3-model/issues/357

davaya:
This issue is caused by trying to mix the logical model with serialization.  We need to agree on what the logical model means.

* logically creation information info is required for every Element - an element can't exist if it wasn't created
* physically we don't want multiple copies of the same info

Both can be achieved if we can agree not to try to represent serialized data in the logical model.

davaya:
#4 is not a problem for RDF/JSON-LD, because they wouldn't use it.  RDF serialization specifies how the elements are retrieved from an RDF payload, JSON serialization specifies how they are retrieved from a JSON payload, and in both cases the deserialized values must be identical (if the logical model defines logical values).

If the logical model is viewed as defining serialized data, that's a whole 'nother problem - it isn't a logical model, and deserializing RDF data will yield a different result than deserializing JSON data or lawyer-readable data.

davaya:
[...]

PR #343, opened 2023-05-16
Propose Individual profile

goneall:
This is one possible solution to issue #306 

The intent is require `creationInfo` on the `Bundle` class and create a profile which requires `creationInfo` on all `Element`s.

This would allow the typical use case of providing a `Bundle` of elements to a user to only be required to specify the `creationInfo` once.  It would also provide producers the option of specifying a `individual` profile which would require `creationInfo` for all `Elements`.  This would be used when a producer provides `Elements` which are not contained in a `Bundle`.

Note that `SpdxDocument` is a subclass of `Bundle` so this approach would be compatible with SPDX 2.X.


goneall:
@sbarnum @tsteenbe @nishakm @kestewart Let me know what you think.

model/Individual/Classes/Element.md
```diff
@@ -0,0 +1,33 @@
+SPDX-License-Identifier: Community-Spec-1.0
```
zvr, line 1: The convention is that you don't create a new class file (as there is not really a new class).

You put the `External properties restrictions` info on the top-level (namespace) file.

model/Individual/Individual.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Description
+
+Although most use cases involve sharing `Elements` inside a `Collection`, there are cases where a stand-alone SPDX element may be made available.
+In those cases, the `Element` `creationInfo` property would be required to identify the `Entity` that created the element.
+
+## Metadata
+
+- id: https://rdf.spdx.org/v3/Individual
+- name: Individual
```
zvr, line 17: you should add the `External properties restrictions` info here.

zvr (changes requested):

model/Core/Classes/Bundle.md
```diff
@@ -22,3 +22,7 @@ A bundle is a collection of Elements that have a shared context.
   - type: xsd:string
   - maxCount: 1
 
+## External properties restrictions
```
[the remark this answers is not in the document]
goneall: I did change my mind after reading the descriptions. I thought that `Bundle` being a subclass of `ElementCollection` and a superclass of `SpdxDocument` being defined as having a "shared context" was a more appropriate to place a restriction.  The reasoning being that if a `Bundle` does not have a shared context, it probably shouldn't be required to share the `creationInfo`.

goneall:
> Now that I think about it, since one can share many elements, not only a single one, maybe call this unbundled profile?

I like it better than `Individual`.  I'll wait until after next week's tech call to rename in case we come up with something even better.

model/Individual/Individual.md
```diff
@@ -0,0 +1,24 @@
...
+
+## External properties restrictions
+
+- /Core/creationInfo
+  - type: /Core/Element/CreationInfo
+  - minCount: 1
+  - maxCount: 1
```
zvr, line 24: This is not the right way. Please use (to mark class, property, and only the values changed):

```
- /Core/Element/creationInfo
  - minCount: 1
```

zvr (changes requested):

model/Individual/Individual.md
goneall: So - is the type /Core/CreationInfo?

model/Individual/Individual.md
```diff
@@ -0,0 +1,24 @@
...
+# Individual
+
+## Summary
+
+A profile to be used when sharing individual SPDX `Element`.
+
+## Description
+
+Although most use cases involve sharing `Elements` inside a `Collection`, there are cases where a stand-alone SPDX element may be made available.
+In those cases, the `Element` `creationInfo` property would be required to identify the `Entity` that created the element.
```
armintaenzertng, line 12: `Entity` will be called `Agent` soon.

model/Individual/Individual.md
```diff
@@ -0,0 +1,24 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Individual
+
+## Summary
+
+A profile to be used when sharing individual SPDX `Element`.
```
armintaenzertng, line 7:
```suggestion
A profile to be used when sharing individual SPDX Elements.
```
I'm not sure we use backticks anywhere else in the markdown files but so far, it does not seem the standard. I think there was a problem with them not working well with the automated documentation, @maxhbr?

model/Individual/Individual.md
armintaenzertng: As you don't change the type of the external property, I don't think you have to specify it here (the same goes for maxCount).

zvr: Right, for external property restrictions, we use _/namespace/class/property_ to point to something and then we list the changed values.

See, for example, how `File` specifies that `name` (a property of `Element`) is mandatory -- this is exactly the same.

model/Individual/Individual.md
goneall: Agree - I left it as the current definition to match the repo

model/Individual/Individual.md
goneall: I'll remove the back ticks

model/Individual/Individual.md
goneall: It has now been updated to Agent

goneall:
Based on discussions on compaction, we decided that having a reference to a common creation info was the preferred path.  Closing this issue.

PR #346, opened 2023-05-18
Security: Add assessmentType property and vocabulary

puerco:
Following the defects meeting on May 17th, @kestewart noted that we need an `assessmentType` property on the subclassed relationships. We want to discuss how to avoid this as it adds needless data but this PR is meant to address the security profile breaking the model while we find a solution.

/cc @kestewart @tsteenbe @rnjudge @jeff-schutt 

The pr introduces a new property `assessmentType` to the abstract class `VulnAssessmentRelationship` that can take a value from an `AssessmentType` vocabulary, comprised of `cvss2`, `cvss3`, ` exploitCatalog`, `ssvc` and `vex`.


model/Security/Vocabularies/AssessmentType.md
```diff
@@ -0,0 +1,25 @@
...
+
+- name: AssessmentType
+
+## Entries
+
+- cvss2: Common Vulnerability Scoring System v2
+- cvss3: Common Vulnerability Scoring System v3
+- exploitCatalog: Known exploits catalog
+- ssvc: Stakeholder-Specific Vulnerability Categorization
+- vex: Vulnerability Exploitability eXchange
```
rnjudge, line 25: I think `epss` is missing from this list to pair with https://github.com/spdx/spdx-3-model/blob/main/model/Security/Classes/EpssVulnAssessmentRelationship.md?

rnjudge (approved):
One small change requested, but otherwise looks good.

model/Security/Vocabularies/AssessmentType.md
rnjudge: epss: Exploit Prediction Scoring System

armintaenzertng:
Could you elaborate on why the model breaks without this addition?

model/Security/Vocabularies/AssessmentType.md
puerco: Good catch Rose, thanks.
Fixed :+1:

puerco:
@armintaenzertng: We were told we need to have an `assessmentType` property to indicate the subclassed property as we cannot rely on the relationship class. I don't fully understand why but, apparently, the `@type` from json-ld has no reliable equivalents in other formats (but don't trust me on that explanation).

kestewart (approved):
These match what has been discussed. 

Thanks for pulling this together!

armintaenzertng:
@puerco Thanks for elaborating, but I don't see that we can afford to lose the information about the type of class. Take as an example `Person`, `Organization` and `SoftwareAgent`, which only differ in the name of the class (but this basically applies to all subclasses of `Element`).

rnjudge:
> @puerco Thanks for elaborating, but I don't see that we can afford to lose the information about the type of class. Take as an example `Person`, `Organization` and `SoftwareAgent`, which only differ in the name of the class (but this basically applies to all subclasses of `Element`).

@armintaenzertng We had similar concerns but were told we need this change. Perhaps @kestewart can elaborate further on why this change is needed?

rnjudge:
In the May 23rd tech call we agreed to discuss this at the May 24th security call. cc @tsteenbe @puerco @jeff-schutt

puerco:
I'm on PTO this week but I'll try to attend from my phone if possible :)

armintaenzertng:
The way I understood it from yesterday's security call:
The `assessmentType` will be a property of a new subclass of relationship called `AssessmentRelationship`. This relationship will be introduced to allow the introduction of more assessment-concerned relationships that are not part of the security profile.
So `assessmentType` would not be a property of `VulnAssessmentRelationship`, but the latter would be a subclass of `AssessmentRelationship` which in turn will get more subclasses in the future.

I hope I got that correctly.

tsteenbe:
Types of data classes will be handled in the serialization so no need to added assessment types to the model - as discussed in the SPDX tech meeting May 30, 2023 and Security team call on May 31, 2023.

Issue #224, opened 2023-04-23
Meta issue - comparison of license model to SPDX 2.3 model


goneall:
I went through and compared the [SPDX 2.3 UML class diagram](https://github.com/spdx/spdx-spec/blob/development/v2.3.1/model/SPDX-UML-Class-Diagram.jpg) to the SPDX 3 license model.

Below is a summary of the differences I found with references to separate issues to track decisions on if we should change from the 2.3 model.

- [x] Class `LicenseExpression` is named `AnyLicenseInfo` in 2.3.  Issue  #222 
- [x] Classes `CopyrightText`, `CopyrightTextField` are just string fields in 2.3.  Related to issue #76 
- [x] Class `CustomerLicense` is named `ExtractedLicenseText` in 2.3. Issue #223
- [x] Class `LicenseField` is not present in the 2.3 model.  Related to issue #76 
- [x] Class `ListedLicenseException` is just `LicenseException` in 2.3.  This probably makes sense with the proposed changes for Custom Exceptions
- [x] Classes `NoAssertionLicense`, `NoneLicense`, `NoAssertionText`, and `NoneText` does not exist in SPDX 2.3. Related to issue #76 
- [x] `OrLaterOperator` and `WithExceptionOperator` license property type is `AnyLicenseInfo` in 2.3 - related to other type issues listed above

zvr:
Basic questions on the modeling of licensing info:

- should licensing info be _properties_ of elements or _relationships_ between elements and licenses?
- to which Elements does it apply? `Artifact` or `Software/Package`, `Software/File`, `Software/Snippet`? Or to a new `Software/SoftwareArtifact` class?

swinslow:
Setting to 3.0 since I suspect there are a few details that would be helpful for us to walk through live and confirm on before closing this.

@zvr for the properties vs. relationships question, I think we're tracking that in #254 so we can discuss that one there. For the question about which Elements it applies to, I agree with having them apply to the new `SoftwareArtifact` class so I think we're resolved on that.

goneall:
It looks like the above punch list has been resolved - if any are still open, please open a fresh issue.

PR #353, opened 2023-05-24
Initial IM description

davaya:


mlieberman85:
What is JIDL? Can you add something to one of the READMEs to clarify what it is? Reading through this as an outsider who is looking to make SPDX tools I'm not sure how to consume this to better understand the actual data model.

rnjudge:
I would expect to see something like: https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json for the 3.0 spec. @goneall I'm pretty sure you're planning on this when the spec is more concrete?

goneall:
> I would expect to see something like: https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json for the 3.0 spec. @goneall I'm pretty sure you're planning on this when the spec is more concrete?

Yes - we will have a JSON schema generated (either directly or indirectly) from the markdown source files. @armintaenzertng is looking into tooling for this and I plan to contribute as well.

serialization/InformationModel/README.md
```diff
@@ -0,0 +1,307 @@
...
+
+The classes of a logical model are either referenceable or datatypes.
+* A **datatype** class is known only by its value.
+Two instances of the class are equivalent if and only if every property in the class is equal.
+* A **referenceable** class has a unique identifier (primary key) and each instance is known by its id.
+Two instances of the class are considered equivalent if and only if their ids are equal.
+If only a single value can correspond to an id, the class is a *mapping* and colliding values
+are treated as an error.
+
+SPDX v3 organizes all of its information into a single referenceable "Element" class - all logical
```
goneall, line 47: Just a note for future reference - there is an[ open question](https://github.com/spdx/spdx-3-model/issues/224#issuecomment-1527294548) as to whether a License should be considered a peer of "Element" or a subclass of element.  The outcome of this may impact the above statement.

serialization/InformationModel/README.md
```diff
@@ -0,0 +1,307 @@
...
+(spdx-tools-3.0.1), and a Person (iamwillbar). The first task is to determine the logical values of these
+three elements. 
+
+The logical values of these elements are:
+
+**Package:**
+```json
+{
+  "id": "urn:spdx.dev:spdx-tools-3.0.1",
+  "type": {
```
goneall, line 138: In the SPDX 2.X JSON format, we took the approach of having a different list/array for each type rather than including a "type" field.  See https://github.com/spdx/spdx-spec/blob/844144b3785dbc6a35065eda3b9d36adda874540/examples/SPDXJSONExample-v2.3.spdx.json#L61 for an example.
I can see some advantages to having the "type" approach proposed, but since it introduces an incompatibility I would like to discuss before agreeing.  @armintaenzertng  - any thoughts on this as another SPDX tool contributor?

serialization/InformationModel/README.md
```diff
@@ -0,0 +1,307 @@
...
+```
+Note that every JSON structure has both leaf-path (RFC 6901) and hierarchical representions. The two represent
+identical values and can be converted back and forth without any semantic knowledge.
+For example, the path representation of the Sbom element is:
+
+**Sbom (path):**
+```json
+{
+  "id": "urn:spdx.dev:null-sbom",
+  "type/sbom/element/0": "urn:spdx.dev:iamwillbar",
```
goneall, line 224: Could you include a fragment in one of the URN examples (e.g. `urn:spdx.dev:people#iamwillbar` just to show it is possible and acceptable?  Important for converting from SPDX 2.X.

serialization/InformationModel/Schemas/spdx-v3.jidl
goneall, line 1: As suggested in one of the comments, we could add a README in the schema's directory describing the various schemas.  We can add JSON schema (and others) when they are available.

goneall (changes requested):
Thanks @davaya for the examples and descriptions.  I found the README's to be very complete and descriptive.  I had one design oriented question related to the use of "type" parameters and one request for using fragments in the URN's.

davaya:
> What is JIDL? Can you add something to one of the READMEs to clarify what it is? Reading through this as an outsider who is looking to make SPDX tools I'm not sure how to consume this to better understand the actual data model.

I'll add a description.

I tried to keep information modeling as a design approach separate from the JSON Abstract Data Notation (JADN) information modeling language, but examples have to be in some actual format.  JADN's native format is JSON data - for an Annotation element
image
the JADN model is:
```json
    ["Annotation", "Record", [], "", [
      [1, "annotationType", "AnnotationType", [], ""],
      [2, "subject", "Element", ["L"], ""],
      [3, "statement", "String", ["[0"], ""],
      [4, "contentType", "MediaType", ["[0"], ""]
    ]],
```
but for documentation purposes an IDL (a text-based "little language" or DSL) is easier to read.  The JADN IDL representation of Annotation is:
```
Annotation = Record
   1 annotationType   AnnotationType
   2 subject          Link(Element)
   3 statement        String optional
   4 contentType      MediaType optional
```
And because some standards (e.g., STIX) use [property tables](https://docs.oasis-open.org/cti/stix/v2.1/os/stix-v2.1-os.html#_4jegwl6ojbes) as their documentation format, JADN also has a property table representation:

**Type: Annotation (Record)**

| ID | Name               | Type           | \#   | Description |
|----|--------------------|----------------|------|-------------|
| 1  | **annotationType** | AnnotationType | 1    |             |
| 2  | **subject**        | Link(Element)  | 1    |             |
| 3  | **statement**      | String         | 0..1 |             |
| 4  | **contentType**    | MediaType      | 0..1 |             |

davaya:
> I would expect to see something like: https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json for the 3.0 spec. @goneall I'm pretty sure you're planning on this when the spec is more concrete?

A JSON schema can be created manually, as well as automatically from a JADN information model.  An example (from a model several months old) is [here](https://gist.github.com/davaya/82489e6fb2eee5a43cbc2f8a17031142).  It will be updated as the logical model, the serialization rules, and the translation tools are matured.

```json
    "Annotation": {
      "title": "Annotation",
      "type": "object",
      "additionalProperties": false,
      "required": [
        "annotationType",
        "subject"
      ],
      "maxProperties": 100,
      "properties": {
        "annotationType": {
          "$ref": "#/definitions/AnnotationType",
          "description": ""
        },
        "subject": {
          "$ref": "#/definitions/Element",
          "description": ""
        },
        "statement": {
          "type": "string",
          "maxLength": 255
        },
        "contentType": {
          "$ref": "#/definitions/MediaType",
          "description": ""
        }
      }
    }
```

serialization/InformationModel/README.md
davaya: Yes.  An information model can define an arbitrary number of referenceable class types, with or without inheriting from one class to another.  It looks like they are leaning toward making license entries Elements, but I have some reservations about throwing so much stuff (like extensions) into the base class.  I might define Element without the kitchen sink, and ExtensableElement from it, and if there is no plausible use case for adding extensions to Licenses and Annotations, they'd inherit from the simpler base.

Thanks for the heads up.

serialization/InformationModel/README.md
davaya: Yes.  A payload and the SpdxDocument element that describes it will have a namespace.  If that Document/Payload contains only Person elements, the namespace could be `urn:spdx.dev:people#` and each Person IRI would be of the form `iamwillbar`.  If the payload has elements of other types (individually typed), the payload namespace could be `urn:spdx.dev:` and the Person IRIs would be `person#iamwillbar`, `package#ctime`, etc.

serialization/InformationModel/README.md
armintaenzertng: I currently lean more towards having a `type`/`@type` property on each element as this will mark better compatibility to json-ld. I agree with David that "compatibility" with SPDX2 should mostly refer to lossless translation from v2 to v3, not retaining the same syntax (which will look drastically different with the new model, anyway).

maxhbr (changes requested):
I do not think that this PR should be merged. It adds a competing JSON serialization that will cause confusion. We should first motivate what its benefit above JSON-LD based JSON is.

PR #344, opened 2023-05-17
Automatically generate the Ontospy documentation

meretp:
With this PR the GitHub Action is updated to automatically generate the html documentation and push the documentation (as multi-page and one-page document) as well as the generated `model.ttl` to the `gh-pages` branch. According to #140 the model.ttl should be pushed to a `generated` branch. To keep things simple I suggest to put everything that is automatically generated in the gh-pages branch, if you disagree I can adapt the PR to only push the generated documentation and we can add another workflow to push the `model.ttl` in another PR.

There is one open PR in the `spec-parser` to be able to fill the `Usage` table for each class with the corresponding properties. (https://github.com/spdx/spec-parser/pull/52). As this is already done in the documentation available via the `gh-pages` branch now, this PR should be merged once this feature is available in the `spec-parser`, so I'll leave this PR as a draft for the time being.

The workflow will only run on main and only push to the gh-pages branch if there were any changes to the generated documentation. Addditionally the generated-documentation will be put in a folder named `auto-generated` to mark these files as automatically generated.

PR #355, opened 2023-05-26
Add json-ld serialization examples

armintaenzertng:
As there seems to be some confusion around how json-ld would actually look like, this is an attempt to provide a place for examples on which further discussions can be based.
I'm not sure where the most interest lies, so please tell me which use cases you would like to see further examples on! :)

Validation of json-ld would be performed against a separate SHACL file generated from the specification. This will be automatically added and kept up-to-date once #344 is merged.
Work is also underway to generate a json-schema that will validate json-ld output.

There is still the hope that the json-ld serialization works fine for everyone so that we don't have to invent a separate json (non-json-ld) serialization. If you have any concerns about this, please speak up!
(For clarification: A "json serialization" will _not_ be json-ld in its expanded form, i.e. with the `@context` used to expand all values to their full URIs).

Current drawbacks:
- The `context.json` needs some touch-up, therefore the `@context` url is currently unreachable. This will be fixed soon.
- Serialization of license expressions is still under discussion, therefore I did not include the licensing profile so far.

### DIFF
--- a/serialization/json_ld/examples/spdx_document.jsonld
+++ b/serialization/json_ld/examples/spdx_document.jsonld
@@ -1,0 +1,105 @@
+{
+  "@context": "https://spdx.github.io/spdx-3-model/rdf/context.json",
+  "@graph": [
+    {
+      "@type": "Person",
+      "@id": "https://some.namespace#SPDXRef-Actor-creatorName-some@mail.com",
+      "creationInfo": {
+        "@id": "_a",
+        "specVersion": "3.0.0",
+        "created": "2022-12-01T00:00:00",
+        "createdBy": [
+          "https://some.namespace#SPDXRef-Actor-creatorName-some@mail.com"
+        ],
+        "profile": [
+          "core",
+          "software"
+        ],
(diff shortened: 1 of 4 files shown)


goneall (approved):
Agree with having these examples - it would help with the discussions.

Thanks @armintaenzertng

armintaenzertng:
I added an explanation of the general way json-ld (and in extension maybe even json) works to alleviate some fears of getting in contact with it.

serialization/json_ld/examples/spdx_document.jsonld
```diff
@@ -0,0 +1,105 @@
...
+      ]
+    },
+    {
+      "@type": "SpdxDocument",
+      "@id": "https://some.namespace#SPDXRef-DOCUMENT",
+      "creationInfo": "_a",
+      "name": "documentName",
+      "comment": "documentComment",
+      "elements": [
+        "https://some.namespace#SPDXRef-DOCUMENT",
```
sbarnum, line 36: This line is self-referential and does not really make sense.

PR #364, opened 2023-06-01
add list of serialization use cases

armintaenzertng:
As discussed in today's serialization meeting, it will be helpful to have a list of use cases that elucidate how serializations will work in action. Examples for these use cases will be included in sub-directories of the different serialization formats.

The current list is only concerned with `Core` and `Software`.
We would be happy if other profiles could contribute use cases they have been or still are struggling with.

### DIFF
--- a/serialization/README.md
+++ b/serialization/README.md
@@ -6,8 +6,35 @@ _TBD_
 
 ## Serialization formats
 
-The files in this directory provide some notes on the specific serialiation formats.
+The files in this directory provide some notes on the specific serialization formats.
 
-The notes are nunbered for easier referencing -- the order is **not** significant.
+The notes are numbered for easier referencing -- the order is **not** significant.
 
+## Use cases
 
+Examples of how to serialize the following cases:
+- Person
+- Agent
+- Annotation
+- File
+- Package
+- Package with ExternalIdentifier
+- Package with ExternalReference
+- Relationship with Package contains two Files
+- Relationship with time properties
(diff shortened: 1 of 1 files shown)


zvr (approved):

davaya:
The [jsonld folder](https://github.com/spdx/spdx-3-model/tree/main/serialization/json_ld/examples) already (prior to this PR) contains three files.  The package and person examples each show a single Element, but the spdxdocument example is a single file containing multiple elements - i.e., it is a serialized payload example, not a logical Element example.

The use cases file should be phrased to avoid confusion between Elements and payloads - all but two of them are worded as "Element with X":

- Package with ExternalReference
- Relationship with Package contains two Files
- Relationship with time properties
- SBOM containing two Files
- SpdxDocument containing two Files
- SpdxDocument with NamespaceMap
- SpdxDocument with ExternalMap
- Person with no CreationInfo

Please change the two use cases of "containing" to "with".  It may sound like wordsmithing, but we should reinforce the idea that the logical graph is composed of individual Elements that "reference" each other, while serialized data may "contain" other serialized data.  For the use cases we need a single SBOM element that references two files, and a single SpdxDocument element that references two files.  When those elements are serialized into a payload, the individual element data may either be nested or listed.

"With" works equally well to describe nested and listed serializations.

armintaenzertng:
I replaced the two "containing" with "with"

maxhbr (approved):
LGTM

PR #370, opened 2023-06-06
Logical Value Examples

davaya:


davaya:
The Serialization team has proposed a set of [serialization use cases](https://github.com/spdx/spdx-3-model/pull/364).  This PR proposes a logical value notation derived from the logical model, and a set of logical value examples for those use cases.

serialization/json_ld/examples/package.jsonld
```diff
@@ -26,7 +26,7 @@
           "algorithm": "sha1",
           "hashValue": "71c4025dd9897b364f3ebbb42c484ff43d00791c"
         },
-                {
```
maxhbr, line 29: I think this change should not be mixed with other changes, as it is very prone to cause merge conflicts

serialization/json.md
```diff
@@ -1,26 +1,85 @@
 # JSON Serialization
 
-Ideally JSON will be very similar (or even the same) to JSON-LD.
+As described in [Logical Values](logical/README.md), JSON is a general purpose data format
```
maxhbr, line 3: Your "logical" serialization should not be a dependency of the json serialization. The json one should be self sufficient and independent.

maxhbr (changes requested):
I think that JSON should not depend on the "logical"-proposal and should.

maxhbr:
We are already observed as having to many and to complicated serializations. Having these exotic and custom `.logval` files in there, before any accepted serialization might make it worse and might promote this view.

serialization/json.md
davaya: "Logical values" exist whether they are documented or not.  An application that de-serializes data will have some in-memory state as a result.  We don't want that state to have any dependency on which serialization format was used - strange, difficult-to-diagnose problems could occur: "the application did this when I read a JSON SBOM, but did something else after reading the same SBOM in RDF"

I anticipated that the application state would be discussed and agreed to before we define any of the serializations, hence the wording.  But the preface clause is unnecessary for the JSON serialization and can be deleted.

davaya:
"Logical Value" comes directly from the Logical Model, it is not a serialization.  It says that after you de-serialize an Annotation Element, every conforming SPDX application must be able to tell you the value of each of the properties defined by Annotation, and the properties inherited by Annotation from Element.

Ask yourself, after looking at the diagram, what the value of "annotation.statement" must be in any serialized data.  Then repeat that exercise for every property of every concrete type.   If we can't understand or agree on what the logical model is saying, then we can't define serialized data that accurately and unambiguously represents the model.

image

PR #374, opened 2023-06-07
Logical Value Examples

davaya:
Replaces #370 to address comments


davaya:
"Logical Value" comes directly from the Logical Model, it is not a serialization. It says that after an instance of an Annotation Element is de-serialized, regardless of data format, an SPDX application must know the count and value of the scalars (strings, numbers, booleans) that make up that instance without regard to structure or naming.

davaya:
"Logical Values" are intended to represent questions an application can ask or answer about components of a logical model, without being confused with serialized data.  The component classes in the [tools-python spdx3 model] answer the identical questions through setters and getters, and would never be confused with serialized data.

Closing because setter/getter methods serve the identical purpose, at least to those familiar with Python programming.

PR #373, opened 2023-06-06
Json examples

davaya:

PR #375, opened 2023-06-07
Json examples

davaya:
Add JSON examples for serialization use cases.
Create initial CBOR README.
Replaces #373 to address comments


maxhbr:
lets wait for the JSON-LD based JSON serialization before merging that. I think there are still two competing alternatives and we should see both before we commit in one direction.

serialization/json/README.md
```diff
@@ -0,0 +1,87 @@
...
+```
+There is no need to transmit RDF information in JSON-serialized data, and there is no reason to
+require all JSON-consuming applications to recognize RDF information, even if only to discard it.
+
+### Definitions:
+* **Logical Value** - the value of an application variable prior to serializing or
+after deserializing an Element instance.
+* **ElementSet** - an arbitrary set of Logical Values.
+* **Payload** - the result of serializing an ElementSet.
+* **SpdxDocument** - an SPDX Element containing metadata about a Payload.
```
maxhbr, line 50: your definitions do not align with https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/SpdxDocument.md. Instead of discussing it here buried down in a long file in some PR, maybe you can create issues and PRs for such statements.

davaya: They do agree, but the distinction between the logical model (which the diagram and model classes represent) and the physical model (which is represented by the payload <<interface>> and which we are just getting around to creating now) is still not clear.

A physical File is a bunch of C code or anything else that goes in files.  A physical Package is something that you install from a package manager.  A physical SpdxDocument is the serialized value of a bunch of Elements.

* An SPDX File logical element as contained in the model doesn't contain C code, it describes the C code.
* An SPDX Package logical element cannot be installed from a package manager, it describes the package that can be installed.
* An SPDX SpdxDocument logical element does not contain serialized SPDX Element values, it describes serialized bytes of those values.  There is only one SpdxDocument logical element that describes a dozen physical SpdxDocument serialized byte strings.

Because the logical model isn't a physical model, we had long discussions about how to represent a series of bytes on the logical model diagram and the logical model files, and William came up with the name "Payload" for the physical SpdxDocument to distinguish it from a logical Element.

davaya:
Add payload example based on serialization team discussion

goneall:
Moving to 3.0 milestone - in the serialization meeting we decided to focus on JSON-LD for RC2 and the other serializations for 3.0.

Issue #368, opened 2023-06-05
Convert SemVer, MediaType and DateTime into Properties


armintaenzertng:
Currently, [SemVer](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/SemVer.md), [MediaType](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/MediaType.md) and [DateTime](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/DateTime.md) are treated as classes that are subclasses of "xsd:string". @sbarnum mentioned that this is invalid. The result in OWL currently looks like this:
```
core:DateTime a owl:Class ;
    rdfs:comment """A Datetime is a string representation of a specific date and time.
It has resolution of seconds and is always expressed in UTC timezone.
The specific format is one of the most commonly used ISO-8601 formats.
Format restriction: pattern: ^\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\dZ$""" ;
    rdfs:subClassOf xsd:string ;
    ns0:term_status "Stable" .
```
As the whole point of these "classes" is to have them restricted via a specific pattern, I propose to change them to simple properties that must follow the pattern. That is, move and convert the SemVer, MediaType and DateTime md-files to the Properties folder. Summary and Description would be kept intact so that we lose no information in the process.

For the OWL this would mean to change for example the following snippet
```
[ sh:datatype core:DateTime ;
            sh:maxCount 1 ;
            sh:name "buildEndTime" ;
            sh:path build:buildEndTime ]
```
into 
```
[ sh:datatype xsd:string ;
            sh:maxCount 1 ;
            sh:name "buildEndTime" ;
            sh:path build:buildEndTime ;
            sh:pattern "^\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\dZ$" ]
```

Please speak up if there are aspects to this that I haven't considered, else I'll open a PR for this.

sbarnum:
Just a note that the above proposed SHACL change is appropriate for application to the NodeShape for the Class that the property applies to. The spec will also need the actual definitions of the properties themselves to also change to have 'rdf:type owl:DatatypeProperty ;' and 'rdfs:range xsd:string ;'

davaya:
[...]

PR #377, opened 2023-06-07
Merge `SemVer` and `specVersion`

armintaenzertng:
this is part of #368

Note: This moves the heading "Format" to Properties files, so the spec-parser needs to be adapted.


goneall (approved):
Agree with the approach - @zvr should we wait for the spec parser to be updated before merging?

armintaenzertng:
I think updating the parser before merging would be better, else we get a continuously failing pipeline again that everybody simply ignores.
I will look into this.

armintaenzertng:
@zvr, @goneall: The spec-parser PR is ready for review here: https://github.com/spdx/spec-parser/pull/57

davaya:
The pattern in specVersion.md conflicts with the SemVer specification:

"2. A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0."

Issue #1: v3.0.0 that appears in our examples is not valid according to SemVer.  Do we want to allow strings in SPDX versions or just non-negative integers?
 
Issue #2:  SPDX is a standard, not a software product.  Do we want to force all specVersions to have a patch version, or should 2.3 and 3.0 be valid SPDX versions?

Issue #3: Since SemVer is used by more than just SPDX, should SemVer be left as a reusable Core type that could be included in profile-defined Elements?

armintaenzertng:
@davaya: Your issues 1 and 2 should be separate issues from this PR as I did not change anything regarding the formatting of `specVersion`.
Regarding your issue 3: When this PR is merged, there is currently nothing in the spec that references the `SemVer` type. I don't think we should keep it around simply for the possibility that it might be used later. We can always reintroduce it at that point.

zvr (approved):

davaya:
> Can you explain, @davaya ? If this is the regex pattern, I took it from the official semver.org site.

The model file is currently:
```
## Metadata
- name: SemVer
- SubclassOf: xsd:string

## Properties

## Format
- pattern: ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
```
The format pattern allows things like RCs and alpha/betas that are useful for software releases, but it conflicts with [SemVer item 2](https://semver.org/#spec-item-2) that uses only integers.   The latter seems more appropriate for documents - are we ever going to have elements with specVersion = `"1.0.0-alpha+001"` or `"1.0.0+20130313144700"` or `"1.0.0-beta+exp.sha.5114f85"`?

> This PR gets rid of the `SemVer` _class_, and changes the property to be a string with a value-matching restriction. The alternative would have been to have a class with an extra property for this, since you can't really subclass data types like `xsd:string`.

XSD doesn't use the word "class", but putting restrictions like length or range or pattern on fundamental classes like string and integer is subclassing them.  For example, "Age" is a SubclassOf: "xs:integer" with range restriction: https://www.w3schools.com/xml/schema_facets.asp.

Being able to re-use restricted fundamental types like SemVer or Age is the reason for defining them in Class model files.

zvr:
To the first point, yes, we definitely will be using `3.1.0-rc1` and such things. That's why we went to semver instead of staying with `X.Y` as SPDXv2.

To the second point, the link you provided shows exactly what I've been saying above; you can add a restriction to the value of the property `age`. You cannot define something ("class", "type", ...) named `Age` and have properties being of this something.

This is what this PR is doing. Since we cannot have `SemVer`, we add the restrictions to the property `specVersion`.

davaya:
As far as age goes, the [XSD specification](https://www.w3.org/XML/Schema) mentions **class** only six times, all referring to programming language classes (C++, Python, .net).   But XSD does define [content type](https://www.w3.org/TR/xmlschema11-1/#Complex_Type_Definition) which, stripped of verbiage related to XML element attributes, reads:

> A complex type definition is a content type applicable to the children of an element information item. The content type may require the children to be empty, or to be a string which belongs to a particular simple type, or to contain a sequence of element information items.

Or to paraphrase, an element can be required to have no children, one simple child element, or multiple child elements.  The text of XSD is quite dense, but the question to be answered is simple: is
```
<SemVer>3.0.1-rc1</SemVer>
```
a valid XML element with type SemVer?  And can XSD be used to define that type?  And can the SemVer type be the type of CreationInfo's semVer property?

Unlike XML, JSON does not send types (e.g., SemVer) over the wire.  But JSON Schema does define types, and the SemVer type can be a JSON string with a pattern, and the CreationInfo type can have a semVer property with type SemVer.

The "Classes" model files define "Types" regardless of what name is used, and the SemVer "class/type" can be the type of a "semVer" property.

davaya:
>  You cannot define something ("class", "type", ...) named Age and have properties being of this something.

How so?  Given the example in https://www.w3.org/TR/xmlschema11-1/#Complex_Type_Definitions:
```
<xs:complexType name="PurchaseOrderType">
  <xs:sequence>
   <xs:element name="shipTo" type="USAddress"/>
   <xs:element name="billTo" type="USAddress"/>
   <xs:element ref="comment" minOccurs="0"/>
   <xs:element name="items"  type="Items"/>
  </xs:sequence>
  <xs:attribute name="orderDate" type="xs:date"/>
 </xs:complexType>
```
what prevents adding a property `name="age" type="Age"` to the PurchaseOrder?

davaya:
Perhaps the confusion comes from the model files - they're all over the place with regard to subclassing - some are blank, some say none, some say xsd:string, which is incorrect, some have profile names, most don't.  It's a mess.

When the model is cleaned up, no datatypes, including CreationInfo, SemVer, DateTime, PositiveIntegerRange, etc, will be shown as subclasses.  They all should be "none" or blank, not a mixture of both.

```
      Class                              Subclass
/Software/Package                    AIPackage
/Core/Element                        Build
Element                              Agent
Element                              Annotation
Payload                              AnonymousPayload
Element                              Artifact
Bundle                               Bom
ElementCollection                    Bundle
                                     CreationInfo
xsd:string                           DateTime
                                     DictionaryEntry
Payload                              Element
Element                              ElementCollection
                                     ExternalIdentifier
none                                 ExternalMap
none                                 ExternalReference
IntegrityMethod                      Hash
                                     IntegrityMethod
Relationship                         LifecycleScopedRelationship
xsd:string                           MediaType
                                     NamespaceMap
Agent                                Organization
none                                 Payload
Agent                                Person
none                                 PositiveIntegerRange
Element                              Relationship
xsd:string                           SemVer
Agent                                SoftwareAgent
Bundle                               SpdxDocument
Element                              Tool
/Software/Package                    Dataset
Licensing/AnyLicenseInfo             ConjunctiveLicenseSet
Licensing/AnyLicenseInfo             DisjunctiveLicenseSet
AnyLicenseInfo                       ExtendableLicense
/Core/Element                        AnyLicenseInfo
License                              CustomLicense
LicenseAddition                      CustomLicenseAddition
ExtendableLicense                    License
/Core/Element                        LicenseAddition
License                              ListedLicense
LicenseAddition                      ListedLicenseException
ExtendableLicense                    OrLaterOperator
AnyLicenseInfo                       WithAdditionOperator
VulnAssessmentRelationship           CvssV2VulnAssessmentRelationship
VulnAssessmentRelationship           CvssV3VulnAssessmentRelationship
VulnAssessmentRelationship           EpssVulnAssessmentRelationship
VulnAssessmentRelationship           ExploitCatalogVulnAssessmentRelationship
VulnAssessmentRelationship           SsvcVulnAssessmentRelationship
VexVulnAssessmentRelationship        VexAffectedVulnAssessmentRelationship
VexVulnAssessmentRelationship        VexFixedVulnAssessmentRelationship
VexVulnAssessmentRelationship        VexNotAffectedVulnAssessmentRelationship
VexVulnAssessmentRelationship        VexUnderInvestigationVulnAssessmentRelationship
VulnAssessmentRelationship           VexVulnAssessmentRelationship
/Core/Relationship                   VulnAssessmentRelationship
/Core/Element                        Vulnerability
/Software/SoftwareArtifact           File
/Software/SoftwareArtifact           Package
/Core/Bom                            Sbom
/Software/SoftwareArtifact           Snippet
/Core/Artifact                       SoftwareArtifact
/Core/LifecycleScopedRelationship    SoftwareDependencyRelationship
```

armintaenzertng:
This is obsoleted by #424.

PR #379, opened 2023-06-08
Added dataset types vocabulary

### LINKED ISSUE #363: Consider making datasetType an enum instead of String
Currently, the [datasetType](https://github.com/spdx/spdx-3-model/blob/main/model/Dataset/Properties/datasetType.md) property is a string. Instead, we could consider introducing a new vocabulary enumerating different Dataset types, and change the property to be a list of these options instead of just a string. We would use a List instead of just one to account for multi-modal datasets (which could have multiple types).

For the list of types, I have this list borrowed this from [Wikipedia](https://en.wikipedia.org/wiki/List_of_datasets_for_machine-learning_research#List_of_sorting_used_for_Datasets):
* Tabular
* Graph
* Text
* Image
* Sound
* Video
* Other

Happy to make a PR if this seems like an acceptable change!

willarmiros:
Closes #363 

This PR introduces a new vocabulary that can be used to describe types of a dataset in a concrete way. Now, instead of a string to describe a `datasetType`, users can choose from the selection for the type(s) that describe their dataset, making it easier for tools to generate and parse standardized dataset types. See #363 for how we got this list of types. Of course users can always specify "Other" if their type isn't included, or "noAssertion" if they don't know the type. I tried to include real-life examples for each entry, but am open to other suggestions for examples!

### DIFF
--- a/model/Dataset/Classes/Dataset.md
+++ b/model/Dataset/Classes/Dataset.md
@@ -19,9 +19,8 @@ Metadata information that can be added to a dataset that may be used in a softwa
 ## Properties
 
 - datasetType
-  - type: xsd:string
+  - type: DatasetType
   - minCount: 1
-  - maxCount: 1
 - dataCollectionProcess
   - type: xsd:string
   - minCount: 0
--- a/model/Dataset/Properties/datasetType.md
+++ b/model/Dataset/Properties/datasetType.md
@@ -8,10 +8,10 @@ Describes the type of the given dataset.
 
 ## Description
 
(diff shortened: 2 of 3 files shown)


willarmiros:
@bennetkl @goneall @rgopikrishnan91 feel free to take a look :)

goneall (approved):
LGTM

I really like having the enumerations.  Just a note on compatibility - we can always add more enumeration values in future release and be backwards compatible.  If we change from an `xsd:String` to an enumeration, it will be a breaking change and need to wait for a major release of the spec. - so I support merging in this PR.

kestewart (approved):
Moving this from a string to an enumerated set is a good idea.   The set of values looks like a reasonable starting point.

PR #385, opened 2023-06-14
Change `purpose` references to `primaryPurpose`

armintaenzertng:
Small fix as `purpose` is now split into `primaryPurpose` and `additionalPurpose`. I changed the reference to the former as having only `additionalPurpose` mandatory wouldn't make much sense to me.


Karlitos88 (approved):

goneall (approved):
Good catch @armintaenzertng

PR #388, opened 2023-06-15
add json-ld context to gh-pages

armintaenzertng:
This adds the current state of the context for JSON-LD to the gh-pages, so that we have a URL and more central place to reference it.


maxhbr:
I like that approach, and this can help us to get forward with JSON-LD!

maxhbr (approved):
LGTM

This is the context, that is used until now by the serialization efforts around JSON-LD. Placing that here at a place where we can collaborate on it, before moving it to its official location.

PR #389, opened 2023-06-15
Add json-ld context

armintaenzertng:
this takes over from #388

The context is now up-to-date to the current commit (c75ce38881f2a2e63335d15d6228cf68df4f6e39)
and sorted alphabetically, starting with uppercase.

### DIFF
--- a/serialization/json-ld.md
+++ b/serialization/json-ld.md
@@ -1,0 +1,15 @@
+# JSON-LD serialization
+
+## JSON-LD context file
+The context file includes most profile names, class names, property names, 
+and vocabulary names from all profiles, as well as the "xsd" namespace and maps them to their respective URIs.
+It also defines the aliases "spdxId" and "type" for "@id" and "@type", respectively.  
+Exceptions have to be made for property names that are duplicated across different namespaces.
+If one of these namespaces is "https://spdx.org/rdf/v3/Core", it takes precedence in the context over other namespaces (TODO: What if the Core namespace is not among them?).
+If the duplicated property name is from a different namespace, it has to be prepended with the namespace when used in the body of JSON-LD (e.g. "security:locator": "https://some.locator").
+
+For properties, the range of their values is also denoted.
+Note that context expansion of the type only works for Literals and not Object nodes.
+This especially means that all Element objects and complex data types (CreationInfo, ExternalReference etc.) must explicitly state their type in the object.
+For vocabularies, a local @vocab is defined to prepend any value with the URI of the correct vocabulary type.
+Note that this does not check that the given vocabulary value and the resulting URI is actually defined in SPDX.
--- a/serialization/json_ld/context.json
+++ b/serialization/json_ld/context.json
@@ -1,0 +1,745 @@
+{
+    "AIPackage": "ai:AIPackage",
+    "Agent": "core:Agent",
+    "Annotation": "core:Annotation",
+    "AnnotationType": "core:AnnotationType",
+    "AnonymousPayload": "core:AnonymousPayload",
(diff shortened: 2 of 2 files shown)


zvr:
OK, but can we document it somehow in a README?

Do we have rules for what is included?

davaya:
In working on the [JSON examples](#375), I noticed that the context file has two "PresenceType" definitions: ai:PresenceType at line 141 and dataset:PresenceType at line 634.   I don't know if this is a conflict or not in JSON-LD, but in a physical data model classtypes and datatypes are both types and their names must be distinct.

Also there are no dataset model files defining a PresenceType, so line 634 may be a bug.

nishakm:
I'm getting a JSON parsing problem with this:
```
json.decoder.JSONDecodeError: Expecting value: line 8 column 1 (char 7)
```

armintaenzertng:
Hmm, I can*t see any problem with line 8 and https://jsonlint.com/ tells me that the file is valid JSON.

serialization/json-ld.md
```diff
@@ -0,0 +1,12 @@
+# JSON-LD serialization
+
+## JSON-LD context file
+The context file includes every profile name, class name, property name, 
+and vocabulary name from all profiles, as well as the "xsd" namespace and maps them to their respective URIs.
```
zvr, line 5: This info ("includes every...") is not enough (nor is it correct).

Based on what rule `"locator"` is `"core:locator"` and not `"security:locator"` ?

zvr (changes requested):

serialization/json_ld/context.json
```diff
@@ -0,0 +1,745 @@
...
+    },
+    "context": {
+        "@id": "core:context",
+        "@type": "xsd:string"
+    },
+    "copyrightText": {
+        "@id": "software:copyrightText",
+        "@type": "xsd:string"
+    },
+    "core": "https://spdx.org/rdf/Core/",
```
zvr, line 236: The URIs should follow the patterns defined in `rdf.md`
Please include the `/v3/` path part everywhere.

zvr (changes requested):

serialization/json_ld/context.json
armintaenzertng: This is a problem with the spec-parser. I have opened a PR: https://github.com/spdx/spec-parser/pull/60

armintaenzertng:
Update: My plan is to include the context generation and upload in the CI workflow, so that the context is kept up-to-date with the spec automatically.
This would be easiest by pushing it to gh-pages, but maybe there is a way to push it to the main repo without opening a PR every time.

zvr:
It should (also?) be pushed to the "generated" branch, together with all the generated .md files

armintaenzertng:
Just to be sure, that branch does not exist yet, does it?

serialization/json-ld.md
armintaenzertng: Good point, I updated the description. There is still an open question on how we handle duplicated property names that are not part of Core. Which namespace wins? No one?

armintaenzertng:
On the July 6 serialization meeting, we decided that duplicated property names (like `core:locator` and `security:locator`) should not get an entry in the context. That is, they should be `{"core:locator": "https://some.locator"}`.

goneall:
@zvr @armintaenzertng - Where are we with the PR?  Are all requested changes implemented?  If so, please mark as reviewed.

goneall:
The JSON-LD context file is now generated and uploaded as an artifact and stored in the github-pages branch, so this PR is no longer needed.  Closing.

PR #399, opened 2023-06-28
Restructure licensing profile

armintaenzertng:
This PR summarizes the discussion we had on the tech team call yesterday (June 27).
This moves all classes used for more in-depth modeling of licenses into the `ExpandedLicensing` profile, keeping only `AnyLicenseInfo`, `LicenseExpression` as well as a new `SimpleCustomLicense` class in the `LicensingProfile`.

The new structure is open to discussion. I believe some descriptions have to be rewritten.


zvr:
I propose `SimpleCustomLicense` be renamed to `SimpleLicensingText` or something.

Then this can be used for License texts as well as for Addition texts. Otherwise, we would need a `SimpleCustomAddition` in the Licensing profile, which will be identical to `SimpleLicense` (only a `string` property for the text).

model/Licensing/Classes/SimpleCustomLicense.md
```diff
@@ -0,0 +1,26 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SimpleCustomLicense
```
zvr, line 3: `SimpleLicensingText`

model/Licensing/Classes/SimpleCustomLicense.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Summary
+
+A license that is not listed on the SPDX License List.
+
+## Description
+
+A SimpleCustomLicense represents a License that is not listed on the SPDX License
```
zvr, line 11: same changes

model/Licensing/Classes/SimpleCustomLicense.md
```diff
@@ -0,0 +1,26 @@
...
+List at https://spdx.org/licenses, and is therefore defined by an SPDX data
+creator.
+
+## Metadata
+
+- name: SimpleCustomLicense
```
zvr, line 17: `SimpleLicensingText`

zvr (changes requested):

model/Licensing/Classes/SimpleCustomLicense.md
armintaenzertng: I took the liberty to also change the corresponding property name, I hope that was in your interest, too.

zvr (approved):

model/Licensing/Classes/SimpleLicensingText.md
```diff
@@ -0,0 +1,26 @@
...
+## Description
+
+A SimpleLicensingText represents a License or Addition that is not listed on the SPDX License
+List at https://spdx.org/licenses, and is therefore defined by an SPDX data
+creator.
+
+## Metadata
+
+- name: SimpleLicensingText
+- SubclassOf: AnyLicenseInfo
```
goneall, line 18: I'm not sure if we want `SimpleLicensingText` to be a subclass of `AnyLicenseInfo` - if someone is making a declared license or concluded license, should we restrict it to always be either a license expression string or an expanded license?  If you want to reference a custom license or addition, I think you would want to use a license expression string with the ID of the `SimpleLicensingText` and keep the `SimpleLicensingText` as a direct subclass of `Element`.  Otherwise, it would be more complex to parse in the simple case and it would be confusing when to use the `SimpleLicensingText` or the `CustomLicense`.

model/Licensing/Licensing.md
```diff
@@ -8,6 +8,8 @@ Additional metadata relating to software licensing.
 
 ## Description
 
+TODO: This text has to be reworked to reflect the new structure.
+
 The Licensing namespace defines metadata fields applicable to software
 licensing. It also defines the classes and properties that comprise the
 SPDX License Expression syntax and that relate to the SPDX License List.
```
goneall, line 15: This makes me think we should rename `ExpandedLicense` to `Licensing` with the same description above and rename the current `Licensing` to `SimpleLicensing`.

model/Licensing/Classes/SimpleLicensingText.md
```diff
@@ -0,0 +1,26 @@
...
+- Instantiability: Concrete
+
+## Properties
+
+- simpleLicensingText
```
goneall, line 23: Could we just use the `licenseText` property and move it back from the `ExpandedLicense` to `Licensing`?  It seems like it has similar semantics

goneall (commented):
More of a request for a discussion on my comments.

goneall:
Thanks @armintaenzertng for creating this - overall structure looks good based on our discussion.  A couple of review comments to consider / discuss.

model/Licensing/Classes/SimpleLicensingText.md
armintaenzertng: I agree that `SimpleLicensingText` is better off subclassing `Element`.

zvr: I also agree.

model/Licensing/Classes/SimpleLicensingText.md
armintaenzertng: `licenseText` is used in the ExpandedLicense profile, so if we move it we would have to reference it from another profile. That would be a first, I think. I'm not sure if we currently support this. @zvr?
I intentionally renamed this so that we don't get the same property name in two different profiles (we had some discussion a while back that we would try to avoid that if possible).

model/Licensing/Licensing.md
goneall: > Maybe it will be clearer to have SimpleLicensing and ExpandedLicensing profiles.

I like this naming.  In `SimpleLicensing` we could refer to `ExpandedLicensing` for the detailed definitions.

I would suggest waiting on any renaming until @swinslow has a chance to weigh in.

model/Licensing/Classes/SimpleLicensingText.md
goneall: > so if we move it we would have to reference it from another profile

I think this is OK, but I'll defer to @zvr for a definitive answer on this

> I intentionally renamed this so that we don't get the same property name in two different profiles (we had some discussion a while back that we would try to avoid that if possible).

Agree we should not use the same property name in 2 different profiles - I would propose referring to the property in a different profile as long as it is supported

model/Licensing/Classes/SimpleLicensingText.md
armintaenzertng: done

model/Licensing/Classes/SimpleLicensingText.md
zvr: We could do it (as we could do almost anything), especially since there is no way to use `ExpandedLicensing` profile without using the `SimpleLicensing` one.
But I think it would introduce complexity in tooling (and in spec-parser) to have classes in a profile having properties from other profiles. Remember, they all end up with IRIs with namespaces.

To Gary's last point "we should not use the same property name in 2 different profiles", we're already past this stage, as you all know: cf. `locator`.

goneall: > But I think it would introduce complexity in tooling (and in spec-parser) to have classes in a profile having properties from other profiles. Remember, they all end up with IRIs with namespaces.

@zvr We already have `Software` classes referencing other classes (e.g. `SoftwareArtifact.concludedLicense` refers to a property in a namespace other than `SoftwareArtifact`).  It this situation more complex?

Since the `licenseText` and `simpleLicensingText` are semantically the same, it would be useful for the RDF users to use the same property name.  BTW - this is solveable with different property names by using `owl:sameAs`, but it would not be as straight-forward as just using the same property name.

zvr: @goneall the example you cited will be replaced by a Relationship :-P

Do we have another?

goneall: `/Core/PositiveIntegerRange` referenced from `Software/Snippet` is another example.

goneall:
@armintaenzertng - Pinging you on this one - can you go back through the review comments and update what you agree with and (re)comment on any outstanding?

I just bumped into this in my java code generator and would like to merge.

model/Licensing/Licensing.md
armintaenzertng: I implemented the renaming.

model/Licensing/Classes/SimpleLicensingText.md
armintaenzertng: The `licenseText` property is now moved to `SimpleLicensing`, while also still used in the `ExpandedLicensing` profile.

model/ExpandedLicense/Classes/WithAdditionOperator.md
```diff
@@ -18,7 +18,7 @@ Syntax by the `WITH` operator.
 ## Metadata
 
 - name: WithAdditionOperator
-- SubclassOf: AnyLicenseInfo
+- SubclassOf: Licensing/AnyLicenseInfo
```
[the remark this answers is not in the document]
armintaenzertng: fixed

armintaenzertng:
@goneall, @zvr, I implemented all outstanding change requests, please take another look.

goneall (approved):
Thanks @armintaenzertng

armintaenzertng:
Found and fixed another little bug.

swinslow:
I have a few concerns about the way this is currently addressed. I'm not going to stand in the way of merging it as a next draft, but from an initial read, these are the points that I think should be addressed prior to the actual 3.0 release:

* As currently structured, it looks like for validation purposes the only requirement is that `LicenseExpression` should be an xsd:string. My assumption is that it additionally would only be valid if that string conforms to the [License Expression Syntax](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/). I'm not sure how we express that in the spdx-3-model repo syntax, but I think that's an important thing for us to retain if all the pieces of the model for license combinations are being moved to a separate profile.

* Relatedly, I think part of the expression being valid is to be able to ensure that the license identifiers are valid, e.g. that any non-LicenseRef's are actual identifiers on the License List. I'm not sure whether that's feasible to do under the current structure without leveraging the ExpandedLicensing profile, if ListedLicense and ListedLicenseException are over in that profile.

* I am worried that it's confusing to have a `simpleLicensingText` class in the SimpleLicensing profile which is comparable to, but distinct from, `CustomLicense` in the ExpandedLicensing profile.

swinslow:
On the first point above, I should note that @goneall has filed #440 to track clarifying this.

goneall:
Per discussion with @swinslow - I'll go ahead and merge this PR and we can discuss the outstanding issues raised [in the comment above](https://github.com/spdx/spdx-3-model/pull/399#issuecomment-1654366122) after merging.

PR #401, opened 2023-06-29
Mandatory creationInfo

davaya:
PR #273 was merged into the model without significant discussion at either the tech team or serialization subteam, and with a rationale that needs to be validated.

Since the beginning of SPDX v3 we have designed the model without in-depth discussion of serialization, leading to an implicit assumption that the model is serialization-independent.  273 violates that assumption, so now we need discuss and agree on what the model represents: 

1) Elements after they have been read into an application, or
2) Elements before they have been read into an application

The model diagram shows that every element must have creationInfo:

image

which represents the "meaning" of Element - no element can exist unless it has been created, and every application must be able to know the creationInfo of every Element it processes.  To make it optional says that there is a use case in which having an Element but not knowing how it was created makes sense.

This PR reverts creationInfo minCount back to 1 to align with the diagram and the use-case-based understanding that Elements must have creationInfo.  It is a call for discussion and consensus-based decision on whether the model represents use cases (logical Elements) or serialized data, and therefore whether the model is independent of serialization.

### DIFF
--- a/model/Core/Classes/Element.md
+++ b/model/Core/Classes/Element.md
@@ -41,7 +41,7 @@ and inter-relatable content objects.
   - maxCount: 1
 - creationInfo
   - type: CreationInfo
-  - minCount: 0
+  - minCount: 1
   - maxCount: 1
 - verifiedUsing
   - type: IntegrityMethod


zvr:
The diagram is not the "single source of truth", has not been kept up to date with various changes, and should not be used to justify changes.

Nevertheless, I agree with this PR, reverting back to having mandatory CreationInfo on all Elements.

We can definitely discuss how to save bytes in a serialization by omitting duplicate data in parallel.
I even welcome a discussion on whether CreationInfo should be mandatory for every Element in the model.

But I am very much in favor of returning to the commonly agreed model for a start and then work from there.

zvr (approved):

goneall (approved):
I also agree we should switch this back since the previous change to make it optional did not follow our governance practices for review.

Issue #390, opened 2023-06-15
Mapping the namespaceMap to serialization specific namespaces


goneall:
This came up in the serialization call on 15 June 2023.

The `ElementCollection` class has a `namespaceMap` which is used to expand/compact URI's used as identifiers within the `ElementCollection`.

Different serialization formats have different "native" mechanisms for expanding and compacting the URI's.  Since `ElmenetCollection` can contain other `ElementCollection`s all the serializations need to support nesting of the namespace mappings.  We will need to document how this works for each of the serialization formats we decide to support:

- [ ] JSON-LD
- [ ] ... other formats TBD ... (don't read anything into your favorite format not being listed, we'll add them once the serialization group formally decides it will be supported in SPDX 3.0)

goneall:
Some additional context.

On the call, I incorrectly stated that that the serialization formats didn't support nesting.  on further research, it is possible to nest these.  In JSON-LD - you can attach a context to a specific property as [described here](https://www.w3.org/TR/json-ld/#scoped-contexts).  For XML, the namespaces are properties of the XML element which can be nested (see this [description](https://www.w3schools.com/xml/xml_namespaces.asp)).

It looks like there is an implementation issue in that not many of the RDF libraries support nesting namespaces - you can only declare the namespace for the entire model.  This may be solvable to serializing each `ElementCollection` independently then somehow manually merging them.  Not very convenient - but possible.

For serialization formats that do not natively support namespaces, we will need to define how these namespaces work.  Some of the data modeling @davaya worked on may be helpful here.

maxhbr:
> It looks like there is an implementation issue in that not many of the RDF libraries support nesting namespaces [...]

So that would reduce the already small set languages with Libraries even further? Is then only Java left? We should avoid using such complex and niche features.

armintaenzertng:
[...]

PR #403, opened 2023-06-30
Clarify namespacemaps used in SPDX documents and collections

goneall:
This PR is an attempt to solve issue #390 by making moving the namespaceMap to SPDX Documents rather than all collections and redefining the semantics to be specific to the round-tripping issue for serialization namespaces

### DIFF
--- a/model/Core/Properties/originalNamespaces.md
+++ b/model/Core/Properties/originalNamespaces.md
@@ -1,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# namespaces
+
+## Summary
+
+Provides a NamespaceMap representing the NamspaceMap used for the original serialization of an SpdxDocument.
+
+## Description
+
+This field is used for providing a reliable mechanism in reproducing an original serialization for a given SPDX Document.
+
+This field is not intended for implementing or representing the NameSpaceMap for all serializations where this SPDX document may be found.
+
+## Metadata
+
+- name: originalNamespaces
+- Nature: DataProperty
+- Range: NamespaceMap
+
--- a/model/Core/Properties/originalNamespaces.md
+++ b/model/Core/Properties/originalNamespaces.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provides a NamespaceMap representing the NamspaceMap used for the original serialization of an SpdxDocument.
+Provides a NamespaceMap representing the NamespaceMap used for the original serialization of an SpdxDocument.
 
 ## Description
 
--- a/model/Core/Properties/originalNamespaces.md
+++ b/model/Core/Properties/originalNamespaces.md
@@ -8,9 +8,9 @@ Provides a NamespaceMap representing the NamspaceMap used for the original seria
 
 ## Description
 
-This field is used for providing a reliable mechanism in reproducing an original serialization for a given SPDX Document.
+This field is used for providing the ability to reproduce an original serialization for a given SPDX Document, especially if the serialization format does not natively support namespace mapping.
 
-This field is not intended for implementing or representing the NameSpaceMap for all serializations where this SPDX document may be found.
+This purely optional field is not intended for implementing or representing the NameSpaceMap for all serializations where this SPDX document may be found.
 
 ## Metadata
 
--- a/model/Core/Classes/SpdxDocument.md
+++ b/model/Core/Classes/SpdxDocument.md
@@ -17,8 +17,13 @@ Commonly used when representing a unit of transfer of SPDX Elements.
 - SubclassOf: Bundle
 - Instantiability: Concrete
 
+=======
+## Properties
+
+- originalNamespaces
+  - type: NamespaceMap
+
 ## External properties restrictions
 
 - /Core/Element/name
   - minCount: 1
-
--- a/model/Core/Classes/SpdxDocument.md
+++ b/model/Core/Classes/SpdxDocument.md
@@ -17,7 +17,6 @@ Commonly used when representing a unit of transfer of SPDX Elements.
 - SubclassOf: Bundle
 - Instantiability: Concrete
 
-=======
 ## Properties
 
 - originalNamespaces


model/Core/Properties/originalNamespaces.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# namespaces
```
armintaenzertng, line 3:
```suggestion
# originalNamespaces
```

model/Core/Properties/originalNamespaces.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Summary
+
+Provides a NamespaceMap representing the NamspaceMap used for the original serialization of an SpdxDocument.
+
+## Description
+
+This field is used for providing a reliable mechanism in reproducing an original serialization for a given SPDX Document.
+
+This field is not intended for implementing or representing the NameSpaceMap for all serializations where this SPDX document may be found.
```
armintaenzertng, line 13:
```suggestion
This field is not intended for implementing or representing the NamespaceMap for all serializations where this SPDX document may be found.
```

model/Core/Properties/originalNamespaces.md
armintaenzertng: I'd even consider using `originalNamespaceMap` instead of just `originalNamespaces`.

armintaenzertng (commented):
Thanks for the clarifying words, Gary! However, as far as I understand it, there can be multiple `SpdxDocument`s per payload, so I think this does not solve the issue I describe in [this comment](https://github.com/spdx/spdx-3-model/issues/390#issuecomment-1604021781).

goneall:
> Thanks for the clarifying words, Gary! However, as far as I understand it, there can be multiple `SpdxDocument`s per payload, so I think this does not solve the issue I describe in [this comment](https://github.com/spdx/spdx-3-model/issues/390#issuecomment-1604021781).

You are correct.  I think of the SPDX Document as a "unit of transfer" of SPDX information where there would only be one SPDX Document per payload for the "original" creation, although downstream users may bundle multiple SPDX documents in a different payload or collection.  Hence defining the namespace map to be the "original".

model/Core/Properties/originalNamespaces.md
goneall: Agree

goneall:
@zvr - I'm getting the following error from the CI - let me know any advice on tracking it down:

```
ERROR   : Duplicate `Property` object found: '{name}:{_prop.name}'
ERROR   : Spec not parsed successfully.
Error: Process completed with exit code 1.
```

Thanks

davaya:
There can be more than one SpdxDocument per payload, but only one of them refers to the payload it is contained in. (That's not strictly true, someone could create 5 SpdxDocuments referring to the same payload, but that would be stupid.)

Any others refer to other payloads, allowing hashes of the others to be verified and their elements imported.  A use case would be an Sbom with 10 elements split over two payloads.

model/Core/Properties/originalNamespaces.md
```diff
@@ -0,0 +1,20 @@
...
+
+# originalNamespaces
+
+## Summary
+
+Provides a NamespaceMap representing the NamspaceMap used for the original serialization of an SpdxDocument.
+
+## Description
+
+This field is used for providing a reliable mechanism in reproducing an original serialization for a given SPDX Document.
```
maxhbr, line 11: instead of `reliable mechanism` for me it feels more like a `hint`.

We probably could also highlight that its support is optional? Especially that it can be omitted, even if some magic shortened IDs while serialization. So, one could serialize using some namespaceMap like features, but is not forced to store this information in that map? Might be helpful to simplify implementation. 


Especially if someone "mints" the SpdxDocument and later wants to serialize it. Otherwise, one would be forced to do the namespace choices before minting the element. And one would need to do the same choice for all serializations.

maxhbr (approved):
Looks good and can be merged. But would like to "downgrade" the `originalNamespaces` from a `reliable mechanism` to just an optional `hint`.

model/Core/Properties/originalNamespaces.md
maxhbr: It being just a hint, also resolves the concerns above.

maxhbr (changes requested):
I actually changed my mind. I would prefer to have two PRs that are discussed independently:

1. drop `namespacesMap` now from the model
2. add some functionality to store the `namespacesMap`, that is supported by existing serializations. We can wait with that till later as it is just an optimization. And later we can actually test our assumptions.

goneall:
> I actually changed my mind. I would prefer to have two PRs that are discussed independently

It's a bit easier for me to agree to both together since the original motivation for adding the namespaces was to support round-trip fidelity for the other format.  That being said, waiting until a decision is made on the formats we plan to support does make sense.

I'll split into 2 PR's.

goneall:
I created PR #411 to remove the namespaces.

@maxhbr - I'll wait until PR #411 is agreed to and merged before opening the second PR.  In the interim, I'll address some of the feedback and change this status to draft.

davaya:
>1. drop namespacesMap now from the model
> 2. add some functionality to store the namespacesMap, that is supported by existing serializations.

I agree.

An analogy would be JSON or XML data - whitespace is insignificant.  An application can call json.dump with an indent=2 parameter, but there is no API to tell dump to look for an indent parameter in the data to be serialized.  With parsing there's no API to return the indentation used to create the data.

So if we want to read namespaceMap when parsing and set it when dumping, the serialization spec for each format will need to define an API to do it.

goneall:
Note related PR #431

goneall:
> I don't care whether it goes back into SpdxDocument because it serves no purpose but causes no harm.

@davaya - There is a specific use case I would like to solve for:

- SPDX producer produces a serialized form of an SPDX collection of elements using a serialization method involving namespaces
- An SPDX consumer deserializes into their database which follows the model but isn't the exact serialized stream of bytes
- The SPDX consumer would like to re-produce something close the original serialized form - in particular keeping the same namespace mappings

This is the use case I'm trying to accomplish with the SPDX Document class.  The information needs to be in the model itself in order to be preserved between different deserialize / serialize forms.

This use case is important for backwards compatibility with SPDX 2.3.

goneall:
> ... add some functionality to store the namespacesMap, that is supported by existing serializations. We can wait with that till later as it is just an optimization. And later we can actually test our assumptions.

@maxhbr - Can you comment on any concerns with this PR w.r.t. adding back in the functionality to store namespace maps?

kestewart (approved):
Adding back in Namespaces as agreed in prior discussion.

maxhbr:
My concern: it is an optional feature and might or might not be serialized. And it might or might not be correctly recovered from parsing a serialized file. Therefore parsing such an SpdxDocument Element could result in inconsistent content. Serializing to multiple formats and de-serializing with multiple tools that do or do not support this feature will result in different SpdxDocument objects. Which will mess up all hashing and consistency checks. Can we clarify the `optional` part first?

goneall:
> My concern: it is an optional feature and might or might not be serialized. And it might or might not be correctly recovered from parsing a serialized file. Therefore parsing such an SpdxDocument Element could result in inconsistent content. Serializing to multiple formats and de-serializing with multiple tools that do or do not support this feature will result in different SpdxDocument objects. Which will mess up all hashing and consistency checks. Can we clarify the `optional` part first?

@maxhbr I'm not sure I completely understand the concern.  I think we've always had the issue that if you deserialize and reserialize an SPDX document, it would have different hashes - that's something the canonicalization team was chartered to address.

From my perspective, it is acceptable for intermediate suppliers to not support the round trip use case - I just want to make it possible to support if all parties are interested in supporting it - so optional is fine.

maxhbr:
> @maxhbr I'm not sure I completely understand the concern. I think we've always had the issue that if you deserialize and reserialize an SPDX document, it would have different hashes - that's something the canonicalization team was chartered to address.

I see issues when serializing and deserialize and then have an Element with a slightly different content.

goneall:
> I see issues when serializing and deserialize and then have an Element with a slightly different content.

The element being the `SpdxDocument`?  Maybe something we should discuss RT in one of the upcoming meetings.

davaya:
> > I see issues when serializing and deserialize and then have an Element with a slightly different content.
> 
> The element being the `SpdxDocument`? Maybe something we should discuss RT in one of the upcoming meetings.

Every `SpdxDocument` element is created by someone at some time, and it is created to describe a payload.  So there can be multiple different `SpdxDocument` elements describing the same payload but they must have different SpdxIds and CreationInfos.

In order for this use case:
```
Payload1 -> database -> Payload2
```
to have any effect Payload1 must include the SpdxDocument describing Payload1.  If it does so and SpdxDocument has NamespaceMap, then an application can reproduce Payload2 that is bit-for-bit identical to Payload1 if you assume everything else in canonicalization has been fully defined and followed by the implementations.

I still don't see this as an interesting use case (just as I don't see the benefit of reproducing JSON documents with identical indentation), but I don't object to:
1) including NamespaceMap information in SpdxDocument, and
2) including a self-referencing SpdxDocument element in a payload.

maxhbr:
This is currently described as:

> This purely optional field ...

so there is a choice. Either a generator or parser can choose to not support or use it. Or do I misunderstand this part? So, that means that after generation and parsing, depending on choices, one has a different result?

goneall:
> This is currently described as:
> 
> > This purely optional field ...
> 
> so there is a choice. Either a generator or parser can choose to not support or use it. Or do I misunderstand this part? So, that means that after generation and parsing, depending on choices, one has a different result?

An SPDX Document is an element, so once it is "minted", it can not change.  What I intended by "purely" optional is that when minting an SPDX document, you do not need to include this property - as in it is not required.  If it is present, then it should always be present in all serializations of the same element ID.

@maxhbr - I think we need to discuss realtime - I don't think my description is anywhere adequate, but I'm not understanding the issues.

davaya:
@maxhbr 
>  Either a generator or parser can choose to not support or use it.

I think Gary's explanation is clear.

* A generator can choose not to support or use it.
* A parser must correctly parse the elements in a document. The parsed value of an SpdxDocument element must include all properties that were serialized into that element regardless of data format.

model/Core/Properties/originalNamespaces.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Summary
+
+Provides a NamespaceMap representing the NamspaceMap used for the original serialization of an SpdxDocument.
+
+## Description
+
+This field is used for providing the ability to reproduce an original serialization for a given SPDX Document, especially if the serialization format does not natively support namespace mapping.
+
+This purely optional field is not intended for implementing or representing the NameSpaceMap for all serializations where this SPDX document may be found.
```
maxhbr, line 13: some questions regarding semantics:

- if this is not provided, is the serialization allowed to use any `NameSpaceMap` feature?
  - or more specific: are only the explicitly stated `NameSpaceMap` entries allowed?
- if there is some `NameSpaceMap` like feature of a serialization used, should it be added to the `NameSpaceMap` of the corresponding `SpdxDocument`?
- What are the semantics, if a serialized blob contains not exactly one `SpdxDocument`?

Whether the `optional` works depends somewhat on these questions.

And I see two levels of optional here. The entity minting the SpdxDocument might choose to not add Namespaces and the serializer might choose to not use the NameSpaceMap in the Document. Are both optional? 


maybe not so relevant:
- How would this approach handle potential inconsistent constraint on mappings? There might be keys, that are allowed in some serialization but disallowed in another one.

goneall:
@maxhbr 

> if this is not provided, is the serialization allowed to use any NameSpaceMap feature? or more specific: are only the explicitly stated NameSpaceMap entries allowed?

Any subsequent serialization is allowed to use any NameSpaceMap - as long as the resulting URI is the same.

> if there is some NameSpaceMap like feature of a serialization used, should it be added to the NameSpaceMap of the corresponding SpdxDocument?

It should be added if the creator of the SpdxDocument desires the ability to recreate the original document utilizing the same NameSpaceMap.  I would probably say the creator "can" use the NameSpaceMap.

What are the semantics, if a serialized blob contains not exactly one SpdxDocument?

The SpdxDocument should relate to one and only one "original" serialization (this could be clarified in the description).  Subsequent serializations may aggregate multiple SpdxDocuments and represent each of the original serializations with separate SpdxDocuments within the same serialization.

davaya:
> What are the semantics, if a serialized blob contains not exactly one SpdxDocument?

* A serialized blob is the controlling entity - the blob may use an IRI compaction mechanism or not.
* The identical set of elements may be serialized without IRI compaction in one blob, and serialized with compaction in a different blob, and serialized with a different compaction in a third blob.
* If a serialized blob contains multiple SpdxDocuments, each describes a specific blob.  The blobs may contain the same, overlapping, or disjoint set of elements.

maxhbr:
> The SpdxDocument should relate to one and only one "original" serialization (this could be clarified in the description). 

This means, that one mints a different `SpdxDocument` for every serialization? If I generate data and want to serialize it in multiple formats, I need to mint multiple `SpdxDocument` elements? I previously thought that one would just create one `SpdxDocument` independent of serialization and would that then pass to different handlers that serialize it.


I still think that this is an issue that we should tackle very late, when we are clear how our serializations will look like and when we are able to test our assumptions on real data.

davaya:
> This means, that one mints a different SpdxDocument for every serialization?

If one chooses to create an SpdxDocument at all, it refers to a single minted blob.  If the SpdxDocument contains verifiedBy (which is its primary purpose), every blob will have its own verification information.

Having a single SpdxDocument that can verify multiple blobs would be possible only if we develop canonical serialization.  I would guess that the canonical form would not use compaction.

maxhbr:
> Even if SpdxDocument does not contain verifiedBy, handlers must be able to serialize a set of elements that includes zero or more SpdxDocument elements. Data to be serialized is the "data plane", serialization parameters are knobs on the "control plane" and the compaction knob can be turned without serializing SpdxDocuments.

How would one compute the hash of something containing the hash a property? If the minted `SpdxDocument` Element is part of the serialization, it can not contain the hash of the blob.

maxhbr:
@davaya , you say that:

>  If the SpdxDocument contains verifiedBy **(which is its primary purpose)**, ...

so the "verifiedBy" is its primary purpose. But the description of the `SpdxDocument` is just:

> An SpdxDocument assembles a collection of Elements under a common string, the name of the document. Commonly used when representing a unit of transfer of SPDX Elements.

I think there is a mismatch between your understanding and the current definition. I think that we need to align on the description of SpdxDocument first.

davaya:
> I think there is a mismatch between your understanding and the current definition. I think that we need to align on the description of SpdxDocument first.

Yes, the serialization sub-team and the tech team need a common understanding of serialization and the model, and until recently the model made no distinction between logical element values and serialized blobs.  Now it does, and discussion will elaborate on how serialized values and the logical model relate to each other.

> How would one compute the hash of something containing the hash a property?

Normally security functions have a payload with protected and unprotected header information wrapped around it.  In cases where envelope values (such as a hash) are to be included inside the payload, the normal approach is to assign a fixed value (e.g., zero or an empty string) to the payload field when computing the signature/hash, then filling in the computed value before transmission.  Validation reverses the process, nulling the payload field before validating.

This is only necessary if serialized blobs contain their own security values, which is why its easier for serialized blobs to only contain security values for *other* blobs, as is the use case for "importing" elements from one blob to another.  Another topic for the model/serialization discussion.

goneall:
@maxhbr 
> This means, that one mints a different SpdxDocument for every serialization?

Absolutely not - there is only one `SpdxDocument` minted once.  Once minted, it can be included in other serializations.

The verifiedBy is an interesting topic, but not the primary reason I wanted have an `SpdxDocument` class in the model.

We definitely need to discuss this real-time - the thread is getting very long and we're mixing a lot of different issues.

maxhbr:
> We definitely need to discuss this real-time - the thread is getting very long and we're mixing a lot of different issues.

I agree, but today and next Tuesday we have public holidays. Will miss the meetings.

In my opinion, the reasons for me derailing the discussion is, that semantics of SpdxDocument are not clear. I think we need to start earlier and at a later point this problem can be solved.

davaya:
>> This means, that one mints a different SpdxDocument for every serialization?
> Absolutely not - there is only one SpdxDocument minted once. Once minted, it can be included in other serializations.

The `SpdxDocument Element` would then contain only serialization-independent information about multiple `SpdxDocuments` serialized as payloads in different formats, i.e. the element ids of the elements included in the payload.

If there is any serialization-dependent information contained in the `SpdxDocument Element` (such as verifiedBy), then the Element would need to be specific to the serialized document.

maxhbr:
Proposal:

I suggest that the parser that wants to persist the information for reproducing the serialization should create a new Element which is minted on parsing, which

* parses the namespace map
* contains the hash(es) of the binary blob that was parsed
* persists other information like: origin of the file and file name

and this Element is then linked to the parsed content.

This new type would need to be added to the Model, maybe in a small profile.

davaya:
@maxhbr +1  That "new type" is already in the model as SpdxDocument - it can be minted by the parser of a payload.

davaya:
> I think the SpdxDocument is minted and added on generation of the serialized blob.

That's one use case.

But the intent of the "X" element (SpdxDocument element, Payload element, File element) is that it can be generated by user Y who parses a Payload created last year.

Maybe we can agree on the on-the-fly element generation as a hard requirement, and then decide what the name of the X element should be, as Gary suggests.

goneall:
> Proposal:
> 
> I suggest that the parser that wants to persist the information for reproducing the serialization should create a new Element which is minted on parsing, which
> 
> * parses the namespace map
> * contains the hash(es) of the binary blob that was parsed
> * persists other information like: origin of the file and file name
> 
> and this Element is then linked to the parsed content.
> 
> This new type would need to be added to the Model, maybe in a small profile.

In general I like this approach as it removes the confusion and ambiguity between the serialization namespace map and the model version of the namespace map.  It is similar to the approach where we handle this in the serialization spec but does provide a model object to store the information - even if it isn't every "serialized".

The edge case that is not covered is if there is an intermediate deserialization / serialization.  In that case we could loose the original name space map if the intermediate chose not follow the same namespace map.

Not sure if this edge case is important, but I thought I would point this out.

model/Core/Properties/originalNamespaces.md
goneall: > if this is not provided, is the serialization allowed to use any NameSpaceMap feature?

If your question is related to the native namespace map used by the serialization format (not some property serialized data), there is no restriction whether or not the original namespaces is provided.

> if there is some NameSpaceMap like feature of a serialization used, should it be added to the NameSpaceMap of the corresponding SpdxDocument?

If the creator of the SPDX data wishes to support namespace round tripping, yes.

> What are the semantics, if a serialized blob contains not exactly one SpdxDocument?

Each SpdxDocument object represents one serialization of SPDX data.  If you have a serialization which makes copies of multiple serializations which support the round tripping of namespace data, you can have multiple SPDX documents.  Each SpdxDocument properties can then be used to recreate the original SPDX namespaces.

> And I see two levels of optional here. The entity minting the SpdxDocument might choose to not add Namespaces and the serializer might choose to not use the NameSpaceMap in the Document. Are both optional?

Yes - These are to support recreating the original namespace map.  The creator does not have to support this and can omit the necessary data.  The consumer can ignore the data and create whatever serialization specific namespace map they want.  This is only for the case where the producer wishes to have the namespace map preserved and the consumer would like to recreate it.

> How would this approach handle potential inconsistent constraint on mappings? There might be keys, that are allowed in some serialization but disallowed in another one.

The producer would just record the namespace mapping they used.  If the consumer can't handle the namespace mapping for their chosen serialization format, they would have to handle that in a method specific to the output serialization format.

davaya:
> ... does provide a model object to store the information - even if it isn't ever "serialized".

> The edge case that is not covered is if there is an intermediate deserialization / serialization. In that case we could loose the original name space map if the intermediate chose not follow the same namespace map.
> 
> Not sure if this edge case is important, but I thought I would point this out

You are describing two different use cases:
1) The model defines a namespace map element that is independent of serialization formats and is independent of payload instances - it can be used by a payload containing elements A, B, and C and by another payload containing elements D, E, and F.  That "golden namespace map" element can always be referenced, it wouldn't contain any payload-specific information like hash or downloadLocation, so there is no intermediate serialization, not necessarily any serialization at all.
2) The model defines a payload-specific element that contains the namespace map used by the payload, along with the downloadLocation and hash of the payload and the SpdxIds of the elements in that payload

Does the consumer follow Max's proposal to create a "golden namespace map" element that can be used to serialize any set of elements in any data format?  In that case, the map is never lost

Or does the consumer follow Max's proposal to create a payload-specific element that lists the format-specific download location and hash of the payload and the SpdxIds A, B, and C?  In that case the map isn't lost either, because the "original namespace map" element that was created by parsing payload 1 can be serialized into intermediate payloads 2, 3, and 4 even if data format 2 doesn't carry namespace map in the serialized data (and thus can never be the source of a golden or original map).

model/Core/Properties/originalNamespaces.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# namespaces
+
+## Summary
+
+Provides a NamespaceMap representing the NamspaceMap used for the original serialization of an SpdxDocument.
```
lumjjb, line 7: typo: s/NamspaceMap/NamespaceMap/

goneall: Fixed

Thanks @lumjjb for the catch!

goneall:
Superseded by #490

PR #404, opened 2023-07-01
Remove Payload from logical model

davaya:
Based on serialization discussions, the logical model does not define Payloads and I took an action to remove it from the model.  This PR adjusts the diagram to indicate that Payload is an interface between logical elements and (opaque) data, and it removes Payload from the model files.


goneall (approved):
Thanks @davaya - I agree with the changes and the description of Payload in the diagram.

maxhbr (approved):
LGTM

PR #406, opened 2023-07-04
Consolidate the JSON-LD format

armintaenzertng:
While there is still discussion surrounding the JSON serialization, this is an attempt to consolidate the JSON-LD format as a pure RDF format. This does not mean it can't be reused as a JSON serialization in the future, but please keep JSON discussions out of this PR for now.

This should bring us closer to a first functioning serialization that people can implement and try out.


goneall (approved):
Looks good and agree we can include the RDF serialization JSON-LD as a starting example.

maxhbr (approved):
LGTM, that is a good next step

serialization/json_ld/examples/spdx_document4.json
```diff
@@ -0,0 +1,59 @@
...
+    },
+    {
+      "type": "Relationship",
+      "spdxId": "myNamespace:relationship1",
+      "from": "myNamespace:SPDXRef-Package",
+      "to": [
+        "myNamespace:file1",
+        "myNamespace:file2"
+      ],
+      "relationshipType": "contains"
```
zvr, line 56: This contradicts the last bullet in `rdf.md`.
Unless the context in #389 will be extended to have all enumerated values.

zvr (changes requested):
I won't go through all of the examples, but I would like to see enumerated values being correct.

serialization/json_ld/examples/spdx_document4.json
goneall: Just a note - in the SPDX 2.X spec, we used a convention of prefixing the enumeration value with the type for RDF values to make sure they are unique across the entire namespace (e.g. `relationshipType_contains`).  Without such a convention, we are likely to have collisions in the enumeration values.

zvr: @goneall the `rdf.md` file already says:

>  IRIs for an enumerated value are of the form: `http://spdx.org/rdf/v3/{Namespacename}/{Classname}/{Valuename}`

zvr: Ah, sorry, @armintaenzertng , I only quickly searched for `"contains"` in the context file and did not see anything, so I assumed it was not handled. Using `@vocab` is perfectly fine as a first step -- or even a final one.

serialization/rdf.md
```diff
@@ -11,3 +11,24 @@ SPDX data can be serialized in RDF. This can be saved in a variety of formats, l
 1. IRIs for an enumerated value are of the form: `http://spdx.org/rdf/v3/{Namespacename}/{Classname}/{Valuename}`
 
 
+## JSON-LD
+The main RDF serialization format of SPDX is JSON-LD, a format that is defined [here](https://www.w3.org/TR/json-ld11/).
```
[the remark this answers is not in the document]
armintaenzertng: I changed the sentence to be more neutral. We might want to add a subsection regarding the structure of the RDF graph?

zvr (approved):

armintaenzertng:
FYI: I deleted the alias `element` for `@graph` as it clashed with the already established key `element`. If someone has a better name for it, please come forth. :)
On the other hand, it might be consistent to have two top-level keys beginning with "@".

zvr:
I think this is fine for merging, and we can iterate improving it afterwards.

PR #407, opened 2023-07-05
Create SpdxId type for spdxId property

davaya:
Using a semantic type for Element identifiers enables software to distinguish between IRIs used as Element IDs and IRIs used for other purposes such as Package download location, package URL, and home page.

This allows NamespaceMap to be applied to serialized Element IDs such as Relationship to and from without affecting other IRI properties.

### DIFF
--- a/model/Core/Classes/Element.md
+++ b/model/Core/Classes/Element.md
@@ -24,7 +24,7 @@ and inter-relatable content objects.
 ## Properties
 
 - spdxId
-  - type: xsd:anyURI
+  - type: SpdxId
   - minCount: 1
   - maxCount: 1
 - name
--- a/model/Core/Classes/SpdxId.md
+++ b/model/Core/Classes/SpdxId.md
@@ -1,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SpdxId
+
+## Summary
+
+An Element Identifier
+
+## Description
+
+The Element spdxId property is an IRI that uniquely identifies an Element instance. 
+
+## Metadata
+
+- name: SemVer
+- SubclassOf: xsd:anyURI
+
+## Properties
--- a/model/Core/Classes/SpdxId.md
+++ b/model/Core/Classes/SpdxId.md
@@ -8,11 +8,12 @@ An Element Identifier
 
 ## Description
 
-The Element spdxId property is an IRI that uniquely identifies an Element instance. 
+The Element spdxId property is an IRI that uniquely identifies an Element instance.
+The SpdxId type distinguishes IRIs that are Element identifiers from IRIs used for other purposes.
 
 ## Metadata
 
-- name: SemVer
+- name: SpdxId
 - SubclassOf: xsd:anyURI
 
 ## Properties


goneall:
Although this would be convenient for some serializations, it would add complexity to others.

In RDF, you can always check the type of the subject to find out if it is an Element.

Also note that the rules for "subclassing" data types are a bit different - see https://www.w3.org/TR/xmlschema11-2/#dt-derived - I'm not sure if our current schema tooling supports deriving one datatype from another.

davaya:
This is the same issue as https://github.com/spdx/spdx-3-model/pull/377#issuecomment-1605395437.  We already have SemVer, DateTime, and MediaType datatypes in the model, and contrary to the assertion that it can't be done, they work fine, as the xsd example shows.

The issue is just a matter of wording - "subclassing" gives the impression that datatypes are subclassed from something.  They aren't.  They are defined as root types themselves, not subclassed from anything:

[CreationInfo](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/CreationInfo.md), DictionaryEntry, ExternalIdentifier and other datatypes aren't a "SubclassOf" anything.

[ExternalMap](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/CreationInfo.md), ExternalReference, PositiveIntegerRange and others are a "SubclassOf" "none", with means the same thing.

[SemVer](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/SemVer.md), DateTime and MediaType say they are "SubclassOf" xsd:string, but they aren't.  They are root datatypes like other datatypes, xsd calls them "[facets](https://www.w3schools.blog/xsd-restrictions-facets)" that have a "base" of xsd:string (as demonstrated in the XSD in #377) but the base is not a subclass relationship..

Get rid of the "SubclassOf" metadata item from all datatypes and the mistaken impression of subclassing goes away.

goneall:
I'm running into other issues with the subclassing conventions used - for example, in RDF, there is no class known as `none`.  There is a class known as `owl:thing` which should be used.

This is complex enough we should probably have a general discussion on:
- what are the conventions we want to use for `nature` and `subclass`?
- how does this get translated to the OWL and SHACL documents?
- do we want to take on defining the datatypes in an xsd document?

zvr:
@goneall please keep in mind that the human-editable markdown files were not meant to be used for anything else besides the spec-parser.
That program will produce lots of generated stuff: for example ontologies, but also complete markdown files.

I'm now seeing more and more attempts to parse these files, which I think it's a step to the wrong direction.

goneall:
> please keep in mind that the human-editable markdown files were not meant to be used for anything else besides the spec-parser.

@zvr - that is exactly my point - I didn't mean to imply we are to use the markdown files for anything beyond translating to other formats via the spec parser.  Since the usage of the markdown files are restricted, we have some freedom to define the semantics and syntax to meet the needs of the output schemas generated from the markdown files.

As it stands, I don't think creating subclasses for data types will generate the correct OWL / SHACL and we don't currently support XSD definitions of data types - so merging this PR in I don't think would result in anything useful.  Let me know my understanding is not correct.

davaya:
@goneall @zvr: I'd ask you both to look at [Relationship](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/b60b3de89af68ba94b53e49ed8f83baa67f1d8b6/model/generated_20230705T100948Z/Core/Relationship.md).

This is the logical value of a relationship element. which has start and end DateTimes. The logical value has nothing to do with any particular serialization, it is the answer to the question "what is the start time of this Relationship".  Regardless of whether RDF or Tag or JSON is being used, the user must be able to look at the serialized data and get the answer to that question, and it doesn't make sense to NOT have a name for what a DateTime looks like.

The question isn't whether tooling currently supports giving names to regex patterns, the question is whether it should.  It is a tooling defect to not do so.

goneall:
> @goneall @zvr: I'd ask you both to look at [Relationship](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/b60b3de89af68ba94b53e49ed8f83baa67f1d8b6/model/generated_20230705T100948Z/Core/Relationship.md).
> 
> This is the logical value of a relationship element. which has start and end DateTimes. The logical value has nothing to do with any particular serialization, it is the answer to the question "what is the start time of this Relationship". Regardless of whether RDF or Tag or JSON is being used, the user must be able to look at the serialized data and get the answer to that question, and it doesn't make sense to NOT have a name for what a DateTime looks like. Given the model markdown files, is the logical value the exactly correct list of questions that can be asked about a Relationship instance, not more nor fewer?
> 
> The question isn't whether tooling currently supports giving names to regex patterns, the question is whether it should. It is a tooling defect to not do so.

I agree it DateTime should be defined - I'm just confused about the use of `SubClass` of a data type. 

When I look at the generated OWL/SHACL file, I see:

```
core:DateTime a owl:Class,
        sh:NodeShape ;
    rdfs:comment """A Datetime is a string representation of a specific date and time.
It has resolution of seconds and is always expressed in UTC timezone.
The specific format is one of the most commonly used ISO-8601 formats.
Format restriction: pattern: ^\\d\\d\\d\\d-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\dZ$""" ;
    rdfs:subClassOf xsd:string ;
    ns0:term_status "Stable" .
```

I'm not completely sure, but I don't believe this is correct.  I don't think you want to have the type to be a subclass of `xsd:string` since `xsd:string` is a datatype.

For the purposes of generating schemas we can validate against, it looks like we should be using SHACL shapes and not RDFS Subclasses for this purpose.

I have to admit, I find this area to be rather complex and I could well be wrong.

Here's one reference to a discussion on a similar topic in [Stack Overflow](https://stackoverflow.com/questions/24817607/why-must-rdfdatatype-subclass-rdfclass-in-rdf).

goneall:
I spent some time trying to get a definitive answer on if you can subclass data types in RDF. 

It looks like I was wrong - you can subclass datatypes.   Based on this, I withdraw my objection and it looks like we are property treating DateTime property correctly.

I wasn't able to figure this out using google searches, but with a bit of help from Google Bard, I found an example.  Based on the dialog, I'm not too embarrassed that I didn't know this from the start - it seems Google Bard was also a bit confused ;)

Mostly for entertainment value, here's my Google Dialog.

Prompt: `In RDF, can you have a subclass of a datatype?`

```
No, you cannot have a subclass of a datatype in RDF. In RDF, a datatype is a class, and classes cannot have subclasses. The only way to create a new datatype in RDF is to extend an existing datatype. For example, you could create a new datatype called xsd:positiveInteger by extending the datatype xsd:integer.

The reason for this is that datatypes in RDF are not just sets of values. They are also classes, and classes in RDF have a specific meaning. A class in RDF is a set of resources that share some common property. In the case of datatypes, the common property is the set of values that the datatype can represent.

If you were able to have subclasses of datatypes, then you would be able to create a class that contains resources that are not values. This would violate the definition of a class in RDF.

However, you can still create new datatypes in RDF by extending existing datatypes. This is a common practice, and it allows you to create more specialized datatypes that meet your specific needs.

```

Which prompted my second question `I thought you could subclass classes in RDF, isn't this true?`

Response:

```
Yes, you can subclass classes in RDF. This is done using the rdfs:subClassOf property. For example, you could define a class called Animal and then subclass it with a class called Dog. This would mean that all dogs are also animals.

The rdfs:subClassOf property is a transitive property. This means that if class A is a subclass of class B, and class B is a subclass of class C, then class A is also a subclass of class C.

Subclassing classes in RDF is a powerful way to represent hierarchical relationships between classes. It allows you to define more complex ontologies that can be used to model real-world concepts.

I apologize for the confusion earlier. I thought you were asking about subclassing datatypes, which is not possible in RDF. However, subclassing classes is possible, and it is a very common practice.
```
My response: `Then how do you explain your answer on not being able to subclass datatypes since "a datatype is a class, and classes cannot have subclasses"?`

Bard:

```
You are correct. A datatype is a class in RDF, and classes cannot have subclasses. However, there is a distinction between datatypes and classes in RDF. Datatypes are a special type of class that represents sets of values. Classes, on the other hand, can represent any set of resources.

This distinction is important because it means that you cannot subclass a datatype in the same way that you can subclass a class. If you were able to subclass a datatype, then you would be able to create a class that contains resources that are not values. This would violate the definition of a class in RDF.

However, you can still create new datatypes in RDF by extending existing datatypes. This is a common practice, and it allows you to create more specialized datatypes that meet your specific needs.

To summarize, the following are the key points to remember:

Datatypes are a special type of class in RDF.
Classes cannot have subclasses.
You can create new datatypes in RDF by extending existing datatypes.
I hope this explanation is more clear. Let me know if you have any other questions.
```

My next prompt: `
In your last response, you state on you 2nd bullet "Classes cannot have subclasses.", but that is not true based on your previous response.`

Bard:

```
You are correct. I apologize for the confusion. Classes can have subclasses in RDF. I misspoke in my previous response.

To be more specific, classes in RDF can have subclasses if the subclasses are also classes. This means that you cannot subclass a datatype with a class, because datatypes are not classes. However, you can subclass a class with another class.

For example, you could define a class called Animal and then subclass it with a class called Dog. This would mean that all dogs are also animals.

I hope this explanation is more clear. Let me know if you have any other questions.
```

My next prompt: `So, how do you extend a datatype in rdf?`

Bard:

```
I'm unable to help you with that, as I'm only a language model and don't have the necessary information or abilities.
```

Prompt: `You said "You can create new datatypes in RDF by extending existing datatypes." - how do you do that?`

Bard:
```
Sure, I can help you with that.

To extend a datatype in RDF, you need to use the rdfs:subClassOf property. The rdfs:subClassOf property is used to indicate that one class is a subclass of another class. In the case of datatypes, this means that the new datatype is a subclass of the existing datatype.

For example, let's say you have an existing datatype called xsd:integer. You want to create a new datatype called xsd:positiveInteger that is a subclass of xsd:integer. You would do this by adding the following statement to your RDF vocabulary:

Code snippet
<http://example.org/datatypes/positiveInteger>
    rdfs:subClassOf <http://www.w3.org/2001/XMLSchema#integer> .
Use code with caution. [Learn more](https://bard.google.com/faq#coding)
This statement tells RDF that the xsd:positiveInteger datatype is a subclass of the xsd:integer datatype. This means that any value that is valid for the xsd:integer datatype is also valid for the xsd:positiveInteger datatype.

In addition to the rdfs:subClassOf property, you can also use the xsd:restriction property to further constrain the values that are valid for a datatype. For example, you could use the xsd:restriction property to specify that the xsd:positiveInteger datatype can only contain values that are greater than or equal to 1.

Here is an example of how you would use the xsd:restriction property to further constrain the values that are valid for the xsd:positiveInteger datatype:

Code snippet
<http://example.org/datatypes/positiveInteger>
    rdfs:subClassOf <http://www.w3.org/2001/XMLSchema#integer> ;
    xsd:restriction
        # Only allow values that are greater than or equal to 1
        {
            xsd:minInclusive "1" .
        }
Use code with caution. [Learn more](https://bard.google.com/faq#coding)
This statement tells RDF that the xsd:positiveInteger datatype is a subclass of the xsd:integer datatype, and that the values that are valid for the xsd:positiveInteger datatype must be greater than or equal to 1.

I hope this explanation is helpful. Let me know if you have any other questions.
```

goneall:
> It looks like I was wrong - you can subclass datatypes. Based on this, I withdraw my objection and it looks like we are property treating DateTime property correctly.

Now I'm not so sure - Bard may have been hallucinating - I couldn't find any spec or examples directly from the internet.

@davaya @zvr @sbarnum - can you provide any references to using `rdfs:subClassOf` for a data type?

goneall:
I found this [W3C note](https://www.w3.org/TR/swbp-xsch-datatypes/) that provides some context on using data types.

For reference, here's the [W3C description of data types](https://www.w3.org/TR/rdf12-concepts/#section-Datatypes) in the concepts document.

My reading of the above is data types need to be extensions of the XML datatypes and not using owl:subclasses.

I did find this reference to [extensible datatypes](https://www.w3.org/TR/owl2-syntax/#Datatype_Definitions) in OWL2.  Note that these extensible datatypes can not be used with literals - so we should probably avoid them.

Back to the conclusion we should not be subclassing datatypes (and an additional conclusion that Bard is hallucinating on the topic).

davaya:
The almost correct:

> Datatypes are a special type of class in RDF.
Classes cannot have subclasses.
You can create new datatypes in RDF by extending existing datatypes.

combined with the almost-correct correction:

> This means that you cannot subclass a datatype with a class, because datatypes are not classes.

gets close to my mental model of the explanation:
* Datatypes are a special type of class in RDF
* Datatype classes cannot have subclasses (while other classes can have subclasses)
* Datatype classes can be "extended" in RDF by adding restrictions to existing datatypes

The last point argues against using both "subclass" and "extend" when talking about datatypes, because "restrict" is technically correct.  DateTime is a "RestrictionOf" or "FacetOf" xsd:string, not an "ExtensionOf" or "SubclassOf".

I've never tried Bard, but I'm quite impressed with your exchanges - thanks for doing and posting them. https://www.reddit.com/r/quotes/comments/ds44r2/the_marvel_is_not_that_the_bear_dances_well_but/

As for how to restrict a dataclass in RDF, @sbarnum will have to answer that.  But Bard's confusion is another symptom of why "JSON" people (@nishakm, @tsteenbe) insist that JSON is simple while JSON-LD is not.  I don't think Bard would have any trouble explaining XSD facets or JSON Schema types.

zvr:
"The result of Generative AI is highly plausible fiction -- which oftentimes happens to be factually correct."

goneall:
I think for RDF, my current opinion is we "restrict" our use to data types to only be pre-defined XSD schema datatypes and use SHACL to further restrict the property.  For `xsd:string` there is a number of [SHACL string based constraints](https://www.w3.org/TR/shacl/#core-components-string) the most interesting being [pattern](https://www.w3.org/TR/shacl/#PatternConstraintComponent).

It looks like implementing pattern restrictions is still a TODO in the spec-parser, so perhaps when we implement that we can switch from using `owl:subClassOf` to SHACL constraints.

goneall:
Back to the original PR - in RDF, the `SpdxId` will always have to be the URI of the object itself for the graph to work.

It seems a bit odd that we would have a property that has to be the same as the object URI.  We would want to restrict the object URI to be a real URI (e.g. no anonymous) - not sure if there is a way to do such a restriction.

@sbarnum any thoughts?

davaya:
I'm not sure the idea of Logical types and values is coming through.  For each Element you need to know:
1) its ID (the type of the ID is SpdxId)
2) its Type (one of the concrete types defined by the model)
3) the IDs of all elements it references, such as creators.  These references are also type SpdxID.

Serialization determines how the answers to those questions are represented in data.

In computing, matrices can be stored in memory in [row-major or column-major order](https://en.wikipedia.org/wiki/Row-_and_column-major_order) - it is the identical matrix regardless of how it is laid out (serialized) in memory. Similarly, Elements can be serialized into payloads in type-major order (as 2.3 does it, with elements indexed by type (Package, File, etc) first, then by id within a type) or id-major order (with elements indexed by id first, then the type of the element is designated individually for each id).  There have been discussions of which is better, but the point is that it makes no difference to the logical model. 
Serialization is invisible - two different serializations of the identical logical model must yield the identical set of elements - so arguing about type-major or id-major order has nothing to do with the model, it's purely a serialization discussion.

> It seems a bit odd that we would have a property that has to be the same as the object URI. We would want to restrict the object URI to be a real URI (e.g. no anonymous) - not sure if there is a way to do such a restriction.

A logical "property" is a question to be answered about an element value.  Regardless of how it is serialized, an element has to have an "object URI" (an id) and a type.  In RDF serialization there would not be two copies of the same id (or worse, two ids with different values).  Regardless of serialization, there is one answer to the id question - the spdxId "logical property" of that element, of type SpdxId.

goneall:
@davaya - My comment above is rather specific to RDF and the Ontology and Schema generated by the spec parser. In it's current form, we are duplicating the ID - I'm not sure that makes sense specifically for the RDF OWL and SHACL schemas.

We're currently using the following transformations:

Model markdown files -> model.ttl file (an RDF Turtle serialization of the OWL and SHACL for RDF) -> Java source files.

When I look at what comes out in the Java files, I'm noticing some oddities - one of which is duplicate object ID's.

davaya:
Sounds like a bug in the software that creates the ttl file or in the software that creates Java source files.  Hopefully the software that creates XSDs and JSON Schemas would not have that bug.

davaya:
I made some progress on the sandbox code that demonstrates datatype classes.  In the model files, a Format section signals that the class is a simple datatype.  A simple datatype has:
* no Properties properties
* Format properties that specify the [SHACL string constraints](https://www.w3.org/TR/shacl/#core-components-string) you mentioned, or a `schema` property that specifies the xsd constraints.

For example, the DateTime class has a pattern constraint under Format: 

## [DateTime](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/DateTime.md)
Model: [828388b](https://github.com/spdx/spdx-3-model/commit/828388b98c2374f1af6b760ab87fee0d4a11e3f4) 2023-07-05T10:09:48Z
```
class DateTime(String):
    pattern: ^\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\dZ$
```

String is the root string class with type xsd:string, AnyUri is a subtype of String with the xsd:anyURI schema, and SpdxId is a subtype of AnyUri that inherits the xsd:anyURI schema using the schema constraint under Format:

## [String]()
Model: [828388b](https://github.com/spdx/spdx-3-model/commit/828388b98c2374f1af6b760ab87fee0d4a11e3f4) 2023-07-05T10:09:48Z
```
class String():
    schema: xsd:string
```
## [AnyUri]()
Model: [828388b](https://github.com/spdx/spdx-3-model/commit/828388b98c2374f1af6b760ab87fee0d4a11e3f4) 2023-07-05T10:09:48Z
```
class AnyUri(String):
    schema: xsd:anyURI
```
## [SpdxId]()
Model: [828388b](https://github.com/spdx/spdx-3-model/commit/828388b98c2374f1af6b760ab87fee0d4a11e3f4) 2023-07-05T10:09:48Z
```
class SpdxId(AnyUri):
    schema: xsd:anyURI
```

Implementing simple datatype classes is simple in Python; I'm not a Java programmer but assume the details would be similar. [Test code for the DateTime class](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/56f383f882295e9d26be5c226aea944d68c5bfe6/model/testing/test_datatype_classes.py) produces the expected output:
```
   OK: 2023-04-13T11:15:00Z
Error: Fred: "Fred" does not match DateTime pattern
Error: 42: expected string or bytes-like object, got 'int'
Error: None: expected string or bytes-like object, got 'NoneType'
```

I hope you are able to find the transformation bugs that are causing duplicate Object IDs; the model markdown files are working fine with the simple datatype Core classes.

zvr:
As above, I am against introducing a type which does not add anything to what is already available.

davaya:
Again, the reason for defining an SpdxId datatype for xsd:anyURL is to distinguish URLs that are SpdxIds (e.g., declaredLicense, concludedLicense, originatedBy, suppliedBy) from URLs that are not SpdxIds (e.g., downloadLocation, packageUrl, homepage).

These are all xsd:anyURL but they are serialized differently.  If RDF does not serialize the originatedBy IRI differently from the packageURL IRI, then it can ignore the SpdxId type; defining it does not cause any problems if it is not used.  In JSON they are serialized differently - namespaceMap is not used to compress IRIs that are not SpdxIds.

goneall:
Since this is targeted for RC2 - @zvr @davaya - we should discuss within the next couple weeks and come to a conclusion

davaya:
> Since this is targeted for RC2 - @zvr @davaya - we should discuss within the next couple weeks and come to a conclusion

This is a long and interesting discussion, including Bard's AI opinions, but the questions to be answered are simple:
* **Does it help?** - does giving names to simple Datatypes like DateTime, SemVer and SpdxId make the model easier to understand and use?
*  **Does it hurt?** - do names make the model harder to understand and use? Are there other disadvantages?
* **Is it necessary?** - can the model be implemented without naming simple Datatypes?

We stipulate that names are not necessary - the model can be implemented without them.

The benefits are discussed above: SpdxId is used in many places, many more than DateTime.  And URIs that are not SpdxIds are also used in many places.  One example is **Package**, where every property is shown along with its type.  The `downloadLocation`, `packageUrl` and `homepage` properties are URIs, and so are `concludedLicense`, `declaredLicense`, `originatedBy` and `suppliedBy`.  It is not necessary to give them different names, but there are two benefits to doing so:
* it becomes obvious by inspection that some URIs identify SPDX elements and others do not
* it identifies for serialization software which URIs can be shortened using NamespaceMap prefixes and which cannot.

In my opinion the benefits of naming Datatypes in general and SpdxId in particular are obvious and substantial.

We have yet to hear any specific harms, costs, or disadvantages.


## [Package](https://github.com/spdx/spdx-3-model/blob/main/model/Software/Classes/Package.md)
Model: [8dff2a3](https://github.com/spdx/spdx-3-model/commit/8dff2a3243c9e00e1eb170fac749450a845ccdd6) 2023-07-28T22:15:03Z
```
class Package(/Software/SoftwareArtifact):
    packageVersion: String = None                      # * optional 
    downloadLocation: AnyUri = None                    # * optional 
    packageUrl: AnyUri = None                          # * optional 
    homePage: AnyUri = None                            # * optional 
    sourceInfo: String = None                          # * optional 
    contentIdentifier: AnyUri = None                   # * optional 
    primaryPurpose: SoftwarePurpose = None             # optional 
    additionalPurpose: SoftwarePurpose = None          # optional Set[1..*]
    concludedLicense: SpdxId = None                    # optional 
    declaredLicense: SpdxId = None                     # optional 
    copyrightText: String = None                       # * optional 
    attributionText: String = None                     # * optional 
    originatedBy: SpdxId = None                        # optional Set[1..*]
    suppliedBy: SpdxId = None                          # optional Set[1..*]
    builtTime: DateTime = None                         # optional 
    releaseTime: DateTime = None                       # optional 
    validUntilTime: DateTime = None                    # optional 
    standard: String = None                            # * optional Set[1..*]
    spdxId: SpdxId = None                              # 
    name: String = None                                # * optional 
    summary: String = None                             # * optional 
    description: String = None                         # * optional 
    comment: String = None                             # * optional 
    creationInfo: CreationInfo = None                  # 
    verifiedUsing: IntegrityMethod = None              # optional Set[1..*]
    externalReference: ExternalReference = None        # optional Set[1..*]
    externalIdentifier: ExternalIdentifier = None      # optional Set[1..*]
    extension: Extension = None                        # optional Set[1..*]
```

goneall:
> We have yet to hear any specific harms, costs, or disadvantages.

I'll register one disadvantage.  It is specific to RDF serializations.  We currently make the `spdxId` property the ID for the object.  In the JSON-LD serialization, this is done in the context file which makes the `spdxId` look like a property when it really isn't - it is actually `@id` which has special meaning and a predefined set of syntax and semantics.

If we create our own definitions of `spdxId` and we retain the current context file which makes `spdxId` the `@id`, then we are defining something that already has a definition.  Even if the definition is the same, it will lead to confusion IMO.

BTW - I don't have a strong opinion on whether this disadvantage outweighs the advantages - but it is something to consider.

davaya:
@zvr - is Gary's concern a blocker (defining an SpdxId class forces two definitions that prevent RDF serialization from working), or just a preference (@id is an IRI in RDF serialization, so there's no reason to define an SpdxId class of type IRI to support other serializations)?

goneall:
@zvr - Going through the old PR's for the 3.1 release - After reading through my old comments, I think adding a "special" class for SPDX ID does have advantages for readability - but we'd need to not include it in the SHACL output.

If you agree, can we enhance the spec parser to handle the special class?  We'd probably also want to update the documentation to indicate it is the `@id` in RDF JSON serialization.

bact:
26 May 2026 Tech call.
Current JSON-LD schema for SPDX 3 already addressed this by enforcing that `@id` should be `spdxId`, for a specific class (Element).

Agrees to close.

PR #410, opened 2023-07-07
Change nature to ObjectProperty for enums and object range

### LINKED ISSUE #409: Properties with a range of a class with properties should not be "Data Types"
Several properties with nature `DataProperty` have a range of a class which contains properties.

These properties should be object properties.

In RDF, a Data property can only have a range of a Datatype (at least in my Protege editor).

@zvr @armintaenzertng, @sbarnum if you agree, I can create a PR to fix.

goneall:
Fixes #409


maxhbr (approved):
LGTM

PR #411, opened 2023-07-09
Remove namespaces from the logical model

goneall:
Partial solution to #390 by removing the `namespaces` and `NamespaceMap` entirely from the model.  Note that this was suggested as a separate PR and comment in PR #403 which still kept the `NamespaceMap` for SPDX Documents.  Once this PR is agreed to / merged, we can create a separate PR to add a namespace map hint to SPDX documents.

See [this comment](https://github.com/spdx/spdx-3-model/pull/403#pullrequestreview-1520919577) for context.


maxhbr (approved):
LGTM

davaya:
LGTM

PR #414, opened 2023-07-11
add CreationInfo to all Elements in json-ld examples

armintaenzertng:
CreationInfo has been made mandatory again in #401, so the examples have to be updated.

This uses the de-duplication method of extracting the CreationInfo as a blank node.

### DIFF
--- a/serialization/json_ld/examples/converted_from_spdx_2.json
+++ b/serialization/json_ld/examples/converted_from_spdx_2.json
@@ -7,42 +7,47 @@
   ],
   "@graph": [
     {
+      "@id": "_:creationInfo1",
+      "type": "CreationInfo",
+      "specVersion": "3.0.0",
+      "created": "2010-01-29T18:30:22Z",
+      "createdBy": [
+        "spdx-example:SPDXRef-Actor-ExampleCodeInspect",
+        "spdx-example:SPDXRef-Actor-JaneDoe"
+      ],
+      "createdUsing": [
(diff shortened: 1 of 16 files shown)


goneall (approved):
Thanks @armintaenzertng  - I like the approach of using the blank node.

serialization/json_ld/examples/converted_from_spdx_2.json
```diff
@@ -6,43 +6,48 @@
...
+      ],
+      "dataLicense": "https://spdx.org/licenses/CC0-1.0",
+      "comment": "This is the SPDX-2.3 JSON example converted to SPDX-3.0. As there is currently no closure on how to treat licenses, they are omitted here for now."
+    },
     {
       "type": "Tool",
       "spdxId": "spdx-example:SPDXRef-Actor-LicenseFind-1.0",
-      "name": "LicenseFind-1.0"
+      "name": "LicenseFind-1.0",
+      "creationInfo": {"@id": "_:creationInfo1"}
```
maxhbr, line 33: Is there a reason, why you do not use this shorthand?
```suggestion
      "creationInfo":  "_:creationInfo1"
```

armintaenzertng: Thanks for the nudge, I realized that we have to change the context to have `"@type": "@id"` for `creationInfo` in order for this to work.

maxhbr (approved):
LGTM

PR #417, opened 2023-07-12
Change subclassOf: none to subclassOf: owl:Thing

goneall:
The current model generates an OWL document with an unknown superclass/subclass reference to "none" which is not valid or at least not expected.  owl:Thing is the expected subclassOf reference.  From the OWL spec: 'Every individual in the OWL world is a member of the class owl:Thing.'

### DIFF
--- a/model/Core/Classes/Element.md
+++ b/model/Core/Classes/Element.md
@@ -18,7 +18,7 @@ and inter-relatable content objects.
 ## Metadata
 
 - name: Element
-- SubclassOf: none
+- SubclassOf: owl:Thing
 - Instantiability: Abstract
 
 ## Properties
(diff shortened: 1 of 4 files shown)


zvr:
I'd actually prefer to leave it as-is in our repo, to not introduce more OWL-knowledge requirements.

Will you be happy if spec-parser performed this translation automatically?

goneall:
> I'd actually prefer to leave it as-is in our repo, to not introduce more OWL-knowledge requirements.
> 
> Will you be happy if spec-parser performed this translation automatically?

That works for me - I'll transfer this issue to the spec repo

goneall:
Closing this PR - opened https://github.com/spdx/spec-parser/issues/62 to request a change to the spec parser.

PR #420, opened 2023-07-13
set `maxCount: 1` for specVersion, created and dataLicense

armintaenzertng:
According to [this comment](https://github.com/spdx/spdx-3-model/issues/393#issuecomment-1602463064), the default for `maxCount` is `*`. This means that `specVersion`, `created` and `dataLicense` in the `CreationInfo` class do not have the right multiplicity, as there can't be more than one of these.


zvr (approved):

zvr:
Thanks for fixing this.

PR #424, opened 2023-07-18
move DateTime, SemVer and MediaType to Datatypes directory

armintaenzertng:
As discussed during the serialization meeting on July 14, this defines the three "string properties with formatting" as separate datatypes by moving them into a new folder "Datatypes" under Core.

### DIFF
--- a/model/Core/Datatypes/DateTime.md
+++ b/model/Core/Datatypes/DateTime.md
@@ -15,9 +15,6 @@ The specific format is one of the most commonly used ISO-8601 formats.
 ## Metadata
 
 - name: DateTime
-- SubclassOf: xsd:string
-
-## Properties
 
 ## Format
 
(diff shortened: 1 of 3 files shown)


zvr (approved):
We assume that all Datatypes are strings with patterns. I'm OK with that (until we find a counter-example).

goneall (approved):
LGTM

davaya:
LGTM.  I had already prepared a PR doing the same thing, adding an RFC-compliant pattern to MediaType, and adding SpdxId. https://github.com/davaya/spdx-3-model/tree/datatypes/model/Core/Datatypes

I'll submit the latter two after this is merged.

armintaenzertng:
Well, it will certainly lead to an invalid ontology for the time being, as the Datatypes folder is not yet parsed by the spec-parser. I opened https://github.com/spdx/spec-parser/issues/63.

PR #426, opened 2023-07-20
Update MediaType to RFC 6838, add SpdxId

davaya:


goneall (approved):
I didn't review the regex for media type in detail, but overall this looks OK.

One question for @zvr is how the spec-parser will handle the `SubclassOf: ` - Having this be a "subclass" of a datatype seems OK to me, but it isn't a standard RDF convention for string restrictions.

model/Core/Datatypes/DateTime.md
```diff
@@ -15,6 +15,7 @@ The specific format is one of the most commonly used ISO-8601 formats.
 ## Metadata
 
 - name: DateTime
+- SubclassOf: xsd:string
```
zvr, line 18: These are not needed, since we've said that all Datatypes are strings with pattern restrictions.

model/Core/Datatypes/MediaType.md
```diff
@@ -4,19 +4,20 @@ SPDX-License-Identifier: Community-Spec-1.0
...
-way of indicating the type of content of an Element.
+A MediaType is a string constrained to the RFC 6838 specification, and includes
+`type-name`, `subtype-name` and optional `parameters`.
+It provides a standardized way of indicating the type of content of an Element.
 A list of all possible media types is available at https://www.iana.org/assignments/media-types/media-types.xhtml.
 
 ## Metadata
 
 - name: MediaType
+- SubclassOf: xsd:string
```
zvr, line 19: Not needed, as above.

model/Core/Datatypes/MediaType.md
```diff
@@ -4,19 +4,20 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Format
 
-- pattern: ^[^\/]+\/[^\/]+$
-
+- pattern: ^([a-zA-Z0-9][-a-zA-Z0-9!#$&^_.+]{0,126})\/([a-zA-Z0-9][-a-zA-Z0-9!#$&^_.+]{0,126})(;.+)?$
```
zvr, line 23: I'd expect the type (left) part to be 

    application|audio|font|example|image|message|model|multipart|text|video

model/Core/Datatypes/SpdxId.md
```diff
@@ -0,0 +1,21 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SpdxId
```
zvr, line 3: No need for this type, which is the same as `xsd:anyURI`.

zvr (changes requested):

model/Core/Datatypes/MediaType.md
davaya: The RFC does not say that - the ABNF allows for additional registries and doesn't restrict their names.  But SPDX could choose to allow just the existing registries.

model/Core/Datatypes/DateTime.md
davaya: SpdxId is a string without pattern restrictions but with the xsd:anyURL restriction.  Other Datatypes could be strings with length restrictions.  Both the base class (to allow the RDF-defined xsd datatypes) and other datatype keywords (such as length) are needed.

model/Core/Datatypes/SpdxId.md
davaya: There is a need because SPDX has anyURI datatypes that are not SpdxIds (such as content identifier, package URL, home page, etc) that are not SpdxIds.  A homepage cannot be used as an Element id.

davaya: As in #407, the reason for defining an SpdxId datatype for xsd:anyURL is to distinguish IRIs that are SpdxIds (e.g., declaredLicense, concludedLicense, originatedBy, suppliedBy) from IRIs that are not SpdxIds (e.g., downloadLocation, packageUrl, homepage). These are all of type xsd:anyURL but they are serialized differently  - namespaceMap is not used to compress IRIs that are not SpdxIds.

Look at [Package](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/main/model/generated_20230719T155944Z/Software/Package.md) for a concrete example, and compare downloadLocation with concludedLicense. 

Naming the SpdxId datatype, just like naming the DateTime datatype, obviously adds to what would be in the model without them.  No rationale has been presented for how naming simple Datatypes is harmful, and the names are helpful.  

Just do it.

goneall: @davaya - Can we remove the SPDX ID from this PR?  You already have this documented in a separate issue or pull request and there doesn't seem to be any question on the MediaType.  I'd like to get the MediaType merged in.  We can continue the debate on SPDX ID in a separate thread.

davaya: Will do. Do you prefer just the 10 registries for the first part, or the ABNF string?

model/Core/Datatypes/MediaType.md
zvr: @goneall we already mention that the list of accepted types are the ones in the IANA list. Why allow arbitrary strings?

model/Core/Datatypes/DateTime.md
zvr: I repeat, all our Datatypes are strings with pattern restrictions. It makes no sense to plan for something we do not have any use for.

davaya: It is not a plan. We already have PositiveIntegerRange and could well need NonNegativeIntegerRange if Snippet byte ranges are to be zero-based.  We already have Package downloadlocation and Package concludedLicense, which are semantically different types.

You see no use for these, "we" (other members of the tech team) do see a use for them.  So again, is there any harm?

model/Core/Datatypes/SpdxId.md
davaya: Include "registries vs. type names" decision on Aug 1 tech team meeting.

goneall:
Since @zvr only has an issue with the `subClass` usage, I'm going to go ahead and merge this.  We can always remove the `subClass` of String if it really is an issue.

kestewart (approved):
This looks good to me.

Issue #415, opened 2023-07-11
Clarify Bom & SpdxDocument in 3.0 Model


jeff-schutt:
In the current version of the [3.0 model](https://github.com/spdx/spdx-3-model/blob/main/model.png), BOM and SpdxDocument both inherit from Bundle. It's unclear when we would use SpdxDocument and when we would use Bom (or Sbom which inherits from Bom). Additional guidance is needed before various profiles can consistently use these classes in JSON examples being created.

Discussed in the 7/11 tech call:
* serialization team agrees it needs to be resolved and had been discussing without resolution
* tech team agreed that it should be discussed in a future tech meeting as the model will likely need to change based on the decision that is reached

davaya:
We agreed to remove Payload (and NamespaceMap) from the logical model, which indicates we're part way toward accepting the difference between logical values and serialized data. This is another symptom of conflating logical and physical data.

* A physical payload carries Element values.
* A Collection/Bundle/Bom/Sbom is a logical Element that groups logical Elements
* SpdxDocument is a logical Element that groups physical values (the ones carried in a specific payload).

So, a Bundle/Bom/Sbom would list a bunch of elements that are members of the logical structure - for example an Sbom with 5 Elements - 1 package, 2 files, 1 relationship, 1 person.

A payload could carry those six element values (Sbom plus its five members).  But a payload could also carry 15 element values - three Sboms and their various contents, or three element values (half of the first Sbom and its contents because the other three are external elements).

It is a mistake for SpdxDocument to inherit from Bundle - it has both a list of elements in the payload and artifact information related to the physical payload; in my opinion Artifact is the more important parent, and element [1..*] is just a property.  And SpdxDocument doesn't need rootElement (which is required by ElementCollection) - that's a logical concept and a payload is just a pile of element values with no structure; the structure only exists after parsing payloads into logical elements.

goneall:
We decided to discuss this in a future tech call.

Note the related issues referenced above.

goneall:
@jeff-schutt I believe this is fixed with the PR's under review, especially PR #490

PR #431, opened 2023-07-23
Replace imports/ExternalMap with SpdxDocument

davaya:
Issue #415 calls for clarifying the roles of Bom and SpdxDocument:
* a Bundle/Bom/Sbom lists elements that are members of a logical collection without regard to where they are serialized.
* an SpdxDocument element lists elements serialized in a payload, used to import elements from another payload.

 This PR addresses logical/physical separation and the import use case by:
* deleting the ExternalMap type and the ElementCollection imports property that references it because "external" isn't a concept that applies to the logical element graph.
* adding locationHint and verifiedUsing properties to SpdxDocument to locate and verify the referenced payload.

The [serialization playground](https://github.com/spdx/spdx-3-serialization-prototype-playground/tree/main/implementations/2) has an example of one payload referencing elements carried in another payload using SpdxDocument.


model/Core/Classes/SpdxDocument.md
```diff
@@ -23,9 +23,15 @@ Commonly used when representing a unit of transfer of SPDX Elements.
   - type: xsd:string
   - minCount: 1
   - maxCount: 1
-
-## External properties restrictions
-
-- /Core/Element/name
+- element
```
maxhbr, line 26: isn't a `Element` with the list of elements a Bundle or Collection? Why not inherit from one of those?

davaya: In general because SpdxDocument has nothing to do with the logical model, or as Sean put it yesterday, with the "SPDX graph". When you define a Bundle element you are saying "the graph is saying something meaningful and worth remembering about this group of elements", whereas SpdxDocument is saying "if you have a payload or find it based on locationHint, you can verify the elements in it".  The fact that some elements are in the same payload is not worth memorializing in the graph, particularly since the same elements can appear in different payloads in different combinations."   ElementCollection is a persistent grouping, the `SpdxDocument Element` is purely transactional, as opposed to the `SpdxDocument` (payload) itself, which is a concrete persistent object itself that has meaning and is used at the physical layer, outside the graph, regardless of whether an `SpdxDocument Element` has ever been created to reference it.

In particular, because even ElementCollection has rootElement, which doesn't have anything to do with serialized data payloads.  The payload itself has to contain everything needed to understand and use it; anything in the `SpdxDocument Element` beyond what is needed to find and verify serialized data is just another dependency / coupling between logical and physical layers that should not exist.

goneall (commented):
Finally got a chance to not only review, but think through the different scenarios.  I believe these changes make sense and better represents the intent of SPDX Document and better separates the logical and physical model.

That being said, since we came up with the current model after lots (and lots) of discussion on the tech call, I'm reluctant to merge before discussing at least briefly.

I'm particularly interested in @iamwillbar and @sbarnum views on this.

goneall:
To avoid confusion, I'm going to close this since it conflicts with our recent decision.

@davaya - we can re-open if our test of the external map proves that it doesn't work efficiently.

PR #441, opened 2023-07-28
add documentation on how to parse JSON-LD as pure JSON

armintaenzertng:
This an attempt at documenting how to parse JSON-LD without any RDF knowledge.
I'm not quite sure where to put this, though.

### DIFF
--- a/serialization/json.md
+++ b/serialization/json.md
@@ -1,26 +1,30 @@
 # JSON Serialization
 
-Ideally JSON will be very similar (or even the same) to JSON-LD.
+## Parsing JSON-LD as JSON
+This is a description of how to deserialize JSON-LD as a pure JSON format without any knowledge of RDF.
+On top-level, JSON-LD has two keys, "@context" and "@graph".
 
-The JSON serialization will consist of two things: A context file and an array of element objects (the "payload").
+### parsing "@context"
+The context is a list of a string and an object. You can ignore the string.
+The object consists of key-value pairs that allow the shortening of IDs and which we will call "namespace map" in the following.
+For deserialization purposes, follow this process:
+- For every string that is an ID (that includes values of the keys "spdxId" and "@id", 
+  as well as all strings where you would expect objects according to the SPDX-3 model),
+  split that string at the first colon into "prefix:suffix".
+- If the suffix does not start with "//" and the prefix is a key in the namespace map,
+  replace "prefix:" with the value found under that key in the namespace map.
+- Else do nothing to that string.
+After you are done applying this process to all IDs, you can ignore the "@context".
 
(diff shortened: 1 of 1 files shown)


serialization/json.md
```diff
@@ -1,26 +1,30 @@
...
+This is a description of how to deserialize JSON-LD as a pure JSON format without any knowledge of RDF.
+On top-level, JSON-LD has two keys, "@context" and "@graph".
 
-The JSON serialization will consist of two things: A context file and an array of element objects (the "payload").
+### parsing "@context"
+The context is a list of a string and an object. You can ignore the string.
+The object consists of key-value pairs that allow the shortening of IDs and which we will call "namespace map" in the following.
+For deserialization purposes, follow this process:
+- For every string that is an ID (that includes values of the keys "spdxId" and "@id", 
+  as well as all strings where you would expect objects according to the SPDX-3 model),
```
goneall, line 12: > as well as all strings where you would expect objects according to the SPDX-3 model

Any hints on how to recognize these?  This may be one of the more confusing points.

serialization/json.md
```diff
@@ -1,26 +1,30 @@
...
-The rest of the serialized file will consist of an array of objects under the "@graph" key, also at top-level of the JSON file.
-Each of these objects must have a property "@type", stating its class name which must be an instantiable subclass of Element.
-It also needs a property "@id", which serves to capture the SPDX ID of the Element.
-All other SPDX properties of the Element are stated via key-value pairs, where the key is the name of the property (e.g. "createdBy").
-The type of the value depends on the type of the property in the SPDX model and can be a string, number, array or object.
-Note, though, that inlining of objects is only allowed for "complex data type" classes (CreationInfo, ExternalReference, etc.).
-If the object were an Element, a string containing its ID would be written instead, thereby referencing another object from the same or even another payload.
+One thing to note is that not all objects in that list have to be subclasses of Element.
+As only Elements have an spdxId, there is no "spdxId" key in these cases but an "@id" key.
+However, the value of "@id" serves the same function of identifying and referencing that object from within other objects.
```
goneall, line 27: Not really a comment on this advice, but it does make me think if it would be more straightfoward to just use `@id` for the SPDX ID's as well - that would simplify the translation - at the expense of making it less human readable.

goneall (approved):
Just a couple thoughts - but I'm fine with this as it is.  If approved by a second person in the tech team, I would suggest merging this and we can use it as a base for further refinement as we go through the serialization discussions.

serialization/json.md
maxhbr: interesting thought. It is still in there from the initial effort. I see good arguments for both directions

goneall:
@maxhbr If you're good with this as and approve it, I'll merge

maxhbr (approved):
LGTM

PR #444, opened 2023-07-28
Add customIdToUri Map to LicenseExpression

### LINKED ISSUE #380: License Expression References to Custom Licenses
How do we represent `LicenseRef-` custom licenses in the string license expressions?

In SPDX 2.X, the interpretation of the custom licenses is very dependent on the SPDX document containing the reference.   We key off the specific string `LicenseRef-` to determine it is a custom license and know we can find the definition within the document itself.  If it is defined outside the document we use a namespace technique of `ExternalDocumentIdX:LicenseRef-xxx`.

Since the license expressions an be created without an enclosing document, we need a way of identifying the local licenses and finding the license definitions.  This could be done by requiring the fully expanded URI's - but that would make for some rather unwieldy expressions.

goneall:
Fixes #380

Implements suggestion made in [this comment](https://github.com/spdx/spdx-3-model/issues/380#issuecomment-1620559842).

See [this comment](https://github.com/spdx/spdx-3-model/issues/380#issuecomment-1636183121) for an example serialization in JSON.

### DIFF
--- a/model/SimpleLicensing/Classes/LicenseExpression.md
+++ b/model/SimpleLicensing/Classes/LicenseExpression.md
@@ -23,3 +23,6 @@ SPDX License Expressions provide a way for one to construct expressions that mor
   - type: xsd:string
   - minCount: 1
   - maxCount: 1
+- customIdToUri
+  - type: /Core/DictionaryEntry
+  - minCount: 0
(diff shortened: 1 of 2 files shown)


zvr (approved):

maxhbr (approved):
I think this technically accomplishes its goal and improves the state and therefore can be merged

But I am afraid that this adds even more complexity to an already hard to grasp concept.

And the current model does not avoid duplicate keys and does not handle them if they would appear.

PR #451, opened 2023-07-30
Add licenseXml property to License and LicenseAddition

### LINKED ISSUE #359: Add a licenseXML field to the License class
Based on https://github.com/spdx/spdx-spec/issues/853, we should add the license XML to the listed license metadata.

We could also remove the license template field, but I would prefer to leave that in as a deprecated field for at last one release to allow migration to the new format.

goneall:
Resolves #359

### DIFF
--- a/model/ExpandedLicensing/Classes/License.md
+++ b/model/ExpandedLicensing/Classes/License.md
@@ -47,3 +47,7 @@ A License represents a license text, whether listed on the SPDX License List
   - type: xsd:string
   - minCount: 0
   - maxCount: 1
+- licenseXml
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
--- a/model/ExpandedLicensing/Classes/LicenseAddition.md
+++ b/model/ExpandedLicensing/Classes/LicenseAddition.md
@@ -41,3 +41,7 @@ or otherwise) which is defined by an SPDX data creator (CustomLicenseAddition).
   - type: xsd:string
   - minCount: 0
   - maxCount: 1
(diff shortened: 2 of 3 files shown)


kestewart (approved):
Looks good to me.

goneall:
@swinslow - If you could do a quick review, I can merge if it looks OK

swinslow:
@goneall My only question is whether the property on LicenseAddition should be called `additionXml` rather than `licenseXml`? To help clarify that it is not identical to a "license" XML file.

I'm not going to lose sleep over this one if you feel strongly that they should be the same  :)

goneall:
> My only question is whether the property on LicenseAddition should be called additionXml rather than licenseXml? To help clarify that it is not identical to a "license" XML file.

@swinslow Your suggestion makes sense to me, however as I was starting to make the changes and reading the text of the licenseXml.md file, I notice it points to the [license XML fields](https://github.com/spdx/license-list-XML/blob/main/DOCS/xml-fields.md) which documents both the "license" and "exception" fields.

Since it is pointing to the same XML format for both license and addition, I'm wondering if they should keep the same term.  It does seem a bit odd to have something called licenseXml in the addition.  Let me know what you think. Should we leave it as is, change the name of the field to something more appropriate or change the documentation for license fields to define two different format?

swinslow:
Yeah.... we refer to it generally as the "license list XML" -- I think we can keep it as `licenseXml` unless / until someone complains. I think it's fine. Thanks @goneall!

swinslow (approved):

Issue #319, opened 2023-05-10
Conversion 2->3: LicenseComment


meretp:
Where should we put the `license_comment` property from `package`, `file` and `snippet` when converting to SPDX 3.0? 

According to the 2.3 [specification](https://spdx.github.io/spdx-spec/v2.3/package-information/#716-comments-on-license-field) `license_comment` should be used to "to record any relevant background information or analysis that went in to arriving at the Concluded License for a package [..] It is also preferable to include an explanation here when the Concluded License is NOASSERTION.".

This raises two issues:
- The new `NoAssertionLicense` class doesn't have a property for a comment.
- Only subclasses of `License` have a property `licenseComment`, so we can only put the possibly general comment about a complex license expression in `concludedLicense` to an instance of `ListedLicense` or `CustomLicense`. Should then every individual license get this comment? E.g. in the 2.3 document we might have a package with a concluded license: `MIT AND LicenseRef-2` and the license comment says something like "the author concluded MIT because [..]". As the `ConjunctiveLicenseSet` doesn't have a `licenseComment` property, we would need to add this comment to the `ListedLicense("MIT")` and the `CustomLicense("LicenseRef-2")`. Is this the correct/ intended way to translate this?

goneall:
In SPDX 2.3, the licenseComment was a property of the SPDX element.

For SPDX 3.0, this seems to be more appropriate for `SoftwareArtifact`.

I'll create a PR with a proposed solution.

goneall:
As suggested by @maxhbr - we could use the Relationship comment for this purpose.

If no one disagrees, I'll close this issue and the related PR.

swinslow:
@goneall @maxhbr I agree, if relationships are being used for concludedLicense / declaredLicense then the relationship comment seems like the right thing to use here.

goneall:
Thanks @swinslow for the review and comments.

With this - I'll close the PR and close this issue as well.

PR #454, opened 2023-07-30
Add licenseComment to SoftwareArtifact

goneall:
Resolves issue #319


kestewart (approved):

model/Software/Classes/SoftwareArtifact.md
```diff
@@ -46,4 +46,7 @@ such as a package, a file, or a snippet.
   - type: xsd:string
   - minCount: 0
   - maxCount: 1
-
+- licenseComment
```
maxhbr, line 49: shouldn't that be comment to the relationship which encodes the conclusion?

maxhbr (changes requested):
Just a question

Issue #204, opened 2023-04-16
Licensing: Where should contextual meanings of NONE / NOASSERTION be stated?


swinslow:
Currently, in the [licensing-profile branch](https://github.com/spdx/spdx-3-model/tree/licensing-profile/model), the Classes [NoneLicense](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Licensing/Classes/NoneLicense.md) and [NoAssertionLicense](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Licensing/Classes/NoAssertionLicense.md) contain brief descriptions of their meanings.

However, because these values have slightly different meanings in the Declared License vs. Concluded License context, the specific meanings of "NONE" and "NOASSERTION" for these uses are spelled out in the [declaredLicense](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Licensing/Properties/declaredLicense.md) and [concludedLicense](https://github.com/spdx/spdx-3-model/blob/licensing-profile/model/Licensing/Properties/concludedLicense.md) Properties files.

I'm fine with either structure, but wanted to confirm with the broader SPDX 3.0 team about which location you feel is preferable.

swinslow:
Just to note, a similar question applies to the NoneText and NoAssertionText classes.

Also, I'm guessing the resolution of this issue to some extent depends upon how we're handling #71 / #76.

goneall:
In the tech call discussion on 17 April 2023, we discussed having an Individual Value for NONE and NOASSERTION rather than different classes / types.  We still need to finish a few details on how this would be implemented consistently across the profiles before making any updates.

zvr:
Are we talking about different individuals per class/type (like _NoneLicense_ and _NoAssertionLicense_) or single individuals _None_ and _NoAssertion_ that will be used in different use cases?

goneall:
> Are we talking about different individuals per class/type (like NoneLicense and NoAssertionLicense) or single individuals None and NoAssertion that will be used in different use cases?

That wasn't specified in the call.  My proposal would be to have it per class/type which is different from the SPDX 2.3 model.  Having it per type would simplify the range values for the properties and better match the programming model used by the various tools.

For the text type, I would propose we just use a single NONE and NOASSERTION individual of type xsd:string for RDF.  For non-RDF serializations, we can just describe the meaning of NONE and NOASSERTION text values in the property descriptions as we do today.

swinslow:
[...]

Issue #76, opened 2023-02-06
Handling NONE and NOASSERTION values in a consistent manner


goneall:
In the SPDX 2.X spec, there are several fields which accept a value of `NONE` and `NOASSERTION` which have specific semantic meaning along with either a text or a class. 

This is defined in RDF as a superclass restriction - you can find an example for [SPDX Item](https://spdx.org/rdf/terms/#d4e3785) in the `License Concluded` property.  This may not be the cleanest approach since we're mixing individual values with other types (e.g. `xsd:string`) and the reasoning is rather clunky IMHO.

@swinslow is proposing having an abstract class for each of the properties where the values can include the `NONE` and `NOASSERTION` values which may yield a slightly different RDF model.

There may also be other alternatives.

Whichever approach we take, we should probably be consistent.

Related to issue #71 

@sbarnum @davaya @zvr - Interested in any thoughts you have on this issue.

swinslow:
Thanks @goneall! I'm closing #65 in favor of #77 so that the initial Licensing profile draft can be merged into a separate branch -- just flagging that here so that it doesn't get lost  :)

I don't have a strong preference either way on the manner in which NONE / NOASSERTION is handled. I had structured it as shown in #65 / #77 mostly out of trying to move away from assuming string values for Licensing fields as you noted, but happy to adjust as appropriate.

maxhbr:
this is a duplicate of https://github.com/spdx/spdx-3-model/issues/71?

goneall:
Discussed on the SPDX Tech call on 18 April 2023:
- For properties with an enumeration, we can just have a value NONE and NOASSERTION be one of the enumeration values
- For RDF properties with a typed value, having a concrete individual representing the NONE and NOASSERTION is the preferred approach.  If we were to take this route, we would need to extend the markdown file format to be able to express individual values in the range for a property.  Perhaps having a Union of a type and an individual value?
- An alternative solution is to "wrap" the typed value with another class that offers the typed value, NONE and NOASSERTION as 3 different possible results (similar to Java's Optional class).  This was considered more complex for the RDF Ontology.

goneall:
@zvr - Any thought on how we can express an Individual value for the range of a property?

goneall:
If we use an Individual approach as suggested on the tech call, should we create separate individuals for each class/type (e.g. NoAssertionLicense) or just use a single individual for all uses (e.g. NoAsserttion).  In SPDX 2.X, we use just the single individual in all cases.

[...]

Issue #299, opened 2023-05-06
Proposal: Add definitions for Individuals


goneall:
To implement NONE and NOASSERTION values for complex types, we need a way to define Individual values.

An individual value is an instance of a Class with specified properties.

I would propose adding these into an Individuals folder.

The file format would be similar to the Vocabularies file format.

Below are proposed sections:

- Summary
- Description
- Metadata
- name within Metadata
- id within Metadata (the ID would be the namespace) followed by a unique ID
- type within Metadata
- propertyValues - any properties with pre-defined values.  Currently, I only expect the ID's to be xsd:string types although we may need to extend these in the future.

goneall:
@zvr - let me know any thoughts - we can implement this after RC

zvr:
Sure.

Not even sure "name" is needed, to be honest.

Here's an example:
```
SPDX-License-Identifier: Community-Spec-1.0

# NoneCopyrightText

## Summary

Individual representing an assertion that no copyright text is present.

## Description

A NoneCopyrightText is the value that is used to indicate
that a software artifact does not contain any copyright information.

## Metadata

- id: NoneCopyrightText
- type: CopyrightText

## Property Values

- text: "NONE"
```

goneall:
> Not even sure "name" is needed, to be honest.

Agree - in fact for our use cases, we probably only need ID and Type.

goneall:
Fixed in PR #456

PR #448, opened 2023-07-29
Change declared and concluded licenses to relationships

goneall:
Implements decision made on whether licenses should be a property or relationship on the 2023-05-02 tech call

Note that this introduces issues as documented in #254

This fixes the parsing errors #443

Also fixes #Signed-off-by: Gary O'Neall <gary@sourceauditor.com>

PR #456, opened 2023-07-31
Update Individual values for NONE and NOASSERTION licenses

goneall:
This PR implements NONE and NOASSERTION licenses as individuals.

It follows the recommended format documented in issue #299 

Note: It may require changes to the spec-parser before merging.


goneall:
Resolves #204

goneall:
Resolves #76

model/ExpandedLicensing/Vocabularies/NoAssertionLicense.md
```diff
@@ -0,0 +1,24 @@
...
+# NoAssertionLicense
+
+## Summary
+An Individual Value for License when no assertion can be made about it's actual value.
+
+
+## Description
+
+NoAssertion should be used if the SPDX creator has attempted to but cannot reach a reasonable objective determination;
+the SPDX creator has made no attempt to determine this field; or
```
maxhbr, line 12: as discussed in https://github.com/spdx/spdx-3-model/issues/289, compared to SPDX2 we have also `null` (or not including any licensing information) as an value and it should be clearly distinct from `NoAssertion`. I think I understood that `NoAssertion` should only be valid if and only if "the SPDX creator has intentionally provided no information".

I think we need to align this with https://github.com/spdx/spdx-3-model/issues/289#issuecomment-1589450596

maxhbr (changes requested):

model/ExpandedLicensing/Vocabularies/NoAssertionLicense.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Metadata
+
+- name: NoAssertionLicense
+- type: License
+
+## Property Values
+
+- name: "NOASSERTION"
+- /SimpleLicensing/licenseText: "Not Applicable: Text for a NOASSERTION License"
```
maxhbr, line 23: nit: I think it is not a `NOASSERTION License` as it is no license:
```suggestion
- /SimpleLicensing/licenseText: "Not Applicable, as there is no assertion about license information"
```

model/ExpandedLicensing/Vocabularies/NoAssertionLicense.md
```diff
@@ -0,0 +1,24 @@
...
+## Description
+
+NoAssertion should be used if the SPDX creator has attempted to but cannot reach a reasonable objective determination;
+the SPDX creator has made no attempt to determine this field; or
+the SPDX creator has intentionally provided no information (no meaning should be implied by doing so).
+
+## Metadata
+
+- name: NoAssertionLicense
+- type: License
```
maxhbr, line 18: I think this should extend `AnyLicenseInfo` or maybe `ExtendableLicense`. It then would also not have a licenseText, that would need to be overwritten with weird text.

model/ExpandedLicensing/Vocabularies/NoAssertionLicense.md
goneall: The comment implies changing the definitions for the properties, not the NoneLicense and NoAssertionLicense.

@maxhbr if you agree, I can add a suggested text change to the properties along with this PR.

@kestewart - see the comments in issue #289 for more context

model/ExpandedLicensing/Vocabularies/NoAssertionLicense.md
goneall: Agree - I'll include a fix in the next update

model/ExpandedLicensing/Vocabularies/NoAssertionLicense.md
maxhbr: > @maxhbr if you agree, I can add a suggested text change to the properties along with this PR.

yes, that would match what was discussed (although I am still not happy with it).


Since the 2.3 Noassertion now maps to two values (not present and noassertion in 3.0), we need to do a choice regarding migration. Maybe that should also be added to the description as a comment.

goneall: @maxhbr @kestewart I added proposed text to concludedLicense.md and declaredLicense.md - please review

zvr:
@goneall why are the files in the `Vocabularies` directory and not in the `Individuals` as they already are?

goneall:
> @goneall why are the files in the Vocabularies directory and not in the Individuals as they already are?

@zvr - Individuals?  Is this a separate directory?  If you prefer me to create a new directory called Individuals, I can move them there.  I'll move them to wherever the spec-parser expects to see them.  Just confirm.

BTW - I think of the enumerations also as individuals, hence the choice of putting them in the Vocabularies directory.

zvr:
@goneall, I mean, it's already there in the main branch: https://github.com/spdx/spdx-3-model/tree/main/model/ExpandedLicensing/Individuals

Yes, Vocabularies also generate individuals, but that's one for each list item, while these are more fleshed out... Let's not mix our parsing, please.

goneall:
> I mean, it's already there in the main branch

I missed that - looks like @armintaenzertng already had the individuals added

I'll more these changes over to the individuals folder

goneall:
I moved the Individual definitions over to the Individuals folder.

Here's a summary of the differences from @armintaenzertng original:
- Updated descriptions - I copied these from some of @swinslow original descriptions and from the SPDX 2.3 spec with appropriate modifications
- Change metadata `id` to `name` - `name` is consistently used through other metadata (e.g. for Classes, Properties, and Vocabularies)
- Replaced the property `text` with property `name` - `AnyLicenseInfo` does not have a `text` property whereas `name` is inherited from `Element`

goneall:
@maxhbr @zvr @armintaenzertng @swinslow - I just created a profile `Licensing` which just contains the descriptions for concluded and declared licenses in this draft PR.

Once we figure out how to define restrictions on relationships, we can add them in this profile.

Please let me know if you disagree with this approach.

model/Licensing/Licensing.md
```diff
@@ -0,0 +1,97 @@
...
+
+If a declaredLicense has a NOASSERTION value (NoAssertionLicense), this
+indicates that one of the following applies:
+* the SPDX data creator has attempted to but cannot reach a reasonable
+  objective determination;
+* the SPDX data creator has made no attempt to determine this field; or
+* the SPDX data creator has intentionally provided no information (no meaning
+  should be implied by doing so).
+  
+If a declaredLicense is not present or a null value, no conclusion can be drawn.
```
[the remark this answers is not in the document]
goneall: @maxhbr - I would leave it as null - I don't think we're changing the semantics?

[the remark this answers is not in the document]
maxhbr: for 2.3 NOASSERTION is equal to `null`, since there was no `null` in the model. We are introducing it with 3.0. So we should treat an explicit NOASSERTION equal to the shorthand `null`.

model/ExpandedLicensing/Individuals/NoAssertionLicense.md
```diff
@@ -3,19 +3,21 @@ SPDX-License-Identifier: Community-Spec-1.0
 # NoAssertionLicense
 
 ## Summary
+An Individual Value for License when no assertion can be made about it's actual value.
```
swinslow, line 6: minor nit: "it's" => "its"

model/ExpandedLicensing/Individuals/NoAssertionLicense.md
```diff
@@ -3,19 +3,21 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
-Individual representing an absence of an assertion about license information.
 
 ## Description
 
-A NoAssertionLicense is the value that is used to indicate that the data creator
-is making no assertion about the license information for the corresponding software artifact.
+NoAssertion should be used if the SPDX creator has attempted to but cannot reach a reasonable objective determination;
```
swinslow, line 11: For the first word, "NoAssertion" or "NoAssertionLicense" or "NOASSERTION"?

I'd think "NoAssertionLicense" if it's referring to the Individual's title, or "NOASSERTION" if it's referring to the concrete string value.

model/ExpandedLicensing/Individuals/NoneLicense.md
```diff
@@ -3,19 +3,18 @@ SPDX-License-Identifier: Community-Spec-1.0
 # NoneLicense
 
 ## Summary
+An Individual Value for License no license is present.
```
swinslow, line 6: I think there's a missing bit: e.g. "for License **where the SPDX data creator determines that** no license is present"

model/ExpandedLicensing/Individuals/NoneLicense.md
```diff
@@ -3,19 +3,18 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
-Individual representing an assertion that no license information is present, as applicable.
 
 ## Description
 
-A NoneLicense is the value that is used to indicate
-the absence of license information from a software artifact.
+The SPDX document creator concludes there is no license available for this package
```
swinslow, line 11: 1. SPDX document creator, or SPDX data creator? I'm assuming "data" creator since I gather for 3.0 this might not be in the context of an SPDX document.
2. I would probably change "package" to "software artifact" since this could also be used for Files or Snippets

model/Licensing/Licensing.md
```diff
@@ -0,0 +1,97 @@
...
+
+# Licensing
+
+## Summary
+
+The Licensing Profile defines a minimum set of license information required to comply with typical license compliance use cases.
+
+## Description
+
+The Licensing profile only contains additional restrictions.
```
swinslow, line 11: What is this meant to be saying?

I probably wouldn't use this phrasing -- "additional restrictions" in the open source licensing context is often used to mean e.g. provisions in license texts that impose additional requirements or prohibitions.

model/Licensing/Licensing.md
```diff
@@ -0,0 +1,97 @@
...
+Classes and Property restrictions are defined in the SimpleLicensingProfile
+(Classes and Properties associated with string license expressions) and in the ExpandedLicensingProfile (Classes and Properties used for a
+fully parsed syntax tree of license expressions).
+
+There are 2 relationships which are required for any Software Artifact - declaredLicense and concludedLicense.
```
swinslow, line 16: This isn't what we had decided would be required for the Licensing profile as a requirement during the 2020/2021 discussions. Only concludedLicense was required.

That said, I'm not clear what it means for it to be "required" in any case where concludedLicense is a relationship now, and where I don't think documents are required?

I'm not sure how other profiles are handling requirements like this, where relationships might exist anywhere in SPDX data not tied to a particular document. Are other profiles scoping it to e.g. "in order to conform with a profile, there must be an SPDX Document which contains a relationship for each of these items"?

model/Licensing/Licensing.md
```diff
@@ -0,0 +1,97 @@
...
+
+A declaredLicense is the license identified in text in the software package,
```
swinslow, line 18: I think this is missing a pretty key piece from the "summary" part of the prior text, which isn't there now but is visible in the PR that removed it, at https://github.com/spdx/spdx-3-model/pull/448/files#diff-7c90a0506e9d8122276df9c2e1d92a0115bc693c4f2cc53c95fab28c7be943f1L7-L8:

A declaredLicense "[i]dentifies the license information actually found in the software Package, File or Snippet, for example as detected by use of automated tooling."

The "as detected by use of automated tooling" is pretty important to make it clear that this is where information about detected licenses should be expressed.

swinslow (commented):
@goneall Thanks! I've included a few comments inline in this PR.

model/Licensing/Licensing.md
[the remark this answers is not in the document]
goneall: > for 2.3 NOASSERTION is equal to null, since there was no null in the model. 

In 2.3 we made the license properties optional, so I believe they can be null.  That is how I'm interpreting a missing value in the JSON file.

If we have a JSON file where a license field is not present in 2.3, we should likewise have it not present in 3.0.  I don't see a need to convert it to a NOASSERTION.

Is the issue with the text `or a null value`?  If so, we can remove it.

model/Licensing/Licensing.md
swinslow: And sorry, just to clarify -- only concludedLicense was required _for the SPDX data to conform to the Licensing profile_.

model/ExpandedLicensing/Individuals/NoAssertionLicense.md
goneall: Updated per suggestion

model/ExpandedLicensing/Individuals/NoAssertionLicense.md
goneall: Updated to use NoAssertionLicense

model/ExpandedLicensing/Individuals/NoneLicense.md
goneall: Updated - along the same lines as the suggestion

model/ExpandedLicensing/Individuals/NoneLicense.md
goneall: Updated per suggestions

model/Licensing/Licensing.md
goneall: I clarified this to be related to the model

model/Licensing/Licensing.md
goneall: On your first point - I fixed it to only state the requirement was on the concluded license.

On you second point, I opened issue #463 to track.  This is targeted to be resolved after RC2 whereas these descriptions should be updated prior to release of RC2.

model/Licensing/Licensing.md
goneall: Updated per suggestion.

goneall:
Thanks @swinslow for the reviews.

I think I updated everything per review comments and then some.  It would be worth another look since there were a lot of changes and I may have missed something.

Also - based on one of your comments, I wonder if I copy/pasted from the wrong source on the Licensing.md file.  Let me know if you see anything else that needs to be fixed.

model/Licensing/Licensing.md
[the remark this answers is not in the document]
maxhbr: Sorry for going in circles here. In 2.3 the properties were optional (cardinality 0 to 1) but the following text was added to the description:

> If the Declared License field is not present for a package, it implies an equivalent meaning to NOASSERTION. 

this effectively collapses the the "not present" and the "NOASSERTION" case again, for 2.3. In my recollection this was purely added to allow for a shorter serialization and the Intend was not to make the model more expressive.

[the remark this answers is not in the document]
goneall: Got it.

I don't know if we intentionally changed the semantics of declared license.  I still think if there is no value in 2.3, we should create a `NoAssertionLicense` relationship in 3.0.

@swinslow - any thoughts?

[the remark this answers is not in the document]
maxhbr: agree, mapping  "not present" and the "NOASSERTION" both to the `NoAssertionLicense` Relationship is consistent for me.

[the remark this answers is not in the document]
goneall: Note that in 3.0 the author of SPDX data can still choose not to have a declared license, so not mapping it to a `NoAssertionLicense` would be OK too.  We just need to decide what to document in the migration guide.

I have a call with @swinslow this afternoon to sync up - I'll check with Steve and update the issue.

@maxhbr - Since this is specific to 2.3 migration and would only be documented in the migration guide IMO, can you approve this PR?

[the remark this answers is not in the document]
swinslow: This raises a question: I'm not sure what "not present" means, now that Declared License is a Relationship rather than a property. "Not present" in this particular SPDX document, BOM, bundle, namespace, ...? (Apologies as I'm not certain which term to use here!)

I believe the purpose of decoupling license values from the underlying software artifacts was in part to make it so that anyone could more easily create licensing values (via Relationships) for software artifacts that someone else had defined as SPDX Elements previously.

Given that, I'm not sure that a "null value" really has relevance anymore, unless I'm misunderstanding. Perhaps this line should just be deleted.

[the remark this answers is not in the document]
goneall: >This raises a question: I'm not sure what "not present" means, now that Declared License is a Relationship rather than a property. "Not present" in this particular SPDX document, BOM, bundle, namespace, ...? (

Couple of thoughts:
- Not present would just indicate that there is no `declaredLicense` relationship type found from a `SoftwareArtifact` to a license - no matter what was serialized - if you don't find it, it isn't present.
- There is no `null` value in the model itself - either a relationship is present or it isn't.  There is a null value if we treat it as a property.

I'll remove the null value reference.

goneall:
@swinslow @kestewart @maxhbr - I updated the PR with all the suggestions - if you could review and, assuming you don't find any new issues, I'll merge.

maxhbr (approved):
LGTM

PR #457, opened 2023-07-31
Rename ExternalReference to ExternalRef

### LINKED ISSUE #234: Rename ExternalReference and externalReference back to ExternalRef and externalRef
SPDX 2.X uses externalRef for the name of this property.

Although there has been an argument that `externalReference` is clearer, I don't think that justification is enough to cause migration effort.  Note that the choice of `externalRef` as the name in 2.X was chosen after a discussion and comparison to other terms such as `externalReference`.

goneall:
Resolves #234

Fixes an unnecessary incompatibility between SPDX 2.3 and SPDX 3.0 spec versions.

### DIFF
--- a/model/Core/Classes/Element.md
+++ b/model/Core/Classes/Element.md
@@ -45,8 +45,8 @@ and inter-relatable content objects.
   - maxCount: 1
 - verifiedUsing
   - type: IntegrityMethod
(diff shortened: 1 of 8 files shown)


kestewart (approved):
Matches what was discussed.

PR #472, opened 2023-08-10
Update suppliedBy.md

jeff-schutt:
Updates summary and description to allow the suppliedBy property to be used more flexibly than only for the SPDX 3.0 Artifact subclass or Element.


goneall (approved):
LGTM - matches what we discussed in the security team.

maxhbr (approved):
LGTM

Issue #436, opened 2023-07-26
Move originatedBy and suppliedBy properties back to Element


rnjudge:
This [PR](https://github.com/spdx/spdx-3-model/pull/292/files) updated the Range of `suppliedBy` from Element to Artifact. This needs to be reverted to the Range of `Element`. This issue was discovered and discussed at the Security profile call on July 26th. A summary of the discussion:

* Core profile and Security profile both include a `suppliedBy` property
* the Security profile uses `suppliedBy` for Vulnerability Elements, with the original intent to use the Element's `suppliedBy` property
* Recent creation of the Artifact class in the core model caused this duplication
* Discussed resolving this by having vulnerability elements inherit from artifact elements
  * Realized that this only solves half the problem, it will not resolve the use of `suppliedBy` in the relationships defined by the security profile (see `suppliedBy` used [here](https://github.com/spdx/spdx-3-model/blob/main/model/Security/Classes/CvssV3VulnAssessmentRelationship.md) in the Security profile. )
* **Recommendation**: move the `originatedBy` and `suppliedBy` properties back to the Element class
* **Justification**: they are being used as properties on the many relationships defined in the security profile, and likely to be used as properties on relationships in other profiles as well, e.g., licensing or AI

goneall:
The logic makes sense to me - I agree with the proposed change.

davaya:
There are two potential approaches:

1) put every property that any subclass uses in Element
2) put properties in the subclasses that use them

Annotation and Agent/Person/Organization don't have originatedBy and suppliedBy - the concept doesn't apply to things that aren't supplied.  Elements describing things that are supplied, like Artifact and its subclasses, do have supply-related properties.

Software defines a `SoftwareArtifact` class with more properties (`contentIdentifier`, `primaryPurpose`, etc.) that other Software classes inherit from.  Security already defines the `VulnAssessmentRelationship` class that other relationships inherit from.  I suggest adding `originatedBy` and `suppliedBy` properties to `VulnAssmentRelationship` to avoid polluting general Relationship elements along with Agents, Annotations, Collections, etc.

goneall:
related to #437

goneall:
@davaya makes a good argument for just adding the property to the Relationship subclasses used by security.

Reconsidering my previous opinion.

jeff-schutt:
Discussed in the security team meeting yesterday and agreed that:
[...]

PR #473, opened 2023-08-10
update Vulnerability.md

jeff-schutt:
Updates Vulnerability.md to inherit from Artifact rather than directly from Element. Part of the solution to #436


goneall (approved):
LGTM - thanks @jeff-schutt

maxhbr (approved):
LGTM

PR #475, opened 2023-08-12
replace JSON-LD text in JSON serialization with JSON text

davaya:


kestewart:
I believe this is stale at this point, and the serialization committee has resolved all open issues.   Please reopen if you disagree.

PR #479, opened 2023-08-17
think about a deserialization profile

maxhbr:
this was created when thinking about https://github.com/spdx/spdx-3-model/pull/403#issuecomment-1681816966


model/Deserialization/Classes/Deserialization.md
```diff
@@ -0,0 +1,36 @@
...
+
+TODO:
+- probably could benefit from a better class name!
+- encode the format?
+- allow to save the content as ByteString?
+
+## Metadata
+
+- name: Deserialization
+- SubclassOf: /Core/Artifact
```
goneall, line 27: I though about suggesting a subclass of File, but after thinking about it we would want to cover the case where something is serialized as a network stream.

model/Deserialization/Classes/Deserialization.md
```diff
@@ -0,0 +1,36 @@
...
+
+An element that can store information about a deserialization event, minted on the fly and not persisted in the payload.
+
+## Description
+
+tbd.
+
+Propreties should be used as follows:
+
+- element, to encode the content that was generated by when parsing the file
```
goneall, line 15: I'm a bit confused about the element property description - would this reference that actual elements in the graph deserialized from the file?

goneall: Makes sense - I may suggest some word smithing if we go down this route.

goneall:
Would I be able to handle the following scenario?:
- A single "Payload" is created by aggregating several existing previous Payloads (the commonly request merge function for the SPDX Tools)  Note that the namespace map used for this merged Payload is irrelevant to this scenario.
- You then want to "break apart" the single Payload back to the individual previous Payload and retain their original namespace maps - in this scenario, you don't care about the namespace map for the merged Payload

maxhbr:
Discussions moved along and are no happening in a different place. Closing this issue. In discussions it is now called the `X-Collection`

PR #480, opened 2023-08-17
Add licenseListVersion back to CreationInfo

### LINKED ISSUE #131: License Profile: add licenseListVersion to Creation Information
In the SPDX 2.3 spec, the license list version used for an SPDX document is captured as specified in [spec 6.7](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#67-license-list-version-field).

The current licensing-profile branch does not include a property of licenseListVersion or an extension to the CreationInformation class to include this as an optional property.

This may have just been an oversight.

goneall:
This fixes #131

### DIFF
--- a/model/Core/Classes/CreationInfo.md
+++ b/model/Core/Classes/CreationInfo.md
@@ -44,4 +44,5 @@ The dateTime created is often the date of last change (e.g., a git commit date),
   - type: xsd:string
   - minCount: 1
   - maxCount: 1
-
+- licenseListVersion
+  - type: xsd:string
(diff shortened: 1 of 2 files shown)


zvr:
I have some reservations about adding this to `CreationInfo` for _every_ Element (although it will not be mandatory). We had some discussions on `dataLicense` on that.

We really want it for things that have licenses -- or, more precisely, not even for licensed _things_ but for the license expressions themselves. So why don't we make it a property of these?

goneall:
> I have some reservations about adding this to `CreationInfo` for _every_ Element (although it will not be mandatory). We had some discussions on `dataLicen`

In anticipation of this - I've already prepared an alternative proposal to add it to just the license expression in SimpleLicensingInfo.

I'll create a pull request for review.

goneall:
Change to draft status until we decide which PR to merge.

goneall:
Closing this PR as #481 is the preferred solution to the license list version.

PR #481, opened 2023-08-18
Add licenseListVersion to LicenseExpression

### LINKED ISSUE #131: License Profile: add licenseListVersion to Creation Information

goneall:
Fixes #131

This adds the `licenseListVersion` which in 2.3 was at the document level to the `licenseExpression`.

This is alternative solution to PR #480

### DIFF
--- a/model/SimpleLicensing/Classes/LicenseExpression.md
+++ b/model/SimpleLicensing/Classes/LicenseExpression.md
@@ -25,6 +25,7 @@ SPDX License Expressions provide a way for one to construct expressions that mor
   - maxCount: 1
 - licenseListVersion
   - type: xsd:string
+  - maxCount: 1
 - customIdToUri
   - type: /Core/DictionaryEntry
   - minCount: 0
--- a/model/SimpleLicensing/Classes/LicenseExpression.md
+++ b/model/SimpleLicensing/Classes/LicenseExpression.md
@@ -23,6 +23,8 @@ SPDX License Expressions provide a way for one to construct expressions that mor
   - type: xsd:string
   - minCount: 1
   - maxCount: 1
+- licenseListVersion
+  - type: xsd:string
 - customIdToUri
   - type: /Core/DictionaryEntry
   - minCount: 0
--- a/model/SimpleLicensing/Classes/LicenseExpression.md
+++ b/model/SimpleLicensing/Classes/LicenseExpression.md
@@ -24,7 +24,7 @@ SPDX License Expressions provide a way for one to construct expressions that mor
   - minCount: 1
   - maxCount: 1
 - licenseListVersion
-  - type: xsd:string
+  - type: /Core/SemVer
   - maxCount: 1
 - customIdToUri
   - type: /Core/DictionaryEntry
--- a/model/SimpleLicensing/Properties/licenseListVersion.md
+++ b/model/SimpleLicensing/Properties/licenseListVersion.md
@@ -1,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# licenseListVersion
+
+## Summary
+
+The version of the SPDX License List used when the license expression was created.
+
+## Description
+
+Recognizing that licenses are added to the SPDX License List with each subsequent version, the intent is to provide consumers with the version of the SPDX License List used. This anticipates that in the future, license expression might have used a version of the SPDX License List that is older than the then current one.
+
+## Metadata
+
+- name: licenseListVersion
+- Nature: DataProperty
+- Range: xsd:string
+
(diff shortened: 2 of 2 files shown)


model/SimpleLicensing/Classes/LicenseExpression.md
```diff
@@ -23,6 +23,8 @@ SPDX License Expressions provide a way for one to construct expressions that mor
   - type: xsd:string
   - minCount: 1
   - maxCount: 1
+- licenseListVersion
+  - type: xsd:string
```
goneall, line 27: Add maxcount 1

goneall:
After discussing with @swinslow  - this seems to be the preferred approach to adding `licenseListVersion` - changing from draft to "ready for review"

@zvr - please review

model/SimpleLicensing/Properties/licenseListVersion.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# licenseListVersion
+
+## Summary
+
+The version of the SPDX License List used when the license expression was created.
```
zvr, line 7: ... "used in the license expression."

model/SimpleLicensing/Properties/licenseListVersion.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Description
+
+Recognizing that licenses are added to the SPDX License List with each subsequent version, the intent is to provide consumers with the version of the SPDX License List used. This anticipates that in the future, license expression might have used a version of the SPDX License List that is older than the then current one.
```
zvr, line 11: "The specific version of the SPDX License List must include all listed licenses and exceptions used in the expression." -- or something like that.

model/SimpleLicensing/Properties/licenseListVersion.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Metadata
+
+- name: licenseListVersion
+- Nature: DataProperty
+- Range: xsd:string
```
zvr, line 17: I know I'm opening a new topic, but should we be using `/Core/SemVer` like the specVersion? @swinslow , what do you think?

Totally fine with ignoring this for now and resolving it later in another issue, if it will add delay.

zvr (changes requested):

model/SimpleLicensing/Properties/licenseListVersion.md
goneall: Updated

model/SimpleLicensing/Properties/licenseListVersion.md
goneall: Updated

model/SimpleLicensing/Properties/licenseListVersion.md
goneall: Let's leave it as is for now.  Would making it `/Core/SemVer` less "simple"?

swinslow: I'd say future issue (as in "far future"). I think we've always used Major.Minor versioning for the license list and I don't know of any current reason to change it to Major.Minor.Patch.

swinslow (approved):

model/SimpleLicensing/Properties/licenseListVersion.md
zvr: Well, we used to use SPDX x.y for the spec number till now and we moved to SemVer; it might be ok to do the same here.

If not, we have to add a pattern for the valid values of this string, i.e making sure they match x.y -- and I think we have said that even single instances of patterns should be separate custom datatypes. Definitely more complicated.

But as I said, this might be a different issue and this can be merged.

goneall: I agree with @zvr that using SemVer would be simpler than creating a different datatype.

@swinslow - if you're OK with the change, I'll update this PR.  If not or if you're not sure, I'll merge this and we can open a separate PR to make progress.

swinslow: @goneall I'm fine with switching to SemVer if you're okay with it. I'm just conscious that it likely means changes to the License List Publisher so I didn't want to commit to it if you have qualms about it!

For the time being I don't expect we'd do anything other than x.y.0 releases, so maybe it's not that big of a deal in any case. But yes, I'm fine with it if you both are good with it.

goneall: Thanks @swinslow for the quick reply - I committed a change to use `/Core/SemVer`

@zvr - if you could take one more look and approve, I can merge

zvr (approved):

Issue #81, opened 2023-02-22
SPDX 2->3 conversion of ExternalPackageReferences


armintaenzertng:
This is part of #74:

Edit: updated after I understood that ExternalPackageReference will be split into two new properties. Some references will convert to ExternalIdentifier, others to ExternalReference. See the [migration documentation](https://docs.google.com/document/d/1-olHRnX1CssUS67Psv_sAq9Vd-pc81HF8MM0hA7M0hg/edit#) for a conversion guideline.

There remain two problems when the `category` is `OTHER`:
- In this case, `type` can be String-valued. Should this then be converted to SPDX3 `ExternalIdentifierType.other` or `ExternalReferenceType.other`?
- Also in this case, `locator` can be an arbitrary String (without spaces) in SPDX2, but the SPDX3 `ExternalReference.locator` must be a URI. `ExternalIdentifier.identifier` would be a String, but can we be sure that the SPDX2 reference of category `OTHER` is actually an identifier?

armintaenzertng:
The following SPDX 2 external package ref types don't seem to have an SPDX 3 counterpart:
- url
- maven-central
- npm
- nuget
- bower
How would they be sorted into identifier vs reference?

goneall:
I would suggest - reference.

kestewart:
This needs to be handled, but not considered a blocker for 3.0-rc

goneall:
Kate and I discussed this.  We probably should handle these in RC2 since it will impact the vocabulary.

goneall:
@maxhbr - with @armintaenzertng being out, would you mind adding a pull request to add the missing external reference types?

armintaenzertng:
I opened #484 to address the issue of missing ExternalRefTypes.

However, the initial question of this issue still stands, namely what to do with [the `OTHER` category](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f5-other) of SPDX2.
I'd think it would probably be safer to convert these to references instead of identifiers, but as pointed out above, SPDX3 identifiers expect a URI while the `OTHER` locator is an arbitrary string.

goneall:
> However, the initial question of this issue still stands, namely what to do with [the OTHER category](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f5-other) of SPDX2.
[...]

PR #484, opened 2023-08-24
Add missing SPDX2 external reference types

armintaenzertng:
part of #81


goneall:
@tsteenbe @rnjudge @jeff-schutt Just checking on the security reference types to make sure we should migrate these over from SPDX 2.X as is.  Please review/confirm.

goneall (approved):
LGTM - we should wait for another review from the security profile team before merging.

goneall:
Since it's been over 2 weeks and we haven't had a review from the security team, I'm going to go ahead and merge

Issue #467, opened 2023-08-04
Proposal: Move data license from CreationInfo to SpdxDocument


goneall:
In reviewing @zvr [comment on change proposal 8](https://github.com/spdx/change-proposal/issues/8#issuecomment-1665306618), we may want to consider moving the data license from `CreationInfo` to `SpdxDocument`.

This would relate the data license to the serialization information which is something we are already doing with fields like the original namespace.

I believe this would solve @zvr's concern and match the semantics of both `CreationInfo` and `SpdxDocument`.

Note that the use of `SpdxDocument` is entirely optional which should no longer be an issue if the data license is also optional.

zvr:
Why not have it a Relationship instead of a property?

goneall:
> Why not have it a Relationship instead of a property?

That is a reasonable alternative.  The advantage of the relationship is that it can be attached to any element and is more flexible.

There are a couple of downsides I can think of:
- You could have 2 different data license relationships pointing to the same element which could be confusing
- The data license could be added by someone other than the creator of the element at a later date

If we consider properties to be "intrinsic" to the element and consider the SpdxDocument to represent a "unit of serialization", then I could make arguments both for a property and a relationship depending on your perspective.

1. If you consider a license being assigned to a serialized SPDX document at creation time and unchanging, then a property is appropriate.
1. If you consider the serialized data to be independent of a license and different licenses can be applied in different situations, then the relationship is more appropriate.

I personally like the property approach since as a consumer, I can more trust the license not to change over time.

I think of a data license somewhat analogous to having a license statement in a package refer to a URL rather than clearly stating what the license is (e.g. an SPDX ID or notice).  As an auditor, this drives me nuts since I have to go to the wayback machine to see which license applied at the time the software was downloaded.

goneall:
Note this is implemented in PR #490

Issue #461, opened 2023-08-01
ML-Schema 


aamedina:
seeAlso: https://ml-schema.github.io/documentation/ML%20Schema.html

There is (or was?) a working group to define a RDF ML model, here is the GitHub: https://github.com/ML-Schema/core

Just adding a note here in case anyone wants to look at it for inspiration for the SPDX AI Profile.

aamedina:
seeAlso: https://bioportal.bioontology.org/ontologies/AIO

goneall:
@aamedina - We just release the RC2 version of the spec and we're going to setup a separate thread and/or meeting to go through this and all the other RDF/SHACL/OWL labelled SPDX issues.  I just sent out an email on the SPDX tech email list: https://lists.spdx.org/g/Spdx-tech/message/5538 

My current thinking is we have separate SPDX tech call just to focus on the RDF related issues and work through them.  Please reply to the email thread if you're available to participate.

goneall:
cc @bennetkl @kestewart @bact @rgopikrishnan91 - I wanted to make sure the AI team was aware of the parallel work mentioned above.

I'm going to close the issue as part of the 3.0 release clean-up

bennetkl:
Thanks @goneall  ,  We will watch for further discussions.

PR #490, opened 2023-08-31
Namespacemap reconstitution

sbarnum:
This is a PR for the proposed namespace map solution addressing all explicitly identified issues (layering of namespace maps, confliction of peer namespace maps, etc) with the previous solution.
It also provides more contextual detail around the intended semantics and use for the namespace map construct in the NamespaceMap.md class specification.


davaya:
The diagram represents what we have been discussing, but not what we have yet to discuss: externalMap, which is a critical consideration.

The three options for the X-collection Element semantics:
1) applies to a specific collection of elements in a specific data format (has a verifiedBy property)
2) applies to a specific collection of elements regardless of data format (does not have a verifiedBy property)
3) can be applied to future collections of elements (is a subclass of Element, not ElementCollection, and has no element property)

Currently externalMap is a property of ElementCollection, which means that an Sbom element applies only to a specific payload in a specific format.  The Consumer/Producer who reads a payload includes the hash of that payload in the list of elements in the newly-produced BOM.  If the Consumer/Producer read a different payload containing the identical elements he would need to create a different Sbom element.

So if ElementCollection is payload-specific, then its subclass X-Collection is also payload-specific (option 1).  At a minimum we want to make ElementCollection serialization-agnostic by removing externalMap from it.  But if we include externalMap in X-collection then X-collection is still serialization-specific (option 1).

**An instance of the X-collection element cannot be created until all of its externalMap property values are known.**

Those values are not known until the Consumer/Producer reads a specific payload, which is why X-collection instances cannot exist in the logical model until the serialized source of an element has been chosen.  As Max says, those instances can be inserted into the graph by a Consumer/Producer after reading a payload.  And as I say, those instances could also be inserted into the graph by the producer, but not until the producer has serialized the payload and knows its hash.

model/Core/Classes/NamespaceMap.md
```diff
@@ -0,0 +1,66 @@
...
+of ElementIDs to be used in SPDX content serialization in order to provide a more
+human-readable and smaller serialized representation of the Elements.
+
+The content of a namespace map are suggestions for consistency in serialization but are not normatively enforced.
+Any serialization of the collection content SHOULD (rather than MUST) utilize the namespace map content.
+A serialization MAY choose to use the namespace map content.
+A serialization MAY choose to use prefixes and namespaces other than the namespace map content.
+A serialization MAY choose to use no prefixes at all and rather use the more verbose full ElementID IRIs.
+
+If utilized the namespace map set of prefixes and namespaces MUST be implemented in a given serialization
```
goneall, line 22: It took me a few passes to parse this statement.  At first, it seemed to contradict the "MAY" choose to use, but now I understand that this is describing that we must not replace the native serialization.

Suggest restructuring the sentence to start with the serialization form - something like "The prefix / namespace mapping of the serialization format must always be used to convey the namespace mapping for that serialization ..."

model/Core/Classes/NamespaceMap.md
```diff
@@ -0,0 +1,66 @@
...
+A serialization MAY choose to use prefixes and namespaces other than the namespace map content.
+A serialization MAY choose to use no prefixes at all and rather use the more verbose full ElementID IRIs.
+
+If utilized the namespace map set of prefixes and namespaces MUST be implemented in a given serialization 
+form (e.g., json-ld or xml) as specified in the binding rules specification for that serialization and 
+utilizing the appropriate inherent or custom specified mechanism for that serialization.
+The namespace map itself is also conveyed as native SPDX content to support clarity, transparency and 
+consistency independent of any particular serialization form.
+
+A given serialization payload (whether file or streaming) MUST NOT contain multiple namespace maps with conflicting mappings.
```
goneall, line 28: I take it the above sentence refers to the "native" namespace mapping.  If so, this may not be a required statement since the serialization format standards already have this requirement.

model/Core/Classes/NamespaceMap.md
```diff
@@ -0,0 +1,66 @@
...
+In this case the serializer MUST do one of the following:
+1) Diverge from the exact provided namespace mappings and utilize deconflicted mappings where appropriate.
+   This could be fully new prefix mappings as if the suggested namespace maps had not been provided.
+3) Use full ElementID IRIs instead of prefixed ElementID IRIs.
+
+
+Namespace maps support a variety of relevant use cases such as:
+
+1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.
+2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serializations of the produced content.
```
goneall, line 40:
```suggestion
2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serialization formats of the produced content.
```

model/Core/Classes/NamespaceMap.md
```diff
@@ -0,0 +1,66 @@
...
+   For example, an SBOM producer wishes to share/publish the SBOM as json-ld, XML and YAML. The producer can specify the preferred prefix mappings they plan to use in all serializations, maintain usage consistency themselves across those serializations and provide clarity to the consumer of the SBOM in any of those serializations independent of which serialization form was received.
+3) An SPDX content consumer/producer wishing to maintain consistent prefix use while round tripping from SPDX content received, deserialized, modified/extended in some way, and then reserialized in the same serialization form.
+   In this case the prefix-namespace mappings utilized in the content are preserved across serializations as native SPDX content. 
+4) An SPDX content consumer/producer wishing to maintain consistent prefix use while trans-serializing SPDX content received in one serialization form, deserialized, and then reserialized into a different serialization form.
```
goneall, line 44: Is this the same as 2?

model/Core/Classes/NamespaceMap.md
```diff
@@ -0,0 +1,66 @@
...
+   In this case the prefix-namespace mappings utilized in the content are preserved across serializations as native SPDX content. 
+5) An SPDX content producer wishing to serialize/share together multiple collections of SPDX content each with their own namespace map in a way any consumer of the content would be able to separate out the multiple collections including their namespace map suggestions.
+   This would involve the conflicted mappings issue briefly characterized above this list of use cases.
+6) An SPDX content consumer wishing to maintain consistent prefix use while receiving serialized content that does not include a namespace map but does utilize prefixes, and at some future point reserializing that content.
+   The consumer can simply "wrap" the received content in a collection with a namespace map and specify the prefix to namespace mappings that were actually implemented in the received content.
+7) It should be possible to derive and maintain namespace mapping provenance for content.
```
goneall, line 50: Why would this be important?

goneall (commented):
Thanks @sbarnum for writing this up - A few comments to consider for the tech call.

model/Core/Classes/ElementCollection.md
```diff
@@ -19,10 +19,10 @@ An SpdxCollection is a collection of Elements, not necessarily with unifying con
 ## Properties
 
 - element
-  - type: Element
+  - type: Element and NOT (X-Collection)
```
goneall, line 22: Note: the spec parser does not currently support this expression

model/Core/Classes/ElementCollection.md
```diff
@@ -19,10 +19,10 @@ An SpdxCollection is a collection of Elements, not necessarily with unifying con
...
   - minCount: 1
 - rootElement
-  - type: Element
+  - type: Element and NOT (X-Collection)
```
goneall, line 25: Same as above

goneall (commented):
Couple more comments

model/Core/Classes/ElementCollection.md
sbarnum: That does not surprise me.
It will need to though.
This sort of issue I why I strongly prefer having the formal ontology specification (RDFS/OWL/SHACL) be THE ground truth specification rather than a prose form that must be converted. 
Specifying this sort of range is simple, native and inherent in the RDFS/OWL/SHACL.

model/Core/Classes/ElementCollection.md
sbarnum: Same as above for me too. :-)

model/Core/Classes/NamespaceMap.md
sbarnum: I am not sure if your characterization is aligned to intent here.
Lines 22-26 all go together.
They are saying that if you are going to actual use the namespace map content as prefixes in a given serialization then how those prefixes are represented in a given serialization form (e.g., json-ld, xml, etc) varies form to form and the implementation of how they are represented in any given form is specified in the binding rules (that bind the model to a given serialization form) for that serialization form. In other words, if you are going to implement them in json-ld you must use the specified form for prefixes in the json-ld binding rules specification. For serialization forms that natively support prefixes it is a little more obvious. For serialization forms that don't and we have to define custom representations this is much more of an important clause.

I believe the whole first sentence (lines 22-24) clearly states this point as is. 
Does the above explanation clarify or do you still find the sentence unclear?

model/Core/Classes/NamespaceMap.md
sbarnum: Serialization formats that natively support prefixes usually have this requirement though some do not enforce it and rather simply utilize the last defined prefix and ignore any earlier conflicting prefixes.
Serialization formats that do not natively support prefixes and where we will have to define custom prefix representations will have no such rules.

In either case, I would propose making this explicit statement is highly useful in the SPDX spec.

sbarnum:
> X-Collection.md says only:
> 
> > An X collection Element is a collection of Elements that does not contain any other X Collection Elements.
> > In this way it can be thought of as an outer shell collection of SPDX content without self-recursion that can be used as a content aggregation target for serialization.
> 
> There are no use cases or requirements for why, where, or when it is used or what it accomplishes, given the requirement in NamespaceMap.md that:
> 

We do not typically put " use cases or requirements for why, where, or when it is used or what it accomplishes" in the class specifications.
The class specification for NamespaceMap (NamespaceMap.md) atypically provides this greater level of detail for the namespace map rather than attempting to put is here at least one level abstracted from its direct application.
That being said I can see that the X-Collection.md could use a brief statement that it is intended to convey a namespace map for a given set of SPDX content and that it MAY (though not MUST) be used as the single outermost enclosing SPDX element in specific instance of serialization for simplicity and consistency.

> > the namespace map set of prefixes and namespaces MUST be implemented in a given serialization form.
> 

The full statement from NamespaceMap.md that this is snippeted from is:
"If utilized the namespace map set of prefixes and namespaces MUST be implemented in a given serialization form (e.g., json-ld or xml) as specified in the binding rules specification for that serialization and utilizing the appropriate inherent or custom specified mechanism for that serialization. The namespace map itself is also conveyed as native SPDX content to support clarity, transparency and consistency independent of any particular serialization form."

This is saying that if you are going to actually use the namespace map content as prefixes in a given serialization then how those prefixes are represented in a given serialization form (e.g., json-ld, xml, etc) varies form to form and the implementation of how they are represented in any given form is specified in the binding rules (that bind the model to a given serialization form) for that serialization form. In other words, if you are going to implement them in json-ld you must use the specified form for prefixes in the json-ld binding rules specification. For serialization forms that natively support prefixes it is a little more obvious. For serialization forms that don't and we have to define custom representations this is much more of an important clause.


> Without a justification, there is no reason to define an X-Collection element. 

As stated above, I can see value in adding a brief statement to X-Collection.md to clarify its intent though I do not believe class specification must or should have to "justify" their existence.

> Or it could be defined in the logical model with the requirement that it never be serialized. 

No. The entire purpose of namespace maps are that they are serialized as part of the content.

> The amorphous "X" concept represents what we have been discussing, but not what we have yet to discuss: externalMap, which is a critical consideration. The three options for the X-collection (or X-file) element semantics are:
> 

I would assert that externalMap is a completely separate and mostly unrelated topic than what we are discussing here and am unaware of any currently identified issues with its current implementation.

I would caution not using any names like X-File in relation to X-Collection. X-Collection is NOT intended to deal with a specific instance of serialization.


> 1. applies to a specific collection/file/payload of elements in a specific data format (has a verifiedBy property)
> 2. applies to a specific logical collection of elements regardless of data format (does not have a verifiedBy property)
> 3. can be applied to future collections of elements (X is a subclass of Element, not ElementCollection, and has no element property)
> 
> One of those three must be chosen before an X element can be defined.
> 

The very explicit intent of namespace map is 2.
It is very explicitly not tied to any specific instance of serialization.

> Currently externalMap is a property of ElementCollection, which means that an Sbom element applies only to a specific payload in a specific format. The Consumer/Producer who reads a payload includes the hash of that payload in the list of elements in the newly-produced BOM. If the Consumer/Producer had read a different payload containing the identical elements he would need to create a different Sbom element.
> 

I am confused here. How did externalMap get into this conversation? It is unrelated.

> Currently externalMap is a property of ElementCollection, which means that an Sbom element applies only to a specific payload in a specific format. 

ExternalMap on ElementCollection makes no such implication on Sbom elements. Sbom elements, or any part of the SPDX model other than the File object, have nothing to do with any specific instance (payload) of serialization.

> So if ElementCollection is payload-specific, then its subclass X-Collection is also payload-specific (option 1). At a minimum we want to make ElementCollection serialization-agnostic by removing externalMap from it. But if we include externalMap in X-collection then X-collection is still serialization-specific (option 1).
> 

ElementCollection is definitely not payload specific and neither is X-Collection.

> **An instance of the X-collection element cannot be created until all of its externalMap property values are known.**
> 

Again, I am confused at how externalMap got introduced here or how its meaning and intent got so confused.

> Those values are not known until the Consumer/Producer reads a specific payload, which is why X-collection instances cannot exist in the logical model until the serialized source of an element has been chosen. As Max says, those instances can be created and inserted into the graph by a Consumer/Producer after reading a payload. And as I say, those instances could also be created and inserted into the graph by the producer, but not until the producer has serialized the payload and knows its signature/hash.
> 
> This boils down to my original objection from [9 months ago](https://github.com/spdx/spdx-3-model/pull/57/files?short_path=24732ee#diff-24732eea2c3cf0bec0853074cbdac3492f56b8a4164ec5cc4331a5e33db4949c): the logical model should be serialization-agnostic. A logical SBOM can contain two File elements, for a total of three. Serialization independence means its serialized payload would contain 3 elements (SBOM, File1, File2), not be forced to serialize 4 elements (X-collection plus the other three.) Consumers of that payload have the option of creating the X-collection element, but don't have to unless they produce other payloads that depend on (reference) the first payload. In particular, the Consumer could copy those three element values into his new payload instead of referencing them, in which case no X-collection element is ever needed.

I agree that the logical model should be serialization-agnostic. It should also be agnostic of any serialization instance.

goneall:
In the namespace meeting on 18 Sept 2023, we decided to move forward with the proposal documented in pull request #491 

PR #491 is now merged into this pull request so we can review a single PR before merging into the base.

Note that the branch used for PR #491 was not deleted, so we can refer back to the changes if needed.  I also did a merge commit so it would be easy to reconstruct the state of this PR prior to the merge.

Please review and comment on the wording for this proposal.

davaya:
PR #500 includes NamespaceMap in payload data, allowing it to be used independently of SerializedCollection.  This does not prevent SerializedCollection from defining NamespaceMap independently of serialization if there are use cases for doing so.

goneall:
@sbarnum - Do you want to update this PR with the decisions from the serialization team?
- Move `externalMap` from `ElementCollection` to `X-Collection`

model/Core/Classes/X-Collection.md
```diff
@@ -0,0 +1,30 @@
...
+# X-Collection
+
+## Summary
+
+A collection of SPDX Elements that could potentially be serialized.
+
+## Description
+
+The X-Collection provides a convenient way to express information about collections of SPDX Elements that could potentially be serialized as complete
+units (e.g., all in-scope SPDX data within a single JSON-LD file). X-Collection is indpendent of any particular serialization format or instance.
```
goneall, line 12:
```suggestion
units (e.g., all in-scope SPDX data within a single JSON-LD file). X-Collection is independent of any particular serialization format or instance.
```

goneall (changes requested):
One minor typo

goneall:
@sbarnum - can you rename the X-Collection to SpdxDocument?

model/Core/Classes/X-Collection.md
sbarnum: @goneall LOL. It took me three reads through to notice my typo you corrected. I will fix it.

goneall (approved):
LGTM - Thanks @sbarnum

goneall:
Note that this is a previous discussion in the PR merged into this: #461

goneall:
Fixes #467

goneall:
Fixes #415

maxhbr:
related: https://github.com/spdx/spdx-3-model/issues/557

nishakm (approved):

goneall:
@nishakm @maxhbr @zvr - I resolved the merge conflicts - ready for review.

Issue #489, opened 2023-08-31
Alternative proposal for namespaceMaps


maxhbr:
add a new Element to the Model, that describes a serialized artifact.

# Data it contains

- It encodes the `NamespaceMap`
- It points to the elements which are serialized in the element (like a collection)
- It might also contain
  - name of the file
  - hash of the file

# semantics
- when deserializing, this element is minted on the fly
  - the NamespaceMap is filled from the actually used namespaces
- when serializing, one can optionally provide this as a description what and how it should be serialized, but it is not written itself to the generated output

goneall:
Since we made a decision on the namespaceMaps, I'm closing this issue to avoid confusion.

Issue #478, opened 2023-08-16
Where do we start in a serialized JSON-LD?


goneall:
In the [current JSON-LD example](https://github.com/spdx/spdx-3-model/blob/main/serialization/json_ld/examples/converted_from_spdx_2.json), all elements are a flat array - If I receive a file, how do I know where to start?  If I send an SBOM along with a package, there my be 10's of thousands of elements including SBOM's within SBOM's.  How do I know which element to start with?

In SPDX 2.X the SPDX document is 1:1 with the serialization and has a describes relationship which give you starting point(s).

goneall:
One possibility is to restrict JSON-LD serializations to one and only one `ElementCollection` type.

davaya:
JSON (#430) is also a flat list of elements because the purpose of serialization is to put chunks of the graph into documents, not for documents to model graphs.  A document is thus a forest (multiple trees) and the root of each tree is an element whose type is a subclass of ElementCollection.  That's just the first graph layer, the second is created by Relationship elements.

Serialization (and thus documents) shouldn't care about graphs.  After parsing a document there will be zero, one, or more ElementCollections, and that's where you start.

goneall:
> JSON (#430) is also a flat list of elements because the purpose of serialization is to put chunks of the graph into documents, not for documents to model graphs. A document is a forest (multiple trees) and the root of each tree is an`ElementCollection` type (plus disconnected elements not referenced by any collection in the document). That's just the first graph layer, the second is created by Relationship elements.
> 
> Serialization (and thus documents) shouldn't care about graphs. After parsing a document there will be zero, one, or more `ElementCollections`, and those are where you start. If collections are nested [topo sort](https://en.wikipedia.org/wiki/Topological_sorting) will find the root(s). Whatever is left over are "leaves on the ground" :-).

In my opinion, this is much more complicated than the 2.X serialization where we always have a top level with a description of where to start.  Although it may be technically possible to find the trees, it doesn't seem very easy.  It seems like it is left to the consumer of an SPDX serialization to reverse engineer what the producer intended to send rather than the being explicitly stated (as in SPDX 2.X).

maxhbr:
> One possibility is to restrict JSON-LD serializations to one and only one `ElementCollection` type.

That sounds fairly restrictive

goneall:
The use case I'm concerned about is:
- A software supplier produces a file containing an SBOM representing the specific package/version being distributed
- A consumer of the software would like to read the SBOM

[...]

PR #491, opened 2023-09-05
Attempt to implement alternative proposal for namespaceMap

goneall:
This is an attempt to implement the proposal documented in issue #489 as a pull request on top of pull request #490 

I took the approach of a pull request on top of the alternative pull request since the structure is basically the same.  It is just the descriptions which are different.

Note that the basic structure is also the same as pull request #403 - the main difference being the names of the properties and classes.

### DIFF
--- a/model/Core/Classes/NamespaceMap.md
+++ b/model/Core/Classes/NamespaceMap.md
@@ -29,7 +29,7 @@ Namespace maps support a variety of relevant use cases such as:
 
 1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.  The consumer of SPDX content wishes to preserve the name space mapping provided by such a producer.  In this case, the consumer would record the namespace map prefixes in the NamespaceMap such that subsequent serializations could reproduce the prefixes / namespaces in the native serialization format.
 2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serialization formats of the produced content.
-   For example, an SBOM producer wishes to share/publish the SBOM as JSON-LD and XML. The producer can specify the preferred prefix mappings in the native serialization format using the information from a single Namespace stored in memory.
+   For example, an SBOM producer wishes to share/publish the SBOM as JSON-LD and XML. The producer can specify the preferred prefix mappings in the native serialization format using information from a single Namespacemap accessible local to the producer.
 3) An SPDX content consumer/producer wishing to maintain consistent prefix use while round tripping from SPDX content received, deserialized, modified/extended in some way, and then reserialized in the same serialization form.
    In this case the prefix-namespace mappings utilized in the content are transformed from the original native namespace/prefix into the in memory NamespaceMap then transformed from the NamespaceMap back into the resultant serialization native namespace / prefix format.
 
--- a/model/Core/Classes/NamespaceMap.md
+++ b/model/Core/Classes/NamespaceMap.md
@@ -22,7 +22,6 @@ A serialization MAY choose to use no prefixes at all and rather use the more ver
 
 If a serialization format supports prefixes or namespaces (e.g., JSON-LD context or XML XSD namespace), the namespace map
 must be represented in that format "native" to the serialization.
-The NamespaceMap itself will never be serialized as part of SPDX data if the serialization format support namespaces or prefixes.
 If the serialization format does not support prefixes, then the full URI's for the elements must be used and the namespace map will not be preserved.
 Any custom serialization format SHOULD implement namespaces in order to preserve the namespace map.
 
--- a/model/Core/Classes/NamespaceMap.md
+++ b/model/Core/Classes/NamespaceMap.md
@@ -14,40 +14,25 @@ of ElementIDs to be used in SPDX content serialization in order to provide a mor
 human-readable and smaller serialized representation of the Elements.
 
 The content of a namespace map are suggestions for consistency in serialization but are not normatively enforced.
-Any serialization of the collection content SHOULD (rather than MUST) utilize the namespace map content.
+When a serialization only contains the elements in "X", the collection content SHOULD (rather than MUST) utilize
+the namespace map.
 A serialization MAY choose to use the namespace map content.
 A serialization MAY choose to use prefixes and namespaces other than the namespace map content.
 A serialization MAY choose to use no prefixes at all and rather use the more verbose full ElementID IRIs.
 
-If utilized the namespace map set of prefixes and namespaces MUST be implemented in a given serialization 
-form (e.g., json-ld or xml) as specified in the binding rules specification for that serialization and 
-utilizing the appropriate inherent or custom specified mechanism for that serialization.
-The namespace map itself is also conveyed as native SPDX content to support clarity, transparency and 
-consistency independent of any particular serialization form.
-
-A given serialization payload (whether file or streaming) MUST NOT contain multiple namespace maps with conflicting mappings.
-Any serialization of the collection content SHOULD (rather than MUST) utilize the namespace map content unless the single 
-serialization payload is intended to contain multiple peer collections each with their own namespace map.
-In this case the serializer MUST do one of the following:
-1) Diverge from the exact provided namespace mappings and utilize deconflicted mappings where appropriate.
-   This could be fully new prefix mappings as if the suggested namespace maps had not been provided.
-3) Use full ElementID IRIs instead of prefixed ElementID IRIs.
-
+If a serialization format supports prefixes or namespaces (e.g., JSON-LD context or XML XSD namespace), the namespace map
+must be represented in that format "native" to the serialization.
+The NamespaceMap itself will never be serialized as part of SPDX data if the serialization format support namespaces or prefixes.
+If the serialization format does not support prefixes, then the full URI's for the elements must be used and the namespace map will not be preserved.
+Any custom serialization format SHOULD implement namespaces in order to preserve the namespace map.
 
 Namespace maps support a variety of relevant use cases such as:
 
-1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.
-2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serializations of the produced content.
-   For example, an SBOM producer wishes to share/publish the SBOM as json-ld, XML and YAML. The producer can specify the preferred prefix mappings they plan to use in all serializations, maintain usage consistency themselves across those serializations and provide clarity to the consumer of the SBOM in any of those serializations independent of which serialization form was received.
+1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.  The consumer of SPDX content wishes to preserve the name space mapping provided by such a producer.  In this case, the consumer would record the namespace map prefixes in the NamespaceMap such that subsequent serializations could reproduce the prefixes / namespaces in the native serialization format.
+2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serialization formats of the produced content.
+   For example, an SBOM producer wishes to share/publish the SBOM as JSON-LD and XML. The producer can specify the preferred prefix mappings in the native serialization format using the information from a single Namespace stored in memory.
 3) An SPDX content consumer/producer wishing to maintain consistent prefix use while round tripping from SPDX content received, deserialized, modified/extended in some way, and then reserialized in the same serialization form.
-   In this case the prefix-namespace mappings utilized in the content are preserved across serializations as native SPDX content. 
-4) An SPDX content consumer/producer wishing to maintain consistent prefix use while trans-serializing SPDX content received in one serialization form, deserialized, and then reserialized into a different serialization form.
-   In this case the prefix-namespace mappings utilized in the content are preserved across serializations as native SPDX content. 
-5) An SPDX content producer wishing to serialize/share together multiple collections of SPDX content each with their own namespace map in a way any consumer of the content would be able to separate out the multiple collections including their namespace map suggestions.
-   This would involve the conflicted mappings issue briefly characterized above this list of use cases.
-6) An SPDX content consumer wishing to maintain consistent prefix use while receiving serialized content that does not include a namespace map but does utilize prefixes, and at some future point reserializing that content.
-   The consumer can simply "wrap" the received content in a collection with a namespace map and specify the prefix to namespace mappings that were actually implemented in the received content.
-7) It should be possible to derive and maintain namespace mapping provenance for content.
+   In this case the prefix-namespace mappings utilized in the content are transformed from the original native namespace/prefix into the in memory NamespaceMap then transformed from the NamespaceMap back into the resultant serialization native namespace / prefix format.
 
 ## Metadata
 
--- a/model/Core/Classes/X-Collection.md
+++ b/model/Core/Classes/X-Collection.md
@@ -12,7 +12,7 @@ The X-Collection provides a convenient way to store information about SPDX data 
 unit (e.g., all SPDX data within a single JSON-LD file).
 Information we wish to preserve about the serialization itself are store as properties of this class.
 
-NOTE: This class is not intended to be serialized itself.
+NOTE: This class is not intended to be serialized within the SPDX data it is representing.
 It is intended to be utilized within a system to handle the following use cases:
 
 1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.  The consumer of SPDX content wishes to preserve the name space mapping provided by such a producer.  In this case, the consumer would record the namespace map prefixes in the NamespaceMap such that subsequent serializations could reproduce the prefixes / namespaces in the native serialization format.
--- a/model/Core/Classes/X-Collection.md
+++ b/model/Core/Classes/X-Collection.md
@@ -4,14 +4,22 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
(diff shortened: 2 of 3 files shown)


goneall:
@maxhbr - I tried to implement your proposal as a pull request.

This is based on a branch in the github repo - feel free to update the branch to better represent your proposal before the tech call.

davaya:
Re: X-Collection.md:

> Information we wish to preserve about the serialization itself are store as properties of this class.
Information we wish to preserve about the serialized data itself are stored as properties of this class.

> NOTE: This class is not intended to be serialized itself.
The use case for ExternalMap still needs to be considered.  It must be removed from ElementCollection if we want to avoid locking SBOMs to a single serialized data instance.

That means it must be put somewhere else, and X-Collection is a candidate for holding it.  In that case, X-Collection instances would be serialized by ConsumerProducers who reference external serialized data from their serialized data.

maxhbr:
I started it at some point in https://github.com/spdx/spdx-3-model/pull/479

model/Core/Classes/X-Collection.md
```diff
@@ -4,14 +4,22 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 ## Description
 
-An X collection Element is a collection of Elements that does not contain any other X Collection Elements.
+The X-Collection provides a convenient way to store information about SPDX data which has been serialized as a complete
+unit (e.g., all SPDX data within a single JSON-LD file).
+Information we wish to preserve about the serialization itself are store as properties of this class.
 
-In this way it can be thought of as an outer shell collection of SPDX content without 
-self-recursion that can be used as a content aggregation target for serialization.
+NOTE: This class is not intended to be serialized itself.
```
maxhbr, line 15: But no one prevents it from being serialized. It just has no implication to the serialized blob it contains.

model/Core/Classes/NamespaceMap.md
```diff
@@ -14,40 +14,25 @@ of ElementIDs to be used in SPDX content serialization in order to provide a mor
...
-Any serialization of the collection content SHOULD (rather than MUST) utilize the namespace map content unless the single 
-serialization payload is intended to contain multiple peer collections each with their own namespace map.
-In this case the serializer MUST do one of the following:
-1) Diverge from the exact provided namespace mappings and utilize deconflicted mappings where appropriate.
-   This could be fully new prefix mappings as if the suggested namespace maps had not been provided.
-3) Use full ElementID IRIs instead of prefixed ElementID IRIs.
-
+If a serialization format supports prefixes or namespaces (e.g., JSON-LD context or XML XSD namespace), the namespace map
+must be represented in that format "native" to the serialization.
+The NamespaceMap itself will never be serialized as part of SPDX data if the serialization format support namespaces or prefixes.
```
maxhbr, line 25: That is not strictly true, as it is just a SPDX element which could be serialized. But it has no semantic relation to the file which contains it.

model/Core/Classes/NamespaceMap.md
```diff
@@ -14,40 +14,25 @@ of ElementIDs to be used in SPDX content serialization in order to provide a mor
...
+If the serialization format does not support prefixes, then the full URI's for the elements must be used and the namespace map will not be preserved.
+Any custom serialization format SHOULD implement namespaces in order to preserve the namespace map.
```
maxhbr, line 27: I think this part should be described in [X-Collection.md](https://github.com/spdx/spdx-3-model/pull/491/files#diff-d87f35d07a17800a2a87a8d1f4a999b5816c4ab00435322d5f48db992bb7521a). The X-Collection should contain a description how it is expected to be created when deserializing a blob.

model/Core/Classes/NamespaceMap.md
goneall: I'll remove this statement

goneall: After reading David's comment below, it sounds like there are different views on whether this should be allowed to be serialized.

When the proposal was described, I thought it would never be included in an actual serialization.

model/Core/Classes/X-Collection.md
goneall: @maxhbr - Can you propose some specific text?

model/Core/Classes/NamespaceMap.md
maxhbr: > When the proposal was described, I thought it would never be included in an actual serialization.

I think it is slightly different.

If I send such a `X` to the serializer. I Expect it to serialize all Elements that are contained while following the namespace map. And therefore it would not serialize the `X` itself. 

*But:* The `X` might contain another instance of `X` which would be serialized. There would then be no implication from the second `X` to the file it contains.

davaya:
September 6 - What we agree on:
- A subset of the use cases (some new ones were added in PR's which haven't been discussed)
- NamespaceMap will be in the model
  - Namespaces are portions of Element ID IRI's
- The model will always use the full IRI in every element
- NamespaceMap are specific to supporting serialization (check)
- All model proposals have an "X-Collection" as an outer shell
- NamespaceMaps are only useful in collections
- NamespaceMaps can only be used on collections
- Serialized data is composed of one or more element values that may or may not be included in an ElementCollection

Bob articulated the distinction between capturing prefixes in model data after serialization or before:
Open questions:
 1. Prefixes can be tied to a specific serialization instance "payload"
 2. Prefixes can be tied to a specific set of elements that are not (yet) serialized
 3. Not tied to any set of elements - just part of the model <- out of scope for current SPDX release

The identical model can support both 1 and 2:
 1. Prefixes are serialized using format-specific syntax; serialized payloads are described by SpdxDocument elements
 2. Prefixes are defined in model data as a property of ElementCollection (Bundle, Sbom, etc) elements.
 
The distinction is operational.  With 2. generating an outer X element, those elements will continue to accumulate in the model data as collections use other collections:  a producer creates  Sbom1 with namespaceMap, another producer creates an Sbom2 with namespaceMap, a third producer creates a Bundle / Sbom3 that includes Sbom1 and Sbom2 with a third namespaceMap, and as the model graph continues to grow, all of the previously outer collections become inner when they are added to a new outer collection.

With 1. SpdxDocument elements are created only as needed when Sbom3 references a document containing Sbom1.  Sbom1 doesn't have to contain namespaceMap at all, so nested namespaceMap data doesn't build up in the model graph as reference chains grow in length.

In the harmonized model namespaceMap is an optional property of ElementCollection, and if present their instances are hints to be used when serializing documents and creating SpdxDocument elements.  The term "X" isn't a new box in the model, it is a reference to either SpdxDocument or ElementCollection.

davaya:
**Payload**:  Sean objects to the term payload because of the implication that it means a file.  But we have been careful to articulate that  payload includes all methods of transferring data including streaming sessions and online data store queries.

Nodes in the element graph are immutable, but nodes will be added over time.  If "serialization" is used to refer to database storage, then a serialized data unit is a specific storage transaction, and prefixes can be used to reduce stored data and/or query result size just as they reduce the size of data serialized into files.  The write transaction is the database equivalent of a filesystem file entry or inode, and the bytes returned by reading the database transaction are equivalent to the bytes returned by reading a filesystem file.  The namespaceMap used in a write transaction is returned when reading it.  If the database software does not support immutable state as of a particular transaction, then it cannot implement an immutable element graph.

maxhbr:
> ...  then it cannot implement an immutable element graph ...

I do not understand the issue. At even in a moving and alive DB every element is immutable and therefore every `Collections` scopes out an immutable sub-graph within the DB. Isn't that sufficient?

davaya:
The question is how to represent units of serialized data (byte sequences that can be hashed) without regard to the content of those bytes.  Is there a one-to-one correspondence between creation of an immutable subgraph within the DB and a byte sequence that represents that creation action?

model/Core/Classes/X-Collection.md
goneall: This was updated after our conversation before the tech call on 5 Sept

model/Core/Classes/NamespaceMap.md
goneall: This statement was removed

model/Core/Classes/NamespaceMap.md
```diff
@@ -14,40 +14,24 @@ of ElementIDs to be used in SPDX content serialization in order to provide a mor
...
+1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.  The consumer of SPDX content wishes to preserve the name space mapping provided by such a producer.  In this case, the consumer would record the namespace map prefixes in the NamespaceMap such that subsequent serializations could reproduce the prefixes / namespaces in the native serialization format.
+2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serialization formats of the produced content.
+   For example, an SBOM producer wishes to share/publish the SBOM as JSON-LD and XML. The producer can specify the preferred prefix mappings in the native serialization format using the information from a single Namespace stored in memory.
```
sbarnum, line 32: > The producer can specify the preferred prefix mappings in the native serialization format using the information from a single Namespace stored in memory.
I do not understand what this added sentence is trying to convey.

model/Core/Classes/NamespaceMap.md
```diff
@@ -14,40 +14,24 @@ of ElementIDs to be used in SPDX content serialization in order to provide a mor
...
-2) An SPDX content producer wishing to maintain consistent prefix use and understanding across multiple different serializations of the produced content.
-   For example, an SBOM producer wishes to share/publish the SBOM as json-ld, XML and YAML. The producer can specify the preferred prefix mappings they plan to use in all serializations, maintain usage consistency themselves across those serializations and provide clarity to the consumer of the SBOM in any of those serializations independent of which serialization form was received.
+1) An SPDX content producer wishing to provide clarity of their serialization of an SPDX 2.X simple style collection where all content is newly minted and a single prefix-namespace is used.  The consumer of SPDX content wishes to preserve the name space mapping provided by such a producer.  In this case, the consumer would record the namespace map prefixes in the NamespaceMap such that subsequent serializations could reproduce the prefixes / namespaces in the native serialization format.
```
sbarnum, line 30: > The consumer of SPDX content wishes to preserve the name space mapping provided by such a producer.  In this case, the consumer would record the namespace map prefixes in the NamespaceMap such that subsequent serializations could reproduce the prefixes / namespaces in the native serialization format.

It is unclear exactly what is meant by the consumer to "record" the namespace map. Is this saying they should store it as SPDX content along with their other SPDX content but never serializa it out?
It is also unclear what is meant by "subsequent serialization". Does this many a single reserialization of the content, ANY future serializations of the entire content or ANY future serializations of portions of this content?

model/Core/Classes/NamespaceMap.md
```diff
@@ -14,40 +14,24 @@ of ElementIDs to be used in SPDX content serialization in order to provide a mor
...
-5) An SPDX content producer wishing to serialize/share together multiple collections of SPDX content each with their own namespace map in a way any consumer of the content would be able to separate out the multiple collections including their namespace map suggestions.
-   This would involve the conflicted mappings issue briefly characterized above this list of use cases.
-6) An SPDX content consumer wishing to maintain consistent prefix use while receiving serialized content that does not include a namespace map but does utilize prefixes, and at some future point reserializing that content.
-   The consumer can simply "wrap" the received content in a collection with a namespace map and specify the prefix to namespace mappings that were actually implemented in the received content.
-7) It should be possible to derive and maintain namespace mapping provenance for content.
+   In this case the prefix-namespace mappings utilized in the content are transformed from the original native namespace/prefix into the in memory NamespaceMap then transformed from the NamespaceMap back into the resultant serialization native namespace / prefix format.
```
sbarnum, line 34: If the namespace map that the consumer is constructing from received content is only in memory (ephemeral) and not even persisted internally then this proposed consumer-centric approach will only ever offer consistency for a single "hop". This question of what this proposal proposes NamespaceMap to be is still quite unclear to me.

model/Core/Classes/NamespaceMap.md
goneall: Change to:

The producer can specify the preferred prefix mappings in the native serialization format using information from a single Namespacemap accessible local to the producer.

model/Core/Classes/NamespaceMap.md
goneall: > It is unclear exactly what is meant by the consumer to "record" the namespace map. Is this saying they should store it as SPDX content along with their other SPDX content but never serializa it out?

This was clarified by the correction above

> t is also unclear what is meant by "subsequent serialization". Does this many a single reserialization of the content, ANY future serializations of the entire content or ANY future serializations of portions of this content?

Both - any future serializations in full or part

model/Core/Classes/NamespaceMap.md
goneall: Clarified by the correction above - locally, not in memory

model/Core/Classes/NamespaceMap.md
goneall: Text updated

model/Core/Classes/X-Collection.md
```diff
@@ -4,14 +4,22 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 ## Description
 
-An X collection Element is a collection of Elements that does not contain any other X Collection Elements.
+The X-Collection provides a convenient way to store information about SPDX data which has been serialized as a complete
+unit (e.g., all SPDX data within a single JSON-LD file).
+Information we wish to preserve about the serialization itself are store as properties of this class.
 
-In this way it can be thought of as an outer shell collection of SPDX content without 
-self-recursion that can be used as a content aggregation target for serialization.
+NOTE: This class is not intended to be serialized within the SPDX data it is representing.
```
goneall, line 15:
```suggestion
NOTE: This class is not intended to be serialized within the SPDX data it is representing.
```
NOTE: This class is not intended to be serialized within the SPDX data it is representing.  It can, however, be serialized in a subsequent serialization to communicate a prior serialization namespace.

goneall:
This is a reply to @davaya [comment](https://github.com/spdx/spdx-3-model/issues/478#issuecomment-1724327367) in issue #478 - putting it here since it is more related to the namespaceMap discussion than the "where do we start" issue.

> So if namespaceMap can be serialized using JSON-LD context, can't rootElement also be serialized that way, without adding an artificial wrapper element to the graph?

I don't know of anything in the native serialization JSON-LD format that can serve as the rootElement.

If we think of the "X-Collection" as the creators expression of what a serialized blob of data is about, it doesn't feel as artificial.  The challenge only comes in when the "X-Collection" conflicts with something supported by the native serialization - such as prefixes we use in the namespaceMap.

What if we think about it this way:
- The "X-Collection" is a class which captures important information about a serialized blob of SPDX data
- When the creator of an "X-Collection" serializes the "X-Collection" itself, any fields which are already defined in the serialization format being used MUST use the native serialization of that data.  For example, if a serialization format has a mapping of prefixes, that native serialization mapping of prefixes MUST be used and the "X-Collection" MUST NOT contain any (potentially conflicting and redundant) prefix definitions.
- In the SPDX specification, we document for each supported serialization format which "X-Collection" properties must use the native format.  For example, we would have a JSON-LD section that would state the namespaceMap for the "X-Collection" must use the context prefixes.  If there are other "X-Collection" fields (such as rootElement) that have a native serialization equivalent, we would document that usage as well.  In other words, what we came up with for "Solution-B" would apply to any properties of the "X-Collection" that would potentially conflict.
- Any "X-Collection" properties not conflicting with the native serialization formats would be serialized as properties of the "X-Collection" just like any other class.
- The "X-Collection" would be "minted" by the creator of the serialized blob of SPDX data and be given an ID which is unique and represents all the associated properties of the "X-Collection" whether they were serialized in the "X-Collection" or if it used the native serialization as documented in the serialization portion of the SPDX spec.
- When deserializing the "X-Collection", the same serialization rules apply to how the "X-Collection" is reconstituted - for example, you would construct the namespace map from the JSON-LD context.
- If you reserialize the same "X-Collection" in a different format, you may end up serializing a different set of properties.  Since we are not changing the "X-Collection" we would use the same "X-Collection" ID.  We don't need to mint a different ID since it really is the same object, just serialized in a different way.  This solves @maxhbr concern on having to "re-mint" the "X-Collection"

davaya:
@goneall: I like Bob Martin's terminology: Solution A defines a `SerializableCollection`, meaning a specific set of elements before that one set is serialized.  In that vein, Solution B's `SerializedCollection` would be the identical set of elements after being serialized.  Obviously we want to put everything that applies to all formats in SerializableCollection, this would include "where do we start", but would exclude everything that applies only to one serialized instance of that collection, including locationHint and verifiedBy.

But SerializableCollection is not a special X-Collection or Y-Collection created just for JSON-LD, it should be a normal part of the model - an Sbom element if the payload is just an Sbom, or a Bundle if the collection includes multiple Sboms or a colllection of Actors or of Licenses.  Since the wrapper X-Collection problems arose in connection with NamespaceMap, I still suggest removing NamespaceMap from every element (no property in the model has NamespaceMap as its Range) and every serialization spec is required to support it.  This supports every use case we identified for NamespaceMap including the lowest priority: provenance.  There is no "hop to hop" or "latest received" namespace map, there is always just the map used by the producer of every serialized SPDX instance, and that map can be re-applied to every other serialized format of that specific element set.

The serialized instance doesn't need to have a Bundle, or a Bundle wrapped in an X-Collection, it can be just a bunch of Licenses with no wrapper at all if there's no reason to permanently name in the Element Store a group of 5 licenses as being different from any other licenses.  If there is a reason to name that group of 5, then it can be a Bundle, but still doesn't need an X-Collection wrapping the Bundle.

davaya:
@goneall - You're the facilitator, but I don't think separate issues (like NamepaceMap, or Where we start) can be decided independently - the solutions interact holistically.  Bob provisionally accepted Solution B as a way to make progress toward RC2, and Solution A as refined by Bob to a specific collection of elements, not intent applicable future collections of elements by the same or different producers, was a critical breakthrough in discovering how to apply Solution A to #478.

davaya:
PR #500 includes support for both where to start (rootElement) and moving from element-level to document-level dataLicense.

PR #498, opened 2023-09-20
Implement integrity methods for packages

### LINKED ISSUE #345: Package Verification Code
In the current SPDX 3 model, we use the `verifiedUsing` property to verify Elements.  We have had past discussions that we would subclass `IntegrityMethod` to implement the [Package Verification Code](https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field) or replace it with some similar algorithm.  This has yet to be implemented.

In the tools implementer call on 17 May 2023 we discussed this and there was general consensus that the Package Verification Code as it is currently defined and used does not work.  This consensus was based on previous "DocFest's" where we compared calculated Package Verification Codes produced from different tools with the same target package - they all were different making the code less than useful.

We have a few options:
1. Don't provide a verification mechanism for Packages
2. Implement the Package Verification Code as discussed
3. Implement an alternative to the Package Verification Code which can be used in 3.0

With the deadline for SPDX 3 quickly approaching, option one may be the best as long as we don't have adoption issues due to this missing functionality.  Something we should check with the user community before implementing.  We can always adopt 2 or 3 in SPDX release 3.1 since the addition will not be a breaking change unless we make it required.

goneall:
Fixes #345

### DIFF
--- a/model/Software/Classes/PackageGitoid.md
+++ b/model/Software/Classes/PackageGitoid.md
@@ -1,0 +1,23 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageGitoid
+
+## Summary
+
+Preferred verification method to be used for a Package which represents an artifact that logically binds a number of single files together.
+
+## Description
+
+https://www.iana.org/assignments/uri-schemes/prov/gitoid Gitoid stands for [Git Object ID](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) and a gitoid of type blob is a unique hash of a binary artifact. 
+
+Note that in addition to being used as a verification method, gitoid is also a reccomended ExternalIdentifier to be used for the package with the ExternalIdentifierType gitoid.
+
+## Metadata
+
+- name: PackageGitoid
+- SubclassOf: /Software/PackageVerification
+
+## Properties
+
+- gitoid
+  - type: xsd:string
--- a/model/Software/Classes/PackageGitoid.md
+++ b/model/Software/Classes/PackageGitoid.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Preferred verification method to be used for a Package which represents an artifact that logically binds a number of single files together.
+Verification method to be used for a Package which represents an artifact that logically binds a number of single files together using a Git Object ID (Gitoid).
 
 ## Description
 
--- a/model/Software/Classes/PackageHash.md
+++ b/model/Software/Classes/PackageHash.md
@@ -1,0 +1,30 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageHash
+
+## Summary
+
+A verification method for software packages represented as a single blob of bytes, such as a tar archive or single executable.
+
+## Description
+
+The PackageHash is a hash of the single blob of bytes representing a software package.
+The packageDownloadLocation should be a download locator that retrieves the package blob.  
+A supplier can define a PackageHash in a Package without providing a PackageDownloadLocation. 
+This lets consumers perform an offline verification of private blobs.
+
+## Metadata
+
+- name: PackageHash
+- SubclassOf: /Software/PackageVerification
+
+## Properties
+
+- algorithm
+  - type: /Core/HashAlgorithm
+  - minCount: 1
+  - maxCount: 1
+- /core/hashValue
+  - type: xsd:string
+  - minCount: 1
+  - maxCount: 1
--- a/model/Software/Classes/PackageHash.md
+++ b/model/Software/Classes/PackageHash.md
@@ -20,7 +20,7 @@ This lets consumers perform an offline verification of private blobs.
 
 ## Properties
 
-- algorithm
+- /Core/algorithm
   - type: /Core/HashAlgorithm
   - minCount: 1
   - maxCount: 1
--- a/model/Software/Classes/PackageVerification.md
+++ b/model/Software/Classes/PackageVerification.md
@@ -1,0 +1,25 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageVerification
+
+## Summary
+
+An integrity method which supports verification of software packages.
+
+## Description
+
+This is an abstract class representing integrity methods applicable for verification of software packages.
+
+Several use cases for SPDX depend on the consumer being able to verify the provenance and integrity of their software. 
+SPDX can support several different scenarios depending on what information is available to the producer, what information is available to the consumer, and how the SPDX document is delivered. 
+These scenarios are described below along with recommended PackageVerification subclass used to verifying the SPDX packages.
+
+- A Package can be represented as a single blob of bytes, such as a tar archive or single executable: subclass PackageHash.  Note: PackageDownloadLocation should be a download locator that retrieves the package blob.  A supplier can define a PackageHash in a Package without providing a PackageDownloadLocation. This lets consumers perform an offline verification of private blobs.
+- A Package represents an artifact that logically binds a number of single files together : subclass PackageGitoid. Gitoid stands for Git Object ID and a gitoid of type blob is a unique hash of a binary artifact. 
+- Package meta data was imported from an SPDX 2.X document: subclass PackageVerificationCode.  The PackageVerificationCode is the algorithm used in SPDX versions prior to 3.0.  The PackageVerificationCode subclass is provided for compatibility.  Note that the PackageGitoid subclass is preferred if compatibility with SPDX version 2.X is not required.
+
+## Metadata
+
+- name: PackageVerification
+- SubclassOf: /Core/IntegrityMethod
+- Instantiability: Abstract
--- a/model/Software/Classes/PackageVerificationCode.md
+++ b/model/Software/Classes/PackageVerificationCode.md
@@ -44,7 +44,7 @@ Required sort order: '0','1','2','3','4','5','6','7','8','9','a','b','c','d','e'
 
 ## Properties
 
-- /core/hashValue
+- /Core/hashValue
   - type: xsd:string
   - minCount: 1
   - maxCount: 1
--- a/model/Software/Classes/PackageVerificationCode.md
+++ b/model/Software/Classes/PackageVerificationCode.md
@@ -8,8 +8,7 @@ An SPDX version 2.X compatible verification method for software packages.
 
 ## Description
 
-This verification method is provided for compatibility with SPDX 2.X.  
-The PackageGitoid is the reccomended verification method for software packages not requiring compatibility with SPDX 2.X.
+This verification method is provided for compatibility with SPDX 2.X.
 
 This verification method provides an independently reproducible mechanism identifying specific contents of a package based on the actual files (except the SPDX document itself, if it is included in the package) that make up each package and that correlates to the data in this SPDX document. 
 This identifier enables a recipient to determine if any file in the original package (that the analysis was done on) has been changed and permits inclusion of an SPDX document as part of a package.
--- a/model/Software/Classes/PackageVerificationCode.md
+++ b/model/Software/Classes/PackageVerificationCode.md
@@ -1,0 +1,53 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageVerificationCode
+
+## Summary
+
+An SPDX version 2.X compatible verification method for software packages.
+
+## Description
+
+This verification method is provided for compatibility with SPDX 2.X.  
+The PackageGitoid is the reccomended verification method for software packages not requiring compatibility with SPDX 2.X.
+
+This verification method provides an independently reproducible mechanism identifying specific contents of a package based on the actual files (except the SPDX document itself, if it is included in the package) that make up each package and that correlates to the data in this SPDX document. 
+This identifier enables a recipient to determine if any file in the original package (that the analysis was done on) has been changed and permits inclusion of an SPDX document as part of a package.
+
+Algorithm:
+
+verificationcode = 0
+
+filelist = templist = ""
+
+for all files in the package {
+
+    if file is an "excludes" file, skip it /* exclude SPDX analysis file(s) */
+
+        append templist with "SHA1(file)/n"
+        
+    }
+    
+sort templist in ascending order by SHA1 value
+
+filelist = templist with "/n"s removed. /* ordered sequence of SHA1 values with no separators */
+
+verificationcode = SHA1(filelist)
+
+Where SHA1(file) applies a SHA1 algorithm on the contents of file and returns the result in lowercase hexadecimal digits.
+
+Required sort order: '0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f' (ASCII order)
+
+## Metadata
+
+- name: PackageVerificationCode
+- SubclassOf: /Software/PackageVerification
+
+## Properties
+
+- /core/hashValue
+  - type: xsd:string
+  - minCount: 1
+  - maxCount: 1
+- packageVerificationCodeExcludedFile
+  - type: xsd:string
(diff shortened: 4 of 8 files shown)


goneall:
@zvr - the formattings a bit messed up on the PackageVerificationCode file - is there a way to fix the markdown without breaking the spec-parser?

model/Software/Classes/PackageVerification.md
```diff
@@ -0,0 +1,25 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageVerification
```
maxhbr, line 3: Why do we need that layer? Can't we just use `IntegrityMethod` instead. This also preserves the flexibility of instead just using other existing VerificationMethods I think that this approach will limit flexibility.

I also think that this restriction goes in the wrong direction. This prevents the package from using other verification methods but aother Elements like SBOM is not prevented from using `PackageVerification`.

model/Software/Classes/PackageVerificationCode.md
```diff
@@ -0,0 +1,53 @@
...
+# PackageVerificationCode
+
+## Summary
+
+An SPDX version 2.X compatible verification method for software packages.
+
+## Description
+
+This verification method is provided for compatibility with SPDX 2.X.  
+The PackageGitoid is the reccomended verification method for software packages not requiring compatibility with SPDX 2.X.
```
maxhbr, line 12: I think that gitoid is the best option, and I would instead also keep the option open, to just use the hash of the artifact (as probably the easiest option).

(gitoid is fragile regarding no significant changes like changing an executable flag of a file, that is better in the PackageVerificationCode. Also exclusion of files is a relevant feature, which might make gitoid unusable in some cases)

model/Software/Properties/gitoid.md
```diff
@@ -0,0 +1,17 @@
...
+
+# gitoid
+
+## Summary
+
+Git Object ID following the IANA gitoid URI scheme.
+
+## Description
+
+https://www.iana.org/assignments/uri-schemes/prov/gitoid Gitoid stands for [Git Object ID](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) and a gitoid of type blob is a unique hash of a binary artifact.
```
maxhbr, line 11: > ... of type blob is a unique hash of a binary artifact.

we also need to accept `tree` here, right?

maxhbr: For the `tree` variant, we might need to define a way, which files were included and relative to which root is was computed. Otherwise it is hard to reproduce. Especially if the SBOM is committed to the source tree.

model/Software/Classes/PackageGitoid.md
```diff
@@ -0,0 +1,23 @@
...
+
+# PackageGitoid
+
+## Summary
+
+Preferred verification method to be used for a Package which represents an artifact that logically binds a number of single files together.
+
+## Description
+
+https://www.iana.org/assignments/uri-schemes/prov/gitoid Gitoid stands for [Git Object ID](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) and a gitoid of type blob is a unique hash of a binary artifact.
```
maxhbr, line 11: This sentence appears twice. Can we drop redundant descriptions?

model/Software/Classes/PackageVerification.md
maxhbr: If we would drop that layer in the class diagram, we would also not need to introduce `PackageHash` which is basically just a `Hash`

maxhbr (commented):
I think that `PackageVerification` adds another layer which adds additional complexity and does not accomplish its goal

model/Software/Classes/PackageVerification.md
goneall: > If we would drop that layer in the class diagram, we would also not need to introduce PackageHash which is basically just a Hash

I think we need `PackageHash` such that we can document the hash refers to the distributed document.

goneall: > Why do we need that layer?

My thinking was we could have a restriction that the package requires an `IntegrityMethod` of type `PackageVerification`.  It also is a way of implementing a proxy for the enumeration types @jeff-schutt proposed.

I personally think it makes the spec easier to read by grouping the package integrity methods together, but I don't feel strongly about this.

model/Software/Classes/PackageVerificationCode.md
goneall: @maxhbr - Can you propose some specific wording to represent your proposed approach?

model/Software/Classes/PackageVerification.md
maxhbr: so it is semantically different to `/Core/Element/verifiedUsing`?


- `/Core/Element/verifiedUsing` is the method to verify the real world object that corrensponds to the Element
- `this` is used to hash the distributed document (you mean artifact?).

model/Software/Properties/gitoid.md
goneall: > For the tree variant, we might need to define a way, which files were included and relative to which root is was computed. Otherwise it is hard to reproduce. Especially if the SBOM is committed to the source tree.

Sounds like we're re-inventing the `PackageVerificationCode`.  Perhaps if there are excluded files, we recommend the `PackageVerificationCode`

goneall: > we also need to accept tree here, right?

Yes - @maxhbr  can you propose some specific language?

model/Software/Classes/PackageVerificationCode.md
maxhbr: just drop that line? The line above is sufficient

model/Software/Classes/PackageVerification.md
maxhbr: > My thinking was we could have a restriction that the package requires an IntegrityMethod of type PackageVerification. 

wouldn't it be more important or helpful to say that non-Packages are not allowed to have an IntegrityMethod of type PackageVerification?

model/Software/Classes/PackageGitoid.md
goneall: I don't see where the sentence appears twice - can you reference the other sentence?

model/Software/Classes/PackageVerification.md
maxhbr: (lets move one of the conversations to https://github.com/spdx/spdx-3-model/pull/498/files/8477afa0b39440a85059c39ad5d6e7dd1c718e95#r1332675795, to stop the interleaved threads)

model/Software/Classes/PackageGitoid.md
maxhbr: https://github.com/spdx/spdx-3-model/pull/498/files#diff-0d379f36e4e378c06fa5c274a23c3c113fbf6190b0ca6bb62a4b21782a54aef0R11

model/Software/Properties/gitoid.md
maxhbr:
```suggestion
https://www.iana.org/assignments/uri-schemes/prov/gitoid Gitoid stands for [Git Object ID](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) and a gitoid is a unique hash of a binary artifact. 
```

model/Software/Classes/PackageGitoid.md
goneall: Since the second reference is in the property name description, I think the redundancy is fine.

model/Software/Classes/PackageVerificationCode.md
goneall: Done

model/Software/Classes/PackageVerification.md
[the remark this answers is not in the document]
goneall: @maxhbr - If you could propose some specific text that implements your proposal - I would like to close this and the related issue.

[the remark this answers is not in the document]
maxhbr: I would suggest to drop the file `PackageHash.md`. And to make `PackageGitoid` and `PackageVerificationCode` a subclass of `Hash`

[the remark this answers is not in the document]
goneall: @maxhbr - I'm about to head out on travel without email - do you want to open a new PR with your proposal and I'll close this one?

model/Software/Classes/PackageGitoid.md
```diff
@@ -0,0 +1,23 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageGitoid
+
+## Summary
+
+Preferred verification method to be used for a Package which represents an artifact that logically binds a number of single files together.
```
zvr, line 7: Why "Preferred" ?

model/Software/Classes/PackageHash.md
```diff
@@ -0,0 +1,30 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PackageHash
```
zvr, line 3: This looks *very* much like /Core/Hash -- I mean, it even has the exact same properties. Why is it here?

If there is really need for a Hash as a verification, there should be a Class with a single property, of type `/Core/Hash`.

model/Software/Classes/PackageHash.md
```diff
@@ -0,0 +1,30 @@
...
+This lets consumers perform an offline verification of private blobs.
+
+## Metadata
+
+- name: PackageHash
+- SubclassOf: /Software/PackageVerification
+
+## Properties
+
+- algorithm
```
zvr, line 23: Isn't this `/Core/algorithm` ? Or is the plan to introduce another `algorithm` property in this namespace?

model/Software/Classes/PackageGitoid.md
```diff
@@ -0,0 +1,23 @@
...
+Note that in addition to being used as a verification method, gitoid is also a reccomended ExternalIdentifier to be used for the package with the ExternalIdentifierType gitoid.
+
+## Metadata
+
+- name: PackageGitoid
+- SubclassOf: /Software/PackageVerification
+
+## Properties
+
+- gitoid
```
zvr, line 22: So, the idea is that we have a verification class which is just a hash. Given that `gitoid` are equivalent with `swhid`, shall these be abstracted or shall there be another class with a single property?

model/Software/Classes/PackageHash.md
```diff
@@ -0,0 +1,30 @@
...
+# PackageHash
+
+## Summary
+
+A verification method for software packages represented as a single blob of bytes, such as a tar archive or single executable.
+
+## Description
+
+The PackageHash is a hash of the single blob of bytes representing a software package.
+The packageDownloadLocation should be a download locator that retrieves the package blob.
```
zvr, line 12: Is this referring to `downloadLocation`? Which is optional in `Package`?

model/Software/Classes/PackageVerificationCode.md
```diff
@@ -0,0 +1,52 @@
...
+Required sort order: '0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f' (ASCII order)
+
+## Metadata
+
+- name: PackageVerificationCode
+- SubclassOf: /Software/PackageVerification
+
+## Properties
+
+- /core/hashValue
```
zvr, line 47:
```suggestion
- /Core/hashValue
```

model/Software/Classes/PackageVerificationCode.md
```diff
@@ -0,0 +1,52 @@
...
+  - type: xsd:string
+  - minCount: 1
+  - maxCount: 1
+- packageVerificationCodeExcludedFile
+  - type: xsd:string
```
zvr, line 52: If it's a `xsd:string` and not a `File`, I suggest the name should be `...FileName`.

model/Software/Properties/packageVerificationCodeExcludedFile.md
```diff
@@ -0,0 +1,18 @@
...
+
+# packageVerificationCodeExcludedFile
+
+## Summary
+
+Files which are excluded from the PackageVerificationCode integrity method.
+
+## Description
+
+Files which are excluded from the PackageVerificationCode integrity method.
```
zvr, line 11: There is need to specify what the string represents: basename, full path, starting from, ... ?

zvr (changes requested):
I think the abstractions here have to be re-considered (maybe it's what @maxhbr is also saying). I admit I got confused after a while.

model/Software/Classes/PackageHash.md
maxhbr: Yes, that is what I was saying. Get rid of PackageHash and just have Hash

model/Software/Classes/PackageVerificationCode.md
maxhbr: this is the same name as in 2.3: `packageVerificationCodeExcludedFile`, see https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field

model/Software/Classes/PackageVerification.md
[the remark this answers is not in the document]
maxhbr: Just found the issue that: if `PackageVerificationCode` extends `Hash` it then needs to set the `hashAlgorithm`. I think it would be correct and OK to set it on `sha1` but that might also be confusing. Not sure what to do

model/Software/Classes/PackageVerificationCode.md
kestewart: Let's keep it the same as 2.3 at this point.

model/Software/Classes/PackageGitoid.md
goneall: Agree - will removed "Preferred"

model/Software/Classes/PackageHash.md
goneall: `PackageHash` is a subclass of `PackageVerification` whereas `Hash` is not.

We are using the type restrictions to restrict what a `PackageVerification` can be.  If we use this approach, then we need a separate class.

If we want to take a different approach, someone needs to describe it in a pull request or some other form of documentation - at this point, I don't see any other reasonable alternatives.

@maxhbr @zvr - It sounds like you don't agree with this approach - can you provide specifics on an alternative that provides a method for verifying packages?  See issue #345 for the alternatives I came up with.  Also note that this PR is the result of a discussion in the tech call where we agreed to go down this path.

model/Software/Classes/PackageHash.md
goneall: Agree - change to /Core/algorithm

model/Software/Classes/PackageGitoid.md
goneall: I agree `swhid` could also be used for package verification.  A consistent approach would be to add another class with a single property.

model/Software/Classes/PackageHash.md
goneall: Yes - note the word "should" rather than "must" as it relates to optional.

model/Software/Classes/PackageVerificationCode.md
goneall: Agree - let's keep it the same

model/Software/Properties/packageVerificationCodeExcludedFile.md
goneall: @zvr - good idea.
I added some text which should resolve this.

goneall:
> I think the abstractions here have to be re-considered (maybe it's what @maxhbr is also saying). I admit I got confused after a while.

I just added a new PR #598 which removes the abstraction layer.  Please take a look let me know if this addresses the issue.  Note: the PR also includes a fix to an unrelated issue of duplicating external identifier information.

goneall:
Based on the package verification meeting (part of the Security profile team) discussion, we decided that the model does support package verification codes as is.  A separate PR will be opened to better document this in the spec.

Closing this PR as it is no longer needed.

PR #500, opened 2023-09-23
Payload and Namespace map

davaya:
The serialization focus group is nearing agreement.  This PR adds payload-related datatype definitions for concreteness but names are still subject to discussion:
* payload - a serialized byte sequence in a specific data format
* Payload - the non-element datatype class that defines payload content
* PhysicalDocument - the element describing a payload (formerly called SerializedElements)


goneall:
Since this is different from the current proposal, I'm going to close this to avoid confusion.

@davaya we can re-open or create a new PR if our evaluation of the external map underway demonstrates significant issues

PR #503, opened 2023-09-29
Initial definition of Extension namespace

sbarnum:


sbarnum:
It looks like the checks on this may have failed due to the inclusion of an "Remote Classes" section within the extension.md property specification.
This is a new concept but is required if we want Extension to be in a separate profile/namespace.
Typically all SHACL NodeShapes (with accompanying property shapes) will be defined directly from the Class specification file.
In this case to avoid having to specify the property shape for 'extension' on the Element class in the core namespace which would invalidate the concept of Extension as a separate profile/namespace, there needs to be an additional SHACL NodeShape targeting the Element class and asserting the property shape (no cardinality constraints; only path name, nodekind and class) for the 'element' property but it must be defined in the Extension profile/namespace.
To accomplish this I introduced the nomenclature of "Remote Classes" as a specification section on some properties that asserts SHACL shapes that need defined within the local profile/namespace for that property on the referenced "remote" (outside the local namespace) Class. While the case of 'element' property on Element is zer0-to-many and thus has no cardinality constraints other potential uses of Remote Classes in the future may.

goneall:
@sbarnum - thanks for writing this up.  On the remote classes, we can open a PR for the [spec-parser](https://github.com/spdx/spec-parser) to understand the section.

Currently, the spec-parser generates a model file with OWL and SHACL and a separate JSON-LD context file.

Can you document what should be added to these files?  We can then update the spec-parser unless there are objections to the format.

goneall:
@sbarnum - could you also add a profile.md for Extension which describes the profile?

zvr:
I'm not sure I understood what "Remote Classes" is about.

Are you saying that this PR introduces the property `extension` (an object property with range of `Extension`) and this property is on the `Core/Element` class?

If yes, the way we have done it until now is to actually modify the `Core/Element` class.

sbarnum:
> I'm not sure I understood what "Remote Classes" is about.
> 
> Are you saying that this PR introduces the property `extension` (an object property with range of `Extension`) and this property is on the `Core/Element` class?
> 
> If yes, the way we have done it until now is to actually modify the `Core/Element` class.

Yes. This PR introduces the property 'extension' (an object property with range of `Extension`) and this property would apply on the `Core/Element` class however it is defined within a separate profile with a separate namespace and therefore cannot be expressed as a modification to the `Core/Element` class specification in core.
If it was defined as a modification to the `Core/Element` class specification in core then you could not separate conformance to the core profile without conformance to the Extension profile from conformance to the core profile with conformance to the Extension profile.
By specifying `Core/Element` as a 'Remote Class' on the `extension` specification within the Extension profile/namespace it can be used to generate a separate SHACL shape in the separate SHACL file for the Extension profile/namespace and only be applied when the content claims conformance to the Extension profile.

sbarnum:
> @sbarnum - could you also add a profile.md for Extension which describes the profile?

I am unclear where this would go and what i should use as an exemplar of what you are looking for.
Could you outline a little more what you are looking for?

goneall:
@zvr and @sbarnum to sync up and resolve

model/Core/Classes/Element.md
```diff
@@ -52,6 +52,6 @@ and inter-relatable content objects.
   - type: ExternalIdentifier
   - minCount: 0
 - extension
-  - type: Extension
+  - type: Extension/Extension
```
zvr, line 55:
```suggestion
  - type: /Extension/Extension
```

model/Core/Properties/extension.md
```diff
@@ -3,14 +3,14 @@ SPDX-License-Identifier: Community-Spec-1.0
...
+`extension` specifies an Extension-based characterization of a particular aspect of an Element.
 
-TODO
 
 ## Metadata
-
 - name: extension
-
+- Nature: ObjectProperty
+- Range: Extension
```
zvr, line 16:
```suggestion
- Range: /Extension/Extension
```

zvr (changes requested):
Minor fixes to the way we reference other namespaces, and then it can be merged.

zvr (approved):

goneall:
It looks like the spec-parser is failing with this PR.  I'll see if I can narrow down the issue.

goneall:
It looks like this is caused by having a profile by the same name as a property.

I created https://github.com/spdx/spec-parser/issues/82 in the spec parser.

@zvr - any ideas how to fix?

We could work around this by either renaming the property or renaming the profile.

zvr:
@goneall the new parser, completely rewritten from scratch, does not have such issues.

goneall:
> the new parser, completely rewritten from scratch, does not have such issues.

Where is the new parser implemented?

zvr:
> Where is the new parser implemented?

https://github.com/spdx/spec-parser/tree/new-parser

goneall:
@zvr - how soon before the new spec-parser is ready?  I noticed it isn't processing args yet - so I wasn't able to test with this branch of the model.

kestewart:
per Jeff on the call,  the new spec parser issues should now be resolved.

goneall (approved):
LGTM

Issue #463, opened 2023-08-01
Adding relationship restrictions


goneall:
With the decision to make `declaredLicense` and `concludedLicense` relationships rather than properties, we lost a mechanism to require the presence of this information.

It has been suggested that we could create a SHACL shape to make certain relationships required for certain profiles.

These need to be specified in our markdown file format and implemented in the spec parser.

Note that this is required for both the licensing related profiles and the the Lite profile.

maxhbr:
Would the profile be applied to the artifact, which as a result would need to have a corresponding relationship? Or would it be applied to an collection which then needs to contain for every artifact also the necessary relationships to encode license information?


Will there also be a "exactly one" restriction, or is it "at least one"?

maxhbr:
as mentioned in https://github.com/spdx/spdx-3-model/issues/449 and https://github.com/spdx/spdx-3-model/pull/293: the restriction does not only need to be, that there is such a relationship, but it must also restrict the `to` side to contain exactly one or at least one license entry.

goneall:
@maxhbr - all good questions.  

I would test our solution against the `concludedLicense` and `declaredLicense` restrictions for a Licensing profile.

I believe the security profile has similar relationship restriction requirements.

goneall:
Note proposal in https://github.com/spdx/spdx-3-model/issues/462#issuecomment-1714744171

henkbirkholz:
@mahbr references https://github.com/spdx/spdx-3-model/issues/449 and I hope it is okay to ask for clarification before this issue is resolved.

In https://github.com/spdx/spdx-3-model/issues/449, @davaya highlighted that `VulnAssessmentRelationship` in Security puts restrictions on the `to` field. Analogously, it is highlighted that`LifecycleScopedRelationship` and `SoftwareDependencyRelationship` in Software do not have any restrictions. (Or had, as https://github.com/spdx/spdx-3-model/issues/449 is from July - I did not check).

Is that issue actually related to this one? If so, is that okay or is that an issue and restrictions should be added?

kestewart:
Closing this as I believe it's been resolved in https://github.com/spdx/spdx-3-model/pull/524.  Please reopen it if there are issues.

PR #507, opened 2023-10-03
Add an external relationship restriction to Licensing profile

goneall:
This PR adds a concluded license restriction for the Licensing profile.  Note that Issue #463 needs to be fixed before this is merged.


goneall:
This is obsoleted #524

PR #508, opened 2023-10-03
Replace describes relationship with rootElement property

### LINKED ISSUE #495: documentDescribes relationship overlaps with rootElement
Also in the diagram, it states it is from a "bundle" to an Artifact which is inconsistent with the description.

The description of documentDescribes:

`Every to Element is described by the from Element. This can be used to denote the root(s) of a tree of elements contained in an SBOM.`

The description of rootElement:

`A rootElement of a collection is the top level Element from which all other Elements are reached via relationships.`

goneall:
Fixes #495 

This is a proposed solution based on the discussion in the tech call on 3 October, 2023.

### DIFF
--- a/model/Core/Properties/rootElement.md
+++ b/model/Core/Properties/rootElement.md
@@ -4,11 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Top level Element from which all other Elements are reached via relationships.
+This property is used to denote the root Element(s) of a tree of elements contained in an SBOM.
 
 ## Description
 
-A rootElement of a collection is the top level Element from which all other Elements are reached via relationships.
+This property is used to denote the root Element(s) of a tree of elements contained in an SBOM.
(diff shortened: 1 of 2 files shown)


kestewart (changes requested):
There are more use cases than SBOMs, and we're using describes relationship for more than the root element.     If we're doing this for SBOM's I'm fine with adding a property to the SBOM document, but I don't want to loose the capability to handle other use cases with the describes relationship.

goneall:
> There are more use cases than SBOMs, and we're using describes relationship for more than the root element. If we're doing this for SBOM's I'm fine with adding a property to the SBOM document, but I don't want to loose the capability to handle other use cases with the describes relationship.

@kestewart - Updated the PR to add back a describes relationship type, but documented the property should be used for root elements.

goneall:
@kestewart This PR now retains the describes relationship consistent with what we discussed on the tech call 31 Oct 2023. Please review.

kestewart (approved):
Thanks.  That works for me.

PR #509, opened 2023-10-05
Document 'XCollection' serialization specifics

goneall:


goneall:
Notes that in the proposed writeup `"XCollection" will be in the serialized data with ONLY the properties not supported by the serialization format.` which implies that all other "XCollection" properties would be serialized.

Since "XCollection" is a subclass of `ElementCollection`, the following required properties would always be serialized:
- `SPDXID`
- `CreationInfo`

There are also a number of other optional properties like `name` and `comment` which are inherited from `Element`.

I can see quite a few benefits to having these present in the serialized data, so I personally don't see any issue - but I thought I would point this out to make sure we're on the same page on how the "XCollection" is serialized.

davaya:
* We decided that every serialization format must support NamespaceMap.  That's easy to see for data formats like XML and JSON, but it's also easy for formats like spreadsheets - just envision it as a "Legend" at the top of the spreadsheet, like a chart legend maps colors to the characteristics represented by those colors.

The serialization group is discussing the difference between:
    * a byte sequence ("Document") containing a specific set of elements in a specific data format
    * an element (SerializedCollection) that describes a byte sequence in a specific data format
    * an element list (XCollection) that may be included in multiple byte sequences

For Documents that do not include an XCollection element (individual elements, and multiple elements that may be roots),
    * SerializedCollection is an element with `SpdxId` and `CreationInfo` properties, and can have a `dataLicense` property

For Documents that include neither SerializedCollection nor XCollection elements,
    * the byte sequence ("Document") can be required to support a `dataLicense` property in every serialization format

It's desirable not to clutter serialization specs with too much non-Element data, but if `dataLicense` must be supported on single-element Documents, then the only possibility is to define it as non-Element data.  SerializedCollection can then capture the non-Element data present in the specific Document it describes.

davaya:
This PR describes X-Collection / SerializableCollection / LogicalDocument.

PR #500 describes SerializedElements / PhysicalDocument.  The class is now called PhysicalDocument for clarity, and it is a subclass of Artifact as agreed in the 10/16 serialization meeting.

davaya:
This PR describes X-Collection / SerializableCollection / LogicalDocument.

PR #500 describes SerializedElements / PhysicalDocument.  The class is now called PhysicalDocument for clarity, and it is a subclass of Artifact as agreed in the 10/16 serialization meeting.

PR #500 also includes the Payload non-element datatype class that defines serialized data.  If there are "quite a few benefits to having these (`name`, `comment`) present in the serialized data", they can be added to the Payload class.  But for now it includes only NamespaceMap and CreationInfoMap properties necessary to compact and beautify the serialized data.  I'd recommend that `name`, `comment`, etc be included in the PhysicalDocument element that refers to serialized data rather than directly in the Payload data itself.

serialization/README.md
```diff
@@ -2,13 +2,36 @@
 
 ## Serialization info
 
-_TBD_
+This directory contains the specifications and examples for all supported SPDX formats.
```
sbarnum, line 5: Suggest this be slightly tweaked to say "This directory contains the specifications and examples for all supported SPDX serialization formats."

serialization/README.md
```diff
@@ -2,13 +2,36 @@
...
+
+There is a specification markdown file for each supported format describing the serialization.
+
+A collection of elements can be serialized in multiple formats.
+
+Within the model, we have an "XCollection" which represents the common properties of the collection of elements across all data formats.
+
+We can refer to the actual serialized bytes as and Artifact in the model.
```
sbarnum, line 13: Suggest slight tweak to "We can represent/characterize the actual serialized bytes using an Artifact element in the model."

serialization/README.md
```diff
@@ -2,13 +2,36 @@
...
+
+A Relationship of type "XCollectionOf" can be used to relate an "XCollection" to one more serialized for of the "XCollection".
+
+Within serialized data, you may need to refer to properties in the "XCollection" - such as "where to start" in a collection.
+
+In that case, the "XCollection" will be in the serialized data with ONLY the properties not supported by the serialization format.
+
+There is at most one "XCollection" in the serialization.
```
sbarnum, line 21: Suggest slight tweak:

"There is at most one "XCollection" element defined in any given instance of serialization."

There can be any number of references via id but only a single XCollection defined within a given instance of serialization.

serialization/README.md
```diff
@@ -2,13 +2,36 @@
...
+
+The serialized "XCollection" shall not contain properties where that property value can be derived from the the serialized data itself.
```
sbarnum, line 23: I suggest removing this line. 
I believe that the following line covers the situation adequately:
""When serializing the physical “XCollection”, any properties of the logical element which can be represented using native mechanisms in the specific serialization format (e.g. @context prefixes in JSON-LD in place of the namespaceMap) should use these mechanisms and the remaining properties should be serialized within the “XCollection” element itself.""

This line attempting to assert the inverse has potential for misunderstanding or misuse asserting that because a consumer could "figure it out" via their own analysis some useful data (e.g., where to start) should not be provided by the producer to the consumer to make it easier for them.

serialization/json-ld.md
```diff
@@ -0,0 +1,41 @@
...
+The namespaceMap uses the [term to IRI mapping](https://www.w3.org/TR/json-ld11/#example-11-term-expansion-from-context-definition) in the [JSON-LD context](https://www.w3.org/TR/json-ld11/#the-context).
+
+### element
+
+The [graph objects](https://www.w3.org/TR/json-ld11/#graph-objects) `@graph` lists the elements for the XCollection.
+
+The RDF graph of an instance of the SPDX model shall contain all Element nodes (i.e. objects that are subclasses of Element) as a list on top-level under the "@graph" key.
+This means that all references to Element nodes have to use the URI of the referenced Element.
+Inlining/Embedding of Element nodes into other nodes is not allowed.
+On the other hand, non-Element nodes (like those of type "ExternalReference" or similar complex data classes) have to be inlined (TODO: we may want to make an exception for CreationInfo, depending on the outcome of the surrounding discussion).
```
sbarnum, line 23: My understanding is that the consensus result regarding serialized compaction for CreationInfo is that WHEN SERIALIZED each instance of CreationInfo containing different data would be serialized (with its @type and @id) as an object within the '@graph' JSON-LD structure and referenced by any Elements with that same CreationInfo. When deserialized each serialized CreationInfo MUST be expanded back out to all of the elements it applies to such that each element MUST include its own CreationInfo (with separate @id from any other elements CreationInfo).

sbarnum, line 23: It also needs calling out explicitly here that each embedded non-Element node MUST have an explicit id. This is true for ALL serialization formats. 
The serialized embedding within the json-ld object structure is a method of compaction to make the content easier to read (and process by some parsers) but it does not inherently represent the native deserialized logical content which is a graph where EVERY class instance (Element nodes and non-Element nodes) are separate nodes in the graph connected using the globally unique id of each node which MUST be present on all nodes.

There is a possible serialization compaction approach where we COULD (though I would not recommend it) assert that in situations where a producer is minting new elements to be directly serialized without creation anywhere internally first which would require ids already be assigned they could use blank node ids (like has been discussed for CreationInfo) within a specific serialization instance. These blank nodes would then be REQUIRED to be skolemized to create full globally unique ids (like done for CreationInfo) upon any deserialization. This could potentially save some space within the serialized content but it would likely be minimal and the extra complexity in handling and potential for misinterpretation makes this approach a questionable option to pursue.

serialization/json-ld.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Context File
+
+The SPDX organization provides a context file that is to be used universally for all SPDX JSON-LD files of a given SPDX version.
```
sbarnum, line 27: Suggested small tweak:

"The SPDX organization provides a global JSON-LD @context file that MUST be used universally for all SPDX JSON-LD files of a given SPDX version."

serialization/json-ld.md
```diff
@@ -0,0 +1,41 @@
+# JSON-LD
+JSON-LD is a JSON-based format to encode RDF graphs. Its documentation can be found [here](https://www.w3.org/TR/json-ld11/).
+
+The JSON-LD is an RDF format and follows the serialization rules of the [rdf SPDX serialization specification](rdf.md).
+
+## XCollection
+
+The following XCollection properties are mapped to JSON-LD syntax specifications.
```
sbarnum, line 8: Suggest small tweak:

"The following XCollection properties are mapped to native JSON-LD mechanisms defined within the JSON-LD syntax specifications."

serialization/json-ld.md
```diff
@@ -0,0 +1,41 @@
...
+Any properties not listed below should be serialized as part of the XCollection element itself within
+the JSON-LD serialized data.
+
```
sbarnum, line 11: Needs new additional line(s)after line 10:

"Deserialization of any JSON-LD serialized SPDX content MUST expand the inverse of these native mappings such that the logical "XCollection" element directly contains its full set of properties."

serialization/README.md
maxhbr: @sbarnum : you can use the `suggestion` feature to make your suggestion easier to read. It is then rendered as:

```suggestion
This directory contains the specifications and examples for all supported SPDX serialization formats.
```

serialization/json-ld.md
goneall: @sbarnum - Let's separate this out into a separate issue we can tackle after merging this.  I did not intend to change the treatment of CreationInfo, just copying what was there before in the rdf.md file.  I also have some concerns in how it is currently worded.

goneall: Same as above - i just copy/pasted what was there before.

serialization/README.md
goneall: Updated per suggestion

serialization/README.md
```diff
@@ -2,13 +2,36 @@
...
 ## Serialization info
 
-_TBD_
+This directory contains the specifications and examples for all supported SPDX formats.
+
+There is a specification markdown file for each supported format describing the serialization.
+
+A collection of elements can be serialized in multiple formats.
+
+Within the model, we have an "XCollection" which represents the common properties of the collection of elements across all data formats.
```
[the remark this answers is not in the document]
goneall: Updated per suggestion

serialization/README.md
```diff
@@ -2,13 +2,36 @@
...
+
+We can refer to the actual serialized bytes as and Artifact in the model.
+
+A Relationship of type "XCollectionOf" can be used to relate an "XCollection" to one more serialized for of the "XCollection".
+
+Within serialized data, you may need to refer to properties in the "XCollection" - such as "where to start" in a collection.
```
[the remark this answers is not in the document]
goneall: Updated per suggestion

serialization/README.md
goneall: Updated per suggestion

serialization/json-ld.md
goneall: Updated per suggestion

serialization/json-ld.md
goneall: Updated per suggestion

sbarnum (approved):
This PR looks good to me.
It accurately covers the consensus results of serialization WG discussions on these topics and aligns with the rest of the model.

There is one issue identified in comments regarding like 23 of the json-ld.md file that needs to be raised and addressed to clarify serialization rules for non-Element classes.

PR #512, opened 2023-10-05
update severity.md

jeff-schutt:
These updates address changes to severity.md in support of https://github.com/spdx/spdx-3-model/pull/501#pullrequestreview-1651152833

CC @goneall @rnjudge @tsteenbe

PR #511, opened 2023-10-05
update CvssV3VulnAssessmentRelationship.md

jeff-schutt:
These updates address changes to CVSSv3VulnAssessmentRelationship.md in support of https://github.com/spdx/spdx-3-model/pull/501#pullrequestreview-1651152833

CC @goneall, @rnjudge, @tsteenbe

PR #510, opened 2023-10-05
remove severity property from CvssV2VulnAssessmentRelationship.md

jeff-schutt:
In working to resolve https://github.com/spdx/spdx-3-model/pull/501#pullrequestreview-1651152833 I noticed that the [CVSS version 2.0 schema](https://www.first.org/cvss/cvss-v2.0.json) does not include a severity property. The changes remove this field from the class.

CC: @goneall @rnjudge @tsteenbe

PR #501, opened 2023-09-28
add CvssV4VulnAssessmentRelationship.md and CvssSeverityType.md

jeff-schutt:
Initial draft of changes to support CVSS version 4

CC: @rnjudge @tsteenbe @goneall

PR #513, opened 2023-10-05
update security vector.md to vectorString.md

jeff-schutt:
With updates to the security profile (#510, #511, #512) to support CVSS version 4 (#501), I noticed that SPDX can better align to the CVSS specifications by naming the vector property the same as is done in CVSS specifications, namely `vectorString` as seen in CVSS schemas [v2.0](https://www.first.org/cvss/cvss-v2.0.json), [v3.0](https://www.first.org/cvss/cvss-v3.0.json), [v3.1](https://www.first.org/cvss/cvss-v3.1.json), and [v4.0](https://www.first.org/cvss/cvss-v4.0.json).

CC @rnjudge @goneall @tsteenbe


goneall (approved):
Agree with the change.  There are probably some references to the old "vector" name that need to be updated.

kestewart (approved):

PR #514, opened 2023-10-06
Update Security EPSS Support

jeff-schutt:
Updates to support and better align SPDX 3 EPSS class and properties to the EPSS data published at https://www.first.org/epss/data_stats

CC @rnjudge @goneall @tsteenbe

### DIFF
--- a/model/Security/Classes/EpssVulnAssessmentRelationship.md
+++ b/model/Security/Classes/EpssVulnAssessmentRelationship.md
@@ -8,27 +8,27 @@ Provides an EPSS assessment for a vulnerability.
 
 ## Description
 
-An EpssVulnAssessmentRelationship relationship describes the likelihood or
-probability that a vulnerability will be exploited in the wild using the Exploit
-Prediction Scoring System (EPSS) as defined on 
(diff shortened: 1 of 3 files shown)


goneall (approved):
LGTM - I verified that `xsd:decimal` is the best fit for a percentage representation between 0 and 1.

jeff-schutt:
Thanks for the second set of eyes and confirmation. When I looked into the best way to represent a percentage I was reminded that xsd:decimal can include any number with decimal places, e.g, 128.425. Is there a constraint that we can use in SPDX to formally limit the decimal value to between 0-1? These commits provide a note in the Description,yet that doesn't appear to be enforceable like a minCount or maxCount.

goneall:
> Is there a constraint that we can use in SPDX to formally limit the decimal value to between 0-1?

I found [this article ](https://www.w3schools.com/xml/schema_facets.asp) on XSD restrictions which may work, but it would require enhancements to the schema parser.

Since it is a bit involved, perhaps we could open a separate issue and add the restriction after RC2.  We could even make the argument it could be added in 3.1 since the documentation states the restriction.

In any case, decimal looks like the right choice for the type.

zvr (approved):

Issue #516, opened 2023-10-16
Add the ability to constrain relationship types to require relationships a specific subclass


goneall:
The build profile team would like to constrain certain relationship types to include the `scope` which is a property in the subclass `LifecycleScopedRelationship`.

goneall:
This looks like it was resolved with PR #517 - closing

Issue #515, opened 2023-10-16
Add the ability to constrain multiple choice relationship types


goneall:
The build profile would like to add a constraint that there must be at least one of several possible relationship types to be considered valid under the build profile.

We already have a proposal to constrain to a single relationship type.

We would need to extend this to allow "one of" and update the spec-parser to output the updated SHACL for validation.

goneall:
We decided for 3.0 we will just document these in text - moving this to 3.1 milesonte

bact:
@ilans SHACL-related and maybe of your interest as well

ilans:
I wasn't aware of this open issue. Thanks @bact !
I actually wrote some notes for myself to inquire about. Here they are:

From [Build Profile](https://spdx.github.io/spdx-spec/v3.0.1/model/Build/Build/):
"All relationships in the Build Profile are scoped to the "build" LifecycleScopeType period.

Possible issues:
- Does this imply that the (currently optional) [scope](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Properties/scope/) property is required when using [LifecycleScopedRelationship](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/LifecycleScopedRelationship/) in an SBOM that has a build [profileConformance](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/ProfileIdentifierType/)?
- Does this imply that any [Relationships](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Relationship/) in an SBOM that has a build [profileConformance](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/ProfileIdentifierType/) is required to be a [LifecycleScopedRelationship](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/LifecycleScopedRelationship/)?
  - If so, what about [LifecycleScopedRelationship](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/LifecycleScopedRelationship/) that aren’t used for documenting build lifecycles (not directly, at least)?

goneall:
> I wasn't aware of this open issue. Thanks [@bact](https://github.com/bact) ! I actually wrote some notes for myself to inquire about. Here they are:
> 
> From [Build Profile](https://spdx.github.io/spdx-spec/v3.0.1/model/Build/Build/): "All relationships in the Build Profile are scoped to the "build" LifecycleScopeType period.
[...]

PR #517, opened 2023-10-17
build: Add LifecycleScopedRelationship constraints

nishakm:


nishakm:
Related PRs:
- https://github.com/spdx/spdx-3-model/issues/515
- https://github.com/spdx/spdx-3-model/issues/516

goneall (approved):
LGTM - Just need to wait for the spec parser to handle the OR's before merging.

nishakm:
I've updated the PR so the restrictions are on the Classes/Build.md file. However, I may not have entered the restrictions according to what spec-parser wants. This is on purpose because the constraints are not agreed upon.
@zvr

kestewart (approved):
Need to have further discussions, but applying this so we can see it all in context, for the relationships.

kestewart:
@nishakm, @zvr - can you look into the parser error, and see if you can sort it so we can get this merged.

zvr:
@kestewart the parser does not understand this new section `## External LifecycleScopedRelationship Constraints` describing validation.

We have not even finalized how such constraints should be expressed -- #522 has to be fleshed out, discussed and resolved for this.
My personal view is that having class names inside the section title is not the way to go.

model/Build/Classes/Build.md
```diff
@@ -57,3 +58,15 @@ Note that buildStart and buildEnd are optional, and may be omitted to simplify c
 - environment
   - type: /Core/DictionaryEntry
   - minCount: 0
+
+## External LifecycleScopedRelationship Constraints
+
+- relationshipType: inputOf OR buildTool OR configOf
+  - toType: /Build/Build
```
goneall, line 65: In the discussion on the relationship type restrictions, we decided to only have a fromType sub-bullet since the toType is typically restricted by the relationshipType across all profiles.  I'm thinking the from type would be `/Core/Artifact`.

goneall (changes requested):
A couple of suggested changes to be compatible with the proposed restrictions syntax.

Note that the proposal hasn't been finalized, so we may end up with another round of updates.

goneall:
@zvr - Need your input on as well

goneall:
Ping @nishakm

zvr:
@nishakm such conformance conditions are to be expressed in plain language in the top-level namespace file.

See #524 for an example.

model/Build/Classes/Build.md
```diff
@@ -57,3 +58,15 @@ Note that buildStart and buildEnd are optional, and may be omitted to simplify c
 - environment
   - type: /Core/DictionaryEntry
   - minCount: 0
+
+## External LifecycleScopedRelationship Constraints
```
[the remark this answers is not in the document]
zvr: This should be written in text, as shown in #524

zvr (changes requested):

kestewart (changes requested):
Need to get the relationships consistent across the profiles.  See #556

nishakm:
@kestewart once #556 is merged, I can update this PR accordingly.

nishakm:
@kestewart Done

kestewart (approved):
Thanks Nisha,  this matches what we discussed last week.

model/Build/Build.md
```diff
@@ -4,13 +4,38 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Metadata
 
 - id: https://rdf.spdx.org/v3/Build
 - name: Build
+
+
+## Profile Conformance
+
+Conformance to the build profile requires one or more instances of the Build class. In addition, there must be at least three instances of type LifecycleScopedRelationship where the "scope" property must be "build".
```
zvr, line 39:
```suggestion
Conformance to the Build profile requires one or more instances of the Build class. In addition, there must be at least three instances `Relationship`s with type `LifecycleScopedRelationship`, where the "scope" property must be "build" and the "from" property must be the Build instance.
```

zvr (changes requested):

zvr (approved):

PR #520, opened 2023-10-20
Fixing typos/syntax errors

zvr:
This fixes various issues that the (new) spec-parser points out.

Among them:
- mismatches between "name" property and file name and title
- vocabularies entries not in lowerCamelCase
- missing sections ("Summary" in most/all cases)
- empty sections ("Properties" in most/all cases)
- ... and probably some others I don't remember

### DIFF
--- a/model/Security/Vocabularies/CvssSeverityType.md
+++ b/model/Security/Vocabularies/CvssSeverityType.md
@@ -16,8 +16,9 @@ CvssSeverityType specifies the CVSS severity type, defined in the CVSS specifica
 
 ## Entries
 
-- CRITICAL: When a CVSS score is between 9.0 - 10.0
-- HIGH: When a CVSS score is between 7.0 - 8.9
-- MEDIUM: When a CVSS score is between 4 - 6.9
-- LOW: When a CVSS score is between 0 - 3.9
-- NONE: When a CVSS score is 0
+- critical: When a CVSS score is between 9.0 - 10.0
+- high: When a CVSS score is between 7.0 - 8.9
+- medium: When a CVSS score is between 4 - 6.9
+- low: When a CVSS score is between 0 - 3.9
+- none: When a CVSS score is 0
+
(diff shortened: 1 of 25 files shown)


model/Security/Vocabularies/CvssSeverityType.md
goneall, line 1: This change is consistent with the naming conventions of SPDX, however, the security profile is attempting to have the resultant JSON Schema match the CVSS schema and this change will make them different.

Since there is a direct mapping, I'm OK with using the SPDX conventions, but @jeff-schutt @tsteenbe @rnjudge should also review

I would move this into a separate PR to give the security team time to review.  If the current state causes the spec-parser to fail, then I would suggest merging it in and adding a new issue to discuss the incompatibility.

goneall (changes requested):
See comment - suggest moving one of the changes to a separate PR unless it causes the spec-parser to fail.

model/Security/Vocabularies/CvssSeverityType.md
zvr: This is about entries for a vocabulary, which are automatically converted to IRIs of individuals in RDF.

The process of generating the JSON Schema already does many transformations, as it's not a 1:1 correspondence. If the requirement is to have them appear as upper-case strings, this is easily done at that stage.

goneall: > The process of generating the JSON Schema already does many transformations, as it's not a 1:1 correspondence. If the requirement is to have them appear as upper-case strings, this is easily done at that stage.

Based on the above, it sounds like we can go ahead and move forward with this PR pending one more review.

I'll leave it up to the security profile group to decide if it is important to maintain uppercase.

goneall (approved):

model/Security/Vocabularies/CvssSeverityType.md
kestewart: ok, let's apply this, and then ask them to do a pass.

kestewart (approved):
Apply these changes, but then let's get the security team to review.

They'll need to update their examples.

PR #524, opened 2023-10-24
Example of conformance in plain language

zvr:
This adds an example `conformance.md` file to a namespace (`Licensing` in this case) to show how we can use plain text to describe conformance restrictions.

Signed-off-by: Alexios Zavras (zvr) <github@zvr.gr>


zvr:
Actually, there is no need for a separate file.

Having a section in the namespace top-level file makes it easier to create and process (and generate results from).

goneall (approved):
LGTM

goneall (approved):
Reviewed on tech call 7 Nov 2023, some concerns expressed but all are OK with merging

PR #535, opened 2023-10-25
Fix capitalization for CVSS severity values

rnjudge:
The `severity` value for the CvssV4VulnAssessmentRelationship should be
uppercased. This commit fixes the current lowercased value and makes it
uppercased. It also uppercases the severity entries in
CvssSeverityType.md for correctness and consistency.


zvr:
The vocabulary for this https://github.com/spdx/spdx-3-model/blob/main/model/Security/Vocabularies/CvssSeverityType.md follows our convention of lowerCamelCase.

But if in JSON there is preference for the shorthand to be UPPERCASE, this can be accommodated.

rnjudge:
@zvr We need to update the CvssSeverityType.md file as well then (I'll do that). Thanks for pointing that out. Uppercase is consistent with CVSS spec/scoring.

model/Security/Vocabularies/CvssSeverityType.md
```diff
@@ -16,9 +16,9 @@ CvssSeverityType specifies the CVSS severity type, defined in the CVSS specifica
 
 ## Entries
 
-- critical: When a CVSS score is between 9.0 - 10.0
-- high: When a CVSS score is between 7.0 - 8.9
-- medium: When a CVSS score is between 4 - 6.9
-- low: When a CVSS score is between 0 - 3.9
-- none: When a CVSS score is 0
+- CRITICAL: When a CVSS score is between 9.0 - 10.0
```
zvr, line 19: No, all our vocabularies are lowerCamelCase, and they get translated to RDF individuals.

zvr (changes requested):

zvr:
@rnjudge you misunderstood me.
The vocabulary and RDF individuals will remain in lowerCamelCase.
But if you want the value in JSON (which is a shorthand) to be UPPERCASE, this can be accommodated in the JSON schema.

Where is CvssV3VulnerabilityAssessment that "is already properly uppercased" ?

maxhbr:
>  But if you want the value in JSON (which is a shorthand) to be UPPERCASE, this can be accommodated in the JSON schema.

the example is JSON-LD and not JSON?

maxhbr:
>  The vocabulary and RDF individuals will remain in lowerCamelCase.

Is that a requirement by RDF? I thought that also `BSD-3-Clause` is an individual and it does not follow this rule.

rnjudge:
> @rnjudge you misunderstood me. The vocabulary and RDF individuals will remain in lowerCamelCase. But if you want the value in JSON (which is a shorthand) to be UPPERCASE, this can be accommodated in the JSON schema.

When a schema defined elsewhere (i.e. https://www.first.org/cvss/cvss-v3.0.json defines an enum with uppercase values for severity) uses a specific string format, shouldn't we follow that convention? Why are we defaulting to lowercase camel case for RDF?


> Where is CvssV3VulnerabilityAssessment that "is already properly uppercased" ?

In the syntax example here: https://github.com/spdx/spdx-3-model/blob/main/model/Security/Classes/CvssV3VulnAssessmentRelationship.md

zvr:
I'm obviously bad at explaining... Let me try once more:

It is our convention to use lowerCamelCase for vocabulary entries -- and we also use UpperCamelCase for classes, namespaces and individuals and lowerCamelCase for properties.
I am extremely reluctant to re-visit this, as we might spend very long time to discuss every relationship type name, or every class or property.

Having said that, this only affects the last part of a IRI like, in this case `http://spdx.org/rdf/v3/Security/CvssSeverityType/medium` (instead of `http://spdx.org/rdf/v3/Security/CvssSeverityType/MEDIUM`).

Without any change, we can have in JSON-LD the shortcut "MEDIUM" be translated to the full IRI, no matter what this is.
Keep in mind, that these are not enum of short strings (words); these are full IRIs.

I am not sure the current JSON-LD context generation does anything about vocabularies, anyway, so this is still to be implemented.

rnjudge:
Ok.. I can back out my changes to CvssSeverityType.md but my initial change should be fine then, since it is only changing the example syntax for a JSON-LD example which you are saying can be shortcut to uppercase (i.e. MEDIUM)

zvr:
Yes, @rnjudge  your first change/commit was fine. I only commented below to let you know about the difference in the vocabulary entries.

As I wrote then:
> But if in JSON there is preference for the shorthand to be UPPERCASE, this can be accommodated.

PR #542, opened 2023-10-27
Update ExternalMap and relationships to reference Artifacts

goneall:
Per the decision in the serialization team, change the type of the definingDocument to Artifact and create a relationshipType to link the definingDocument to the SpdxDocument

### DIFF
--- a/model/Core/Properties/definingDocument.md
+++ b/model/Core/Properties/definingDocument.md
@@ -4,15 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Artifact containing SPDX data which defines one or more SPDX elements.
+Artifact representing a serialization instance of SPDX data containing the definition of a particular Element.
 
 ## Description
 
-A definingDocument property is used to link an Element identifier to an Artifact which contains the definition for the Element.
+A definingDocument property is used to link the Element identifier for an Element defined external to a given SpdxDocument to an Artifact Element representing the SPDX serialization instance which contains the definition for the Element.
 
 ## Metadata
 
 - name: definingDocument
 - Nature: ObjectProperty
 - Range: Artifact
-
--- a/model/Core/Properties/definingDocument.md
+++ b/model/Core/Properties/definingDocument.md
@@ -4,15 +4,15 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-URI for an SPDX document which defines an SPDX element.
+Artifact containing SPDX data which defines one or more SPDX elements.
 
 ## Description
 
-A definingDocument property is used to link an Element identifier to an SpdxDocument which contains the definition for the Element.
+A definingDocument property is used to link an Element identifier to an Artifact which contains the definition for the Element.
 
 ## Metadata
 
 - name: definingDocument
-- Nature: DataProperty
-- Range: xsd:anyURI
+- Nature: ObjectProperty
+- Range: Artifact
 
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -72,7 +72,7 @@ between a Package and a File, between two Packages, or between one SPDXDocument 
 - republishedBy: (Security) Designates the agent that tracked, aggregated, and/or enriched vulnerability details to improve context (i.e. NVD)
 - requirementFor: Every `to` Element is required for the `from` Element
 - runtimeDependency: Every `to` Element is a runtime dependency for the `from` Element
-- spdxDocumentForArtfact: Every `to` Element is an SpdxDocument representing the logical serialization metadata for the `from`  Artifact containing the serliazed data
+- serializedInArtifact: The from SPDXDocument can be found in a serialized form in each to Artifact
 - specificationFor: Every `to` Element is a specification for the `from` Element
 - staticLink: Every `to` Element is statically linked to the `from` Element
 - test: Every `to` Element is a test artifact for the `from` Element
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -72,7 +72,7 @@ between a Package and a File, between two Packages, or between one SPDXDocument 
 - republishedBy: (Security) Designates the agent that tracked, aggregated, and/or enriched vulnerability details to improve context (i.e. NVD)
 - requirementFor: Every `to` Element is required for the `from` Element
 - runtimeDependency: Every `to` Element is a runtime dependency for the `from` Element
-- serializedArtifactFor: Every `to` Element is an artifact serialized from the `from` SpdxDocument Element
+- spdxDocumentForArtfact: Every `to` Element is an SpdxDocument representing the logical serialization metadata for the `from`  Artifact containing the serliazed data
 - specificationFor: Every `to` Element is a specification for the `from` Element
 - staticLink: Every `to` Element is statically linked to the `from` Element
 - test: Every `to` Element is a test artifact for the `from` Element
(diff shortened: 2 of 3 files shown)


model/Core/Properties/definingDocument.md
```diff
@@ -4,15 +4,15 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-URI for an SPDX document which defines an SPDX element.
+Artifact containing SPDX data which defines one or more SPDX elements.
```
sbarnum, line 7: Suggest small tweak:

"Artifact representing a serialization instance of SPDX data containing the definition of a particular Element."

model/Core/Properties/definingDocument.md
```diff
@@ -4,15 +4,15 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Description
 
-A definingDocument property is used to link an Element identifier to an SpdxDocument which contains the definition for the Element.
+A definingDocument property is used to link an Element identifier to an Artifact which contains the definition for the Element.
```
sbarnum, line 11: Suggest small tweak:

"A definingDocument property is used to link the Element identifier for an Element defined external to a given "XCollection" to an Artifact Element representing the SPDX serialization instance which contains the definition for the Element."

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -72,6 +72,7 @@ between a Package and a File, between two Packages, or between one SPDXDocument
 - republishedBy: (Security) Designates the agent that tracked, aggregated, and/or enriched vulnerability details to improve context (i.e. NVD)
 - requirementFor: Every `to` Element is required for the `from` Element
 - runtimeDependency: Every `to` Element is a runtime dependency for the `from` Element
+- serializedArtifactFor: Every `to` Element is an artifact serialized from the `from` SpdxDocument Element
```
sbarnum, line 75: I would strongly recommend this RelationshipType be 'from' an "XCollection" 'to' an Artifact and not the other way around.
A given Artifact is never going to be a serialization of multiple XCollections but a given XCollection will often have multiple serializations represented by different Artifacts.

I also strongly suggest the RelationshipType string "hasSerialization" as it is extremely clear and descriptive in its meaning and unlikely to be misunderstood.

model/Core/Properties/definingDocument.md
goneall: Updated per suggestion

model/Core/Properties/definingDocument.md
goneall: Updated per suggestion

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -72,6 +72,7 @@ between a Package and a File, between two Packages, or between one SPDXDocument
 - republishedBy: (Security) Designates the agent that tracked, aggregated, and/or enriched vulnerability details to improve context (i.e. NVD)
 - requirementFor: Every `to` Element is required for the `from` Element
 - runtimeDependency: Every `to` Element is a runtime dependency for the `from` Element
+- spdxDocumentForArtfact: Every `to` Element is an SpdxDocument representing the logical serialization metadata for the `from`  Artifact containing the serliazed data
```
goneall, line 75:
```suggestion
- serializedArtifactFor: Every `to` Element is an artifact serialized from the `from` SpdxDocument Element
```

@sbarnum - when I tried to implement the suggested direction in the serialization example, it didn't seem to work - you would want to relate one logical SPDX document to multiple artifacts - not the other way around.  Thinking of changing this back.  What do you think?

model/Core/Vocabularies/RelationshipType.md
sbarnum: @goneall I think the modified description is correct but the relationshipType name is still confusing and is open to misinterpretation of directionality.
If you look at the set of other RelationshipType names I think you will find that the most clear and least confusing are the ones that follow a pattern of "<verb> <preposition>" or "<verb> <noun>". These have clarity of directionality, clarity of the nature of the relationship between the 'to' element and the 'from' element(s), and they can easily be read in the fashion "<'from' element> <RelationshipType> <'to' element(s)>" (e.g., "https://foo.org/spdx/element-4f888b42-12f0-4183-88b8-3b6d60162243" "dependsOn" "https://foo.org/spdx/element-55fc5492-b1b3-4152-a2ab-206e763e2bf7"). Current RelationshipType names that do not follow this pattern (e.g., "devTool", "example", "prerequisite", etc.) are far more ambiguous and open to confusion or misinterpretation. Simply adding a "has" to the beginnings of most of these (e.g., "hasDevTool", "hasExample", "hasPrerequisite") adds a great deal of clarity and significantly reduces the likelihood of confusion or misinterpretation.

Again, I would strongly recommend using "hasSerialization" as the RelationshipType name. The description can provide the detail of what types of elements are valid in the 'to' and 'from position (and this can be formally added to the SHACL once the spec generator is ready to tackle this). It is not needed in the name.

sbarnum: Oops. I just realized that the directionality in the description is reversed. I misread it.

It should be 'from' and SpdxDocument 'to' an Artifact. This is called out in the comment below and is always whatI supported (as my proposal diagrams all show).
I apologize for the confusion.

model/Core/Vocabularies/RelationshipType.md
sbarnum: Oops, you are correct. I just realized that the directionality in the description is reversed. I misread it.

It should be 'from' and SpdxDocument 'to' an Artifact. 
This is what my initial comment above proposed the change be ("I would strongly recommend this RelationshipType be 'from' an "XCollection" 'to' an Artifact and not the other way around.").

I apologize for the confusion.

goneall:
> I believe this PR looks good with one issue remaining to be addressed before approval and merging.
> The RelationshipType name should be a clearer name like "hasSerialization" and the description should describe directionality from 'from' is SpdxDocument' and 'to' is Artifact.

Just updated.  @sbarnum feel free to suggest different wording

sbarnum (approved):
This PR looks good to me.
It conforms to the consensus results of discussions within the Serialization WG and to the broader model.

model/Core/Vocabularies/RelationshipType.md
JPEWdev: The name is still confusing IMHO. Relationships are most clear if they complete the sentence `from` (is) (a) `RELATIONSHIP` `to`. In this case, you are saying:
`SpdxDocument` is a `serializedArtifactFor` `Artifact`

The confusing part here is that the `spdxDocument` is not the serialized artifact (at least AFACIT, correct me if I'm wrong?)

More clear would be either to flip the `from` and `to`
`Artifact` is a `serializedArtifactFor` `SpdxDocument`

or reword:

`SpdxDocument` is `serializedFromArtifact` `Artifact`

JPEWdev: Also, it's helpful if your documentation matches the `from` -> `to` in it's ordering also, so I would say:

`serializedFromArtifact`: The `from` SpdxDocument was serialized by the `to` Artifact

goneall: @JPEWdev @sbarnum - if you could leave a very specific suggestion to the updated name and documentation, I'll accept it in the PR.  Otherwise, we may be going back and forth a bit before I completely meet your expected result ;)

goneall:
@JPEWdev - I tend to agree the current description of the relationship is confusing, but I'm struggling to come up with a clearer description.

The idea is the artifact is generated from the spdx document metadata.  The preferred direction would be generated from, but that  kind of conflicts with the `from` property name.

If you have any ideas how to clarify this and keep the direction, please propose.

JPEWdev:
> @JPEWdev - I tend to agree the current description of the relationship is confusing, but I'm struggling to come up with a clearer description.
> 
> The idea is the artifact is generated from the spdx document metadata. The preferred direction would be generated from, but that kind of conflicts with the `from` property name.
> 
> If you have any ideas how to clarify this and keep the direction, please propose.

Perhaps:

serializedInArtifact: Each `from` SPDXDocument can be found in a serialized form in each `to` Artifact

?

zvr (approved):

goneall:
Discussed on 7 Nov 2023 call - agree to merge

PR #548, opened 2023-11-03
Clarify RelationshipTypes

### LINKED ISSUE #533: All relationshipType descriptions should explicitly state `from` and `to`, or clearly indicate they are directionless
It should be unambiguous when reading the description of a relationshipType what the direction is. Some of the descriptions do a good job of this, but some do not. All of the descriptions should explicitly state the direction using the terms `from` and `to`, or explicitly state they are directionless.

As an example, this description is clear and unambiguous about the direction:

* contains: Every `to` Element is contained by the `from` Element

Whereas this one is ambiguous:

* descendant: This relationship may be used to describe child builds of a Build instance.

And yes I realize the irony that I helped write the latter :)

### LINKED ISSUE #534: relationshipType names should clearly indicate direction
It would be really nice if the relationshipType names clearly indicate the direction of the relationship without necessarily having to refer to the documentation. Many of the current names are _very_ unintuitive, and are difficult if not impossible to get correct without reading the documentation. This currently classifies as a 3 in [API Design][1] which is not very good. We should be able to move this all the way up to 6 ("The name tells you how to use it") by rewording the names. IMHO, the relationship names should unambiguously complete the sentance: `from` RELATIONSHIP `to`. Currently relationships like `ammeds` technically do this, but they are somewhat ambiguous to the direction, and worse the documentation contradicts the intuitive understanding by saying: "Every `to` Element amends the `from` Element". Even other relationships that uses the same sort of terminology (e.g. `contains`, `depends`) have different directionality to the definition, which makes it _very_ hard to figure out how relationships should be written without very careful reading of the documentation.

I would recommend renaming the relationships so it is unambiguous as to the directionality, which will also make it eaiser to add relationships that go in the opposite direction to deal with fan-in vs. fan-out. As an example, `amends` might be renamed `ammenedBy`. If a high fan-out relationship is needed in the future, it could be named `ammendOf`

[1]: https://ozlabs.org/~rusty/index.cgi/tech/2008-03-30.html

JPEWdev:
The current wording of RelationshipTypes is very difficult to understand and use correctly. It is very difficult to determine the directionality of relationships, and often the documentation is difficult to parse as its inverted from the expected directionality.

The renames most relationship types to complete the sentence: `from` (is) (a) `RELATIONSHIP` `to`

Also, as much as possible the documentation is worded so that `from` comes before `to` to further reduce confusion

All relationships are renamed assuming that the text description is the authoritative description of the direction, not the name of the relationship itself. This means existing implementation should be able to find and replace with the new names and still be correct (assuming they were able to figure them out in the first place).

For the sake of easier diffing, the relationships are not sorted based on their new names.

Many of the security relationship were not clearly documented enough to determine the correct directionality and still need to be addressed

Props to AI & Dataset for giving clear relationships that didn't need modified.

Closes #534
Closes #533

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -12,74 +12,80 @@ Provides information about the relationship between two Elements.
 For example, you can represent a relationship between two different Files,
 between a Package and a File, between two Packages, or between one SPDXDocument and another SPDXDocument.
 
+Relationship names be descriptive enough to easily deduce the correct direction
+from their name. The best way to do this is to make sure that the relationship
+name completes the sentence:
+
+`from` (is) (a) `RELATIONSHIP` `to`
+
 ## Metadata
 
 - name: RelationshipType
 
 ## Entries
 
-- affects: (Security/VEX) Designates one or more elements as affected by a vulnerability
-- amends: Every `to` Element amends the `from` Element
-- ancestor: Every `to` Element is an ancestor of the `from` Element
-- availableFrom:  This relationship is used to identify additional suppliers where an artifact is available from.
-- buildDependency: Every `to` Element is a build dependency of the `from` Element
-- buildTool: Build tool used to build an Element. This may be used to describe the build tool of a Build instance
+- affects: (Security/VEX) The `from` vulnerability affect each `to` Element
+- amendedBy: The `from` Element is amended by each `to` Element
+- hasAncestor: The `from` Element has ancestor described by each `to` Element
+- availableFrom: The `from` Element is available from the additional supplier described by each `to` Element
+- hasBuildDependencyOn: The `from` Element has a build dependency on each `to` Element
+- usesBuildTool: The `from` Element uses each `to` Element as a build tool
 - coordinatedBy: (Security) Used to identify the vendor, researcher, or consumer agent performing coordination for a vulnerability
-- concludedLicense: The `to` specifices the license the SPDX data creator has concluded as governing the `from` software Artifact.
-- contains: Every `to` Element is contained by the `from` Element
-- configOf: (Build) Configuration information applied to an Element instance during a LifeycleScopeType period.  Example: Build configuration of the build instance
-- copy: Every `to` Element is a copy of the `from` Element
-- dataFile: Every `to` Element is a data file related to the the `from` Element
-- declaredLicense: The `to` license information identifies the license information actually found in the `from` software Artifact, for example as detected by use of automated tooling.
-- dependencyManifest: Every `to` Element is manifest file containing dependency information related to the `from` Element
-- dependsOn: Every `to` Element is a dependecy of the `from` Element
-- descendant: This relationship may be used to describe child builds of a Build instance.
-- describes: Every `to` Element is described by the `from` Element. To denote the root(s) of a tree of elements in a collection, the rootElement property should be used.
-- devDependency: Every `to` Element is a development dependency for the `from` Element
-- devTool: Every `to` Element is a development tool for the `from` Element
-- distributionArtifact: Every `to` Element is an artifact intended for distribution of the `from` Element (e.g. an RPM or archive file)
-- documentation: Every `to` Element is documentation for the `from` Element
-- doesNotAffect: (Security/VEX) Specifies a vulnerability has no impact on one or more elements
-- dynamicLink: Every `to` Element is dynamically linked to the `from` Element
-- example: Every `to` Element is an example for the `from` Element
-- evidenceFor: (Dataset) Every `to` Element is can be considered as evidence for the `from` Element
-- expandedFromArchive: Every `to` Element is an artifact expanded from the `from` archive file
+- hasConcludedLicense: The `from` Software Artifact is concluded by the SPDX data creator to be governed by each `to` license
+- contains: The `from` Element contains each `to` Element
+- configOf: The `from` Element is a configuration applied to each `to` Element during a LifecycleScopeType period
+- copiedTo: The `from` Element has been copied to each `to` Element
+- hasDataFile: The `from` Element treats each `to` Element as a data file
+- hasDeclaredLicense: The `from` Software Artifact was discovered to actually contain each `to` license, for example as detected by use of automated tooling.
+- hasDependencyManifest: The `from` Element has manifest files that contain dependency information in each `to` Element
+- dependsOn: The `from` Element depends on each `to` Element
+- descendantOf: The `from` Element is a descendant of each `to` Element
+- describes: The `from` Element describes each `to` Element. To denote the root(s) of a tree of elements in a collection, the rootElement property should be used.
+- hasDevDependency: The `from` Element has a development dependency on each `to` Element
+- hasDevTool: The `from` Element uses each `to` Element as a development tool
+- hasDistributionArtifact: The `from` Element is distributed as an artifact in each Element `to`, (e.g. an RPM or archive file)
+- hasDocumentation: The `from` Element is documented by each `to` Element
+- doesNotAffect: (Security/VEX) The `from` Vulnerability has no impact on each `to` Element
+- hasDynamicLink: The `from` Element dynamically links in each `to` Element
+- hasExample: Every `to` Element is an example for the `from` Element (`from` hasExample `to`)
+- hasEvidence: (Dataset) Every `to` Element is considered as evidence for the `from` Element (`from` hasEvidence `to`)
+- expandsTo: The `from` archive expands out as an artifact described by each `to` Element
 - exploitCreatedBy: (Security) Designates an agent has created an exploit against a vulnerability
-- fileAdded: Every `to` Element is is a file added to the `from` Element
-- fileDeleted: Every `to` Element is a file deleted from the `from` Element
-- fileModified: Every `to` Element is a modification of the `from` Element
+- hasAddedFile: Every `to` Element is is a file added to the `from` Element (`from` hasAddedFile `to`)
+- hasDeletedFile: Every `to` Element is a file deleted from the `from` Element (`from` hasDeletedFile `to`)
+- modifiedBy: The `from` Element is modified by each `to` Element
 - fixedBy: (Security) Designates a vulnerability has been fixed by an agent
 - fixedIn: (Security/VEX) A vulnerability has been fixed in one or more elements
 - foundBy: (Security) Designates an agent was the original discoverer of a security vulnerability
-- generates: Every `to` Element is generated from the `from` Element
+- generates: The `from` Element generates each `to` Element
 - hasAssessmentFor: (Security) Relates a Vulnerability and an Element with a security assessment.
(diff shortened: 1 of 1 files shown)


rnjudge:
In addition to reviewing this PR, the security team will open a separate PR to clarify the remaining security relationships.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
 
 ## Entries
 
-- affects: (Security/VEX) Designates one or more elements as affected by a vulnerability
-- amends: Every `to` Element amends the `from` Element
-- ancestor: Every `to` Element is an ancestor of the `from` Element
-- availableFrom:  This relationship is used to identify additional suppliers where an artifact is available from.
-- buildDependency: Every `to` Element is a build dependency of the `from` Element
-- buildTool: Build tool used to build an Element. This may be used to describe the build tool of a Build instance
+- affects: The `from` VEX/vulnerability affect each `to` Element
```
rnjudge, line 27: Instead of `VEX/vulnerability` --> `The `from` Vulnerability element affects each `to` Element`

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
+- amendedBy: The `From` Element is amended by each `to` Element
```
rnjudge, line 28: small nit: `From` should be lowercased for consistency with other relationship definitions

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
+- hasDeclaredLicense: The `from` software Artifact was discovered to actually contain each `to` license, for example as detected by use of automated tooling.
+- hasDependencyManifest: The `from` Element has manifest files that contain dependency information in each `to` Element
+- dependsOn: The `from` Element depends on each `to` Element
+- descendantOf: The `from` Element is a descendant of each `to` Element
+- describes: The `from` Element describes each `to` Element. This can be used to denote the root(s) of a tree of elements contained in an SBOM.
+- hasDevDependency: The `from` Element has a development dependency on each `to` Element
+- hasDevTool: The `from` Element uses each `to` Element as a development tool
+- hasDistributionArtifact: The `from` Element is distributed as an artifact in each Element `to`, (e.g. an RPM or archive file)
+- hasDocumentation: The `from` Element is documented by each `to` Element
+- doesNotAffect: The `from` (Security/VEX) vulnerability has no impact on each `to` Element
```
rnjudge, line 48: Please use: "(Security/VEX) The `from` Vulnerability element does not affect each `to` Element (i.e. product)"

rnjudge (changes requested):
There's a few small changes for the security definitions being changed in this PR. Otherwise looks good. Thanks for all your work on thi8s @JPEWdev!

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
-- ancestor: Every `to` Element is an ancestor of the `from` Element
-- availableFrom:  This relationship is used to identify additional suppliers where an artifact is available from.
-- buildDependency: Every `to` Element is a build dependency of the `from` Element
-- buildTool: Build tool used to build an Element. This may be used to describe the build tool of a Build instance
+- affects: (Security/VEX) The `from` vulnerability affect each `to` Element
+- amendedBy: The `from` Element is amended by each `to` Element
+- hasAncestor: The `from` Element has ancestor described by each `to` Element
+- availableFrom: The `from` Element is available from the additional supplier described by each `to` Element
+- hasBuildDependencyOn: The `from` element has a build dependency on each `to` Element
+- usesBuildTool: The `from` Build element uses each `to` element as a build tool
```
kestewart, line 32: Suggest" The `from` Element uses each `to` Element as a build tool" will align more with the model.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
-- amends: Every `to` Element amends the `from` Element
-- ancestor: Every `to` Element is an ancestor of the `from` Element
-- availableFrom:  This relationship is used to identify additional suppliers where an artifact is available from.
-- buildDependency: Every `to` Element is a build dependency of the `from` Element
-- buildTool: Build tool used to build an Element. This may be used to describe the build tool of a Build instance
+- affects: (Security/VEX) The `from` vulnerability affect each `to` Element
+- amendedBy: The `from` Element is amended by each `to` Element
+- hasAncestor: The `from` Element has ancestor described by each `to` Element
+- availableFrom: The `from` Element is available from the additional supplier described by each `to` Element
+- hasBuildDependencyOn: The `from` element has a build dependency on each `to` Element
```
kestewart, line 31: s/ `from` element/`from` Element/

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
-- devDependency: Every `to` Element is a development dependency for the `from` Element
-- devTool: Every `to` Element is a development tool for the `from` Element
-- distributionArtifact: Every `to` Element is an artifact intended for distribution of the `from` Element (e.g. an RPM or archive file)
-- documentation: Every `to` Element is documentation for the `from` Element
-- doesNotAffect: (Security/VEX) Specifies a vulnerability has no impact on one or more elements
-- dynamicLink: Every `to` Element is dynamically linked to the `from` Element
-- example: Every `to` Element is an example for the `from` Element
-- evidenceFor: (Dataset) Every `to` Element is can be considered as evidence for the `from` Element
-- expandedFromArchive: Every `to` Element is an artifact expanded from the `from` archive file
+- hasConcludedLicense: The `from` software Artifact is concluded by the SPDX data creator to be governed by each `to` license
```
kestewart, line 34: not sure we want to change from concludedLicense to hasConcludedLicense?

s/software Artifact/Artifact/ will future proof this.  Otherwise  software should be capitalized.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
+- contains: The `from` Element contains each `to` Element
+- configOf: The `from` Element is a configuration applied to each `to` Build during a LifecycleScopeType period
```
kestewart, line 36: configOf could be applied to other LifecycleScopeType,  so let's remove "Build" in this sentence.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
+- copiedTo: The `from` Element has been copied to each `to` Element
+- hasDataFile: The `from` Element treats each `to` Element as a data file
+- hasDeclaredLicense: The `from` software Artifact was discovered to actually contain each `to` license, for example as detected by use of automated tooling.
+- hasDependencyManifest: The `from` Element has manifest files that contain dependency information in each `to` Element
+- dependsOn: The `from` Element depends on each `to` Element
+- descendantOf: The `from` Element is a descendant of each `to` Element
+- describes: The `from` Element describes each `to` Element. This can be used to denote the root(s) of a tree of elements contained in an SBOM.
+- hasDevDependency: The `from` Element has a development dependency on each `to` Element
```
kestewart, line 44: We need to use LifecycleScoped relatioship here, and unify this with dependsOn.    But we can do that in a second pass.   Leave it for now like this, and we'll get this applied, and take the next pass at unifying around lifecyclescope.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
+- hasDevTool: The `from` Element uses each `to` Element as a development tool
```
kestewart, line 45: ditto here.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -12,73 +12,79 @@ Provides information about the relationship between two Elements.
...
-- fileDeleted: Every `to` Element is a file deleted from the `from` Element
-- fileModified: Every `to` Element is a modification of the `from` Element
+- hasAddedFile: Every `to` Element is is a file added to the `from` Element (`from` hasAddedFile `to`)
+- hasDeletedFile: Every `to` Element is a file deleted from the `from` Element (`from` hasDeletedFile `to`)
+- modifiedBy: The `from` Element is modified by each `to` Element
 - fixedBy: (Security) Designates a vulnerability has been fixed by an agent
 - fixedIn: (Security/VEX) A vulnerability has been fixed in one or more elements
 - foundBy: (Security) Designates an agent was the original discoverer of a security vulnerability
-- generates: Every `to` Element is generated from the `from` Element
+- generates: The `from` Element generates each `to` Element
```
kestewart, line 60: When I look at the 2.2 relationships in use:
Relationship: SPDXRef-File-libkernel.a GENERATED_FROM DocumentRef-zephyr:SPDXRef-File-main-weak.c
Relationship: SPDXRef-File-libkernel.a GENERATED_FROM DocumentRef-zephyr:SPDXRef-File-banner.c
Relationship: SPDXRef-File-libkernel.a GENERATED_FROM DocumentRef-zephyr:SPDXRef-File-busy-wait.c

so I think this should have the properties swapped to be consistent.

generatedBy may be clearer and consistent with other relationships? 
so something like

generatedBy: The `from` Element is generated by each of the `to` Element(s)

kestewart (changes requested):
Thanks for doing this Joshua - excellent work.   I think there are a few more inconsistencies to still address (see comment).    But let's discuss this in the call, and get this updated and merged.

model/Core/Vocabularies/RelationshipType.md
JPEWdev: Right. I was trying to preserve the originally stated order, but I'm happy to flip it around if we like it better. `generatedFrom` is fine, as it is still clear what the directionality should be from the name.

model/Core/Vocabularies/RelationshipType.md
JPEWdev: The intention is to preserve the original order of the relationship, but make it clear from the name what the directionality is (e.g. I'm not changing how the relationship is used, just the name). `concludedLicense` is ambiguous as to the direction, and TBH I'd rather have good names that are easy for new users to understand than worry about exactly maintaining the names from SPDX 2

model/Core/Vocabularies/RelationshipType.md
kestewart: From discussion,  ok to flip order and match SPDX 2.

JPEWdev: Will be flipped in a separate PR

kestewart (approved):
Changes look good.  Applying this and we'll refine further if needed.

PR #552, opened 2023-11-09
Allow ElementCollection to have 0 elements

### LINKED ISSUE #526: Allow ElementCollection to contain 0 elements
Currently, the minimum count for the `element` field of `ElementCollection` is 1. This means that any time a producer want to return no elements to a consumer, it must do so using and out-of-band mechanism, which is annoying for implementation.

For example, imagine a server that implements a query API and returns the elements that match the query in an `ElementCollection`. Currently, if you provided a query that doesn't match any elements, the server cannot return an empty `ElementCollection` to indicate this and instead must implement some other out-of-band mechanism to tell the clients that no elements match the query, even though this is likely not really an error.

### LINKED ISSUE #551: SpdxDocument.element minimum cardinality of 1 requires special casing for JSON-LD
When writing a JSON-LD document, one is supposed to take the `SpdxDocument.element` list and instead seralize it as the `@graph` in the document. While this makes the document neater and more "JSON-LD", it's not actually considered valid because the minimum cardinality of `SpdxDocument.element` is 1, and it will never have anything in that list in reality. This means validation requires a special case, which means standard SHACL validators cannot be used.

If #526 is accepted, it would fix this also

### LINKED ISSUE #562: Missing license cross reference properties
The ["Other web pages for license/exception"](https://github.com/spdx/license-list-XML/blob/main/DOCS/license-fields.md) is missing from the `ExpandedLicensing` License `LicenseAddition` classes.

From the [XML Schema](https://github.com/spdx/license-list-XML/blob/main/schema/ListedLicense.xsd), the following property should be added:
- ~~crossRefs~~ `seeAlso` of type `AnyURI`

JPEWdev:
As discussed in #526, an ElementCollection have 0 Elements is useful to prevent parties that are exchanging SPDX content from needing to have an out-of-band mechanism to indicate the case where there are no elements to be transferred.

Closes #526
Closes #551

### DIFF
--- a/model/Core/Classes/ElementCollection.md
+++ b/model/Core/Classes/ElementCollection.md
@@ -13,6 +13,9 @@ An ElementCollection is a collection of Elements, not necessarily with unifying 
 Note that all ElementCollections must conform to the core profile even if the core profile is no specified in the profileConformance property.
 If the profileConformance property is not provided, core is to be assumed as the default.
 
+**Constraints**
+If the ElementCollection has at least 1 element, it must also have at least 1 rootElement
+
 ## Metadata
 
 - name: ElementCollection
@@ -23,10 +26,10 @@ If the profileConformance property is not provided, core is to be assumed as the
 
 - element
   - type: Element
-  - minCount: 1
+  - minCount: 0
 - rootElement
   - type: Element
-  - minCount: 1
+  - minCount: 0
 - imports
   - type: ExternalMap
 - profileConformance


kestewart (approved):

model/Core/Classes/ElementCollection.md
```diff
@@ -13,6 +13,9 @@ An ElementCollection is a collection of Elements, not necessarily with unifying
 Note that all ElementCollections must conform to the core profile even if the core profile is no specified in the profileConformance property.
 If the profileConformance property is not provided, core is to be assumed as the default.
 
+**Constraints**
+If the ElementCollection has at least 1 element, it must also have at least 1 rootElement
```
maxhbr, line 17: Is that the way how we encode constraints?

This feels like a step in the direction of `filesAnalyzed` and it introduces more complexity. Not sure if I like that.

JPEWdev: I thought there was a better way to encode the constraints, but I was unable to find an example, so I put this in as a placeholder for now

sbarnum (approved):
Looks like there is consensus on this change and specific implementation.

PR #556, opened 2023-11-20
Cleanup relationships

JPEWdev:
Performs several cleanups to the relationship types per discussion with @kestewart 

Closes #550


kestewart (approved):
This matches what we discussed,  and cleans up the security relationship ordering collision.

nishakm (approved):

PR #559, opened 2023-11-28
Fix context URLs in JSON-LD docs

BigBlueHat:
Thanks for building this context file! It's great to have this--and seems useful already even with SPDX 2.x files.

Cheers!
:tophat:


goneall:
Thanks @BigBlueHat!  

@zvr - should we update the links here or change the CI to publish to the rdf directory?

kestewart (approved):

zvr:
Any of the two alternatives will work, and since we already have the PR for implementing one, let's go with this one. This is temporary, anyway, as it will eventually be pushed to spdx.org.

zvr (approved):

PR #563, opened 2023-12-03
Updated Lite profile definition, to eliminate PackageUrl and added decs. for conformance to other profiles.

yoshi-i:
Due to write permissions, commit applied to forked repository

### DIFF
--- a/model/Lite/Lite.md
+++ b/model/Lite/Lite.md
@@ -26,13 +26,12 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
 For a /Software/Package to be conformant with this profile,
 the following has to hold:
 
+1. The mincount for /Core/Element/spdxId is 1
 1. The mincount for /Core/Element/name is 1
+1. The mincount for /Core/Element/creationInfo is 1
+1. The mincount for /Core/Artifact/suppliedBy is 1
 1. The mincount for /Software/Package/packageVersion is 1
-1. The mincount for /Software/Package/packageUrl is 1
 1. The mincount for /Software/SoftwareArtifact/copyrightText is 1
-
-For a /Software/Sbom to be conformant with this profile,
-the following has to hold:
 
 1. for every `/Software/Package` there MUST exist exactly one `/Core/Relationship`
    of type `concludedLicense` having that element as its `from` property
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM - Thanks @yoshi-i

kestewart (approved):
Changes match what was discussed.   Approved.

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
 For a /Software/Package to be conformant with this profile,
 the following has to hold:
 
+1. The mincount for /Core/Element/spdxId is 1
 1. The mincount for /Core/Element/name is 1
+1. The mincount for /Core/Element/creationInfo is 1
```
zvr, line 31: Again, already a requirement in Core, no need to re-state it here.

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
    of type `declaredLicense` having that element as its `from` property
    and an `/SimpleLicensing/AnyLicenseInfo` as its `to` property.
 
-- TODO - add other class restrictions
+For a /Software/Sbom to be conformant with this profile,
+the following has to hold:
+
+1. The mincount for /Core/ElementCollection/element is 1
+1. The mincount for /Core/ElementCollection/rootElement is 1
+1. The mincount for /Core/Element/spdxId is 1
```
zvr, line 48: same as above

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
+1. The mincount for /Core/Element/creationInfo is 1
```
zvr, line 49: same as above

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
+
+For a /Core/Element/CreationInfo to be conformant with this profile,
+the following has to hold:
```
zvr, line 52: All of these are not needed:
- `specVersion`, `created`, and `createdBy` have already mincount 1 in Core
- `profile` was moved (I think)
- `dataLicense` is... yet to be determined, but it most probably won't be there

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
+
+1. The mincount for /Core/Element/CreationInfo/specVersion is 1
+1. The mincount for /Core/Element/CreationInfo/created is 1
+1. The mincount for /Core/Element/CreationInfo/createdBy is 1
+1. The mincount for /Core/Element/CreationInfo/profile is 1
+1. The mincount for /Core/Element/CreationInfo/dataLicense is 1
+
+For a /Core/Agent to be conformant with this profile,
+the following has to hold:
```
zvr, line 61: Again, not needed, `spdxId` and `creationInfo` have mincount 1 for all Elements (therefore for Agents)

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
+
+1. The mincount for /Core/Agent/spdxId is 1
+1. The mincount for /Core/Agent/creationInfo is 1
+
+For a /Core/Relationship to be conformant with this profile,
+the following has to hold:
```
zvr, line 67: All of these already in Core

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
+
+1. The mincount for /Core/Relationship/from is 1
+1. The mincount for /Core/Relationship/relationshipType is 1
+1. The mincount for /Core/Relationship/spdxId is 1
+1. The mincount for /Core/Relationship/creationInfo is 1
+
+For a /SimpleLicensing/LicenseExpression to be conformant with this profile,
+the following has to hold:
```
zvr, line 75: These restrictions already exist.

model/Lite/Lite.md
```diff
@@ -26,19 +26,61 @@ An SPDX Lite document can be used in parallel with SPDX documents in software su
...
+
+1. The mincount for /SimpleLicensing/LicenseExoression/licenseExpression is 1
+1. The mincount for /SimpleLicensing/LicenseExoression/spdxId is 1
+1. The mincount for /SimpleLicensing/LicenseExoression/creationInfo is 1
+
+For a /SimpleLicensing/SimpleLicensingText to be conformant with this profile,
+the following has to hold:
```
zvr, line 82: These as well.

goneall:
> Most of what is specified here can be deleted, as the restrictions are already present in the other profiles.

@zvr - Thanks for the review.  Agree that the information is redundant and technically not needed.  We discussed this on the call with the Lite team and we agreed that the documentation would be clearer if we included the duplicated information.

Do you think including the duplicated information would cause any confusion or other issues?

goneall:
Thanks @zvr for the review.

Perhaps we should remove the redundant information as you suggested and solve the documentation in another way so that it doesn't trip up the future creation of schema generation.

@yoshi-i - if you agree, please update the PR per suggestions.  Otherwise, we can discuss on Tuesday's tech call.

yoshi-i:
@zvr -san and @goneall -san,

Our intention is to clarify that it make Lite profile specification as it can be understood at a glance which key-value pairs should be included in the actual SBOM, such as a JSON file, as similar situations as using SPDX Lite on SPDX 2.3 Annex G.

> But there is always the underlying assumption that the Core profile is present and conformant! The Lite profile will not be available on its own. 

It prefer to cut out the necessary items, and I want to keep the each portions cut-out are compatible to original, but it's not intend to incorporate all of them...

I believe that a final specification document is generated from each markdown files, we can agree to remove redundant definitions from Lite.md if the essential parts of other markdown files can be compiled into a Lite profile specification page or formal example for consistency with other required profiles.

When the above features are provided, it read the Lite profile section of the specification should provide sufficient understanding of practical use cases, I think.

If it seems necessary to discuss during tech call, please provide a response.

yoshi-i:
@zvr -san, @goneall -san, @NorioKobota -san and @tak -san,

Thank you for discussing about duplicated definitions and how we describe. I'd added commit into this PR eliminating dups. and added short wording of reference to an Annex.

goneall:
Thanks @yoshi-i - changes look good to me.

@zvr - if these look OK to you, please go ahead and merge.

zvr (approved):

PR #564, opened 2023-12-05
fix cardinality of contentType in Annotation

maxhbr:
Was discussed in tech call, during review of drawio.


kestewart (approved):
Matches understanding from the notes.   Merging.

JPEWdev (approved):

PR #579, opened 2023-12-08
The CI run should also commit the json model dump to gh-pages

maxhbr:
The spec parser can generate a json representation of the content of the md files. This is for example used in https://github.com/spdx/tools-python/pull/726 and it would be helpful if the CI here would generate and persist these files.

### DIFF
--- a/.github/workflows/spec-parser.yml
+++ b/.github/workflows/spec-parser.yml
@@ -36,6 +36,7 @@ jobs:
               --gen-md \
               --gen-refs \
               --gen-rdf \
+              --json-dump \
               --out-dir /tmp/spec-parser.out \
               ../model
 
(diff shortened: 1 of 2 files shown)


zvr:
The json dump is arbitrary, completely undocumented, and it should not be used for anything. 
The new parser also generates a json dump, but of course the structure is different.

.github/workflows/spec-parser.yml
```diff
@@ -87,13 +88,14 @@ jobs:
...
           cp /tmp/spec-parser.out/context.json .
+          cp /tmp/spec-parser.out/model_dump.json .
 
           rm -rf /auto-generated/
           cp -r /tmp/auto-generated/ .
 
           if [ -n "$(git status -s -- "auto-generated/" "*.ttl" "*.jsonld" "*.json")" ]
           then
-            git add model.ttl model.jsonld context.json auto-generated/
+            git add model.ttl model.jsonld context.json model_dump.json auto-generated/
```
zvr, line 98: the file is named `model.json` in the new parser

zvr (changes requested):

kestewart:
Per discussion with Joshua in call this is depending on whether the SHACL model is good enough or not.

maxhbr:
this will be replaced by output or use of https://github.com/JPEWdev/shacl2code

PR #582, opened 2023-12-19
Create an initial browseable model for the gh-pages site.

licquia:
Once this is merged, the new browseable model site will live here:

https://spdx.github.io/spdx-3-model/browseable-spec

The long-term goal is to create a complete spec here; the model files are the first step.  Also, any improvements to the look and feel of the model files should be easy to integrate here once available.

### DIFF
--- a/browseable/gen-index-page.py
+++ b/browseable/gen-index-page.py
@@ -1,0 +1,52 @@
+#!/usr/bin/env python3
+
+import sys
+import os
+import re
+
+
+def generate_main_index(toplevel_path):
+    template_path = os.path.join(toplevel_path, 'index.md.in')
+    with open(os.path.join(toplevel_path, 'index.md'), 'w') as index:
+        with open(template_path) as template:
+            index.write(template.read())
+        for item in os.listdir(toplevel_path):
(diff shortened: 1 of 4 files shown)


zvr (changes requested):
It was decided that the browsable spec will be in the `spdx-spec` repo, not here.

So all this has to be done there.

licquia:
That's true, but some parts of this will also have to be done here.

What I've pushed here is a first step, which provides the first pass at a browseable spec.  It's not official, and is certainly not complete, but it does represent putting part of the spec online in a way that's accessible.  My plan is to build from this base.

Do you have specific objections to this approach?

browseable/gen-index-page.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
```
maxhbr, line 1: missing license header? I think such code should be under an source code license and not the "Community Specification License 1.0"

goneall:
@licquia - from our recent discussions, we'll probably move this to the spdx-spec repo.  Should we close this PR?

licquia:
Yup.  Closing.

PR #583, opened 2023-12-19
Move AnyLicenseInfo to Core

JPEWdev:
Moves the AnyLicenseInfo class to core. This makes sense so that specific profiles can implement specific licensing implementations, but other profiles do not need to be aware of them and can use the AnyLicenseInfo from core to refer to a license without any specific implementation.

As an example, SpdxDocument from Core references AnyLicenseInfo, so it shouldn't live in SimpleLicense profile.

### DIFF
--- a/model/AI/AI.md
+++ b/model/AI/AI.md
@@ -15,14 +15,14 @@ The AI profile namespace defines concepts related to AI application and model ar
 - id: https://rdf.spdx.org/v3/AI
 - name: AI
 
-## Profile conformance 
+## Profile conformance
 
 For an element collection to be conformant with this profile,
 the following has to hold:
 
 1. for every `/AI/AIPackage` there MUST exist exactly one `/Core/Relationship`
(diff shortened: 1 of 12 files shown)


maxhbr:
why not move everything from SimpleLicensing to core? Maybe we can get rid of that profile?

goneall:
In general, should we have a rule the the Core profile should not reference any classes or properties outside of core?  If so, should Software profile not reference any classes or properties outside of core and software?

JPEWdev:
> In general, should we have a rule the the Core profile should not reference any classes or properties outside of core? If so, should Software profile not reference any classes or properties outside of core and software?

IMHO, Core should not reference anything outside of Core. If other profiles have inter-core dependencies, that's less of a concern to me; making a rule like that _might_ be too limiting. Maybe just a strong recommendation?

JPEWdev:
Since `AnyLicenseInfo` is not mandatory, it's fine to live in SimpleLicensing. #593 fixes this

PR #585, opened 2023-12-22
Remove the SoftwareDependencyRelationship Class and it's related properties. 

kestewart:
Removing class (and it's properties)  as discussed on in November 21 Tech call.


goneall:
./model/Software/Classes/SoftwareDependencyRelationship.md also needs to be deleted

goneall (approved):
LGTM - based on the tech call decision

PR #586, opened 2023-12-22
Put max 1 cardinality on dataLicense

goneall:


swinslow (approved):
+1, LGTM. dataLicense should be optional, but no more than one value specified. If multiple licenses apply, then an SPDX License Expression can be used to combine them appropriately.

PR #588, opened 2023-12-23
Create concrete classes for license individuals

### LINKED ISSUE #574: ExpandedLicensing profile/namespace specifies two individuals (NoAssertionLicense and NoneLIcense) as of type AnyLicenseInfo which is invalid since AnyLIcenseInfo is abstract
The two individuals (NoAssertionLicense and NoneLIcense) need to be redefined as instances of a concrete subclassof AnyLicenseInfo.

goneall:
Fixes #574

This PR implements concrete classes by the same name as the individuals.  The concrete classes inherit from a comment `IndividualLicense` class.

An alternative solution to #574 is to have the individuals use the `IndividualLicense` type directly.

The only reason I chose this approach is that having different types may match some language implementations more closely - but I could easily be persuaded to take the alternative solution.


swinslow:
@goneall Thank you for this! Taking a look now and I will circle back.

model/ExpandedLicensing/Classes/IndividualLicense.md
```diff
@@ -0,0 +1,21 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# IndividualLicense
+
+## Summary
+
+A concrete subclass of AnyLicenseInfo used by Individuals in the ExpandedLicensing profile.
```
swinslow, line 7: See my overall comment about whether to use the approach described here or the alternative you mentioned in the discussion thread. Assuming we use the one you did here, though, then is it accurate that IndividualLicense is a "concrete" subclass? Or is it an abstract subclass, since only the NoneLicense and NoAssertionLicense classes should actually be instantiated?

model/ExpandedLicensing/Classes/IndividualLicense.md
```diff
@@ -0,0 +1,21 @@
...
+
+Individuals, such as NoneLicense and NoAssertionLicense, need to reference a concrete subclass of AnyLicenseInfo.
+
+This class provides the type used by the individuals.
+
+## Metadata
+
+- name: IndividualLicense
+- SubclassOf: /SimpleLicensing/AnyLicenseInfo
+- Instantiability: Concrete
```
swinslow, line 20: Same question / comment as above

model/ExpandedLicensing/Classes/NoAssertionLicense.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# NoAssertionLicense
+
+## Summary
+
+A License when no assertion can be made about its actual value.
```
swinslow, line 7: To align with the NoneLicense class text below, should this say something like:

"A concrete subclass of IndividualLicense used by the NoAssertionLicense Individual in the ExpandedLicensing profile."

model/ExpandedLicensing/Classes/NoneLicense.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# NoneLicense
+
+## Summary
+
+A concrete subclass of AnyLicenseInfo used by Individuals in the ExpandedLicensing profile.
```
swinslow, line 7: Should this reference `IndividualLicense` instead of `AnyLicenseInfo`? And also specify which Individual uses it? E.g.:

"A concrete subclass of IndividualLicense used by the NoneLicense Individual in the ExpandedLicensing profile."

swinslow (commented):
@goneall Thanks again for this!

I don't have a particular preference for whether to take the approach you described here, or the alternate approach you mentioned of having the individuals NoneLicense / NoAssertionLicense use `IndividualLicense` directly.

I'll defer to others if there are strong feelings on that point, but since there haven't been any comments here for the past couple of weeks, I'm inclined to go with what you proposed here.

Assuming we take this approach, I have a few comments -- please see below and let me know what you think!

model/ExpandedLicensing/Classes/IndividualLicense.md
goneall: Thanks @swinslow - really good catches.

After reviewing the comments, I no longer see a need for the IndividualLicense class, I can just subclass AnyLicenseInfo directly for None and NoAssertion license types.

I'll update the PR accordingly.

model/ExpandedLicensing/Classes/NoAssertionLicense.md
goneall: I'll align with the text below.

model/ExpandedLicensing/Classes/NoneLicense.md
goneall: We'll solve this by getting rid of `IndividualLicense`

zvr (changes requested):
I don't agree with creating two new classes, with the exact same name as the individuals(!). People will confuse them, they might use the class instead of the individual, etc. It will become messy very quickly,..

And having these classes does not really help. They are concrete, so... can I use them to create more instances/individuals?

I think this approach creates more problems than it solves.
What is the problem we're trying to solve?

sbarnum:
I agree with @zvr that this approach seems overly complicated and unlikely to lead to the targeted goal.

It seems to me that the simple solution to the issue identified in #574 is simply to make NoAssertionLicense and NoneLicense to be individuals of the LicenseExpression class which is an already existing concrete subclass of AnyLicenseInfo. You could simply have `licenseExpression`="NONE" or "NOASSERTION" respectively for each Individual.

In the definition of the `licenseExpression` property on the LicenseExpression class it states "A license expression could be **a single license identifier found on the SPDX License List**; a user defined license reference denoted by the LicenseRef-idString; a license identifier combined with an SPDX exception; or some combination of license identifiers, license references and exceptions constructed using a small set of defined operators (e.g., AND, OR, WITH and +)."
This should cover the NoAssertionLicense and NoneLicense shouldn't it?

**NOTE:** Individuals are simply instance objects. They are asserted to be members of one or more classes and they MUST have an globally unique ID (in our case an spdxId). This means that in the SPDX spec where we specify Individuals we need to be explicit that the "name" specified for the Individual is actually its spdxId and those names need to explicitly include the base SPDX id prefix for SPDX defined Individuals (i.e., something more like "https://ontology.spdx.org/3.0/licenses/individuals/NoAssertionLicense" rather than just "NoAssertionLicense"). 
It seems like it would make sense to keep the "name" field in the markdown spec for an Individual and simply add a new "Id" field with the full Id that the then the value of the "Id" field could be used to generate the id of the formal rdf/owl Individual.


In a nutshell, the only changes required in a PR would be:

1. In model/ExpandedLicensing/Individuals/NoAssertionLicense.md 
    1.1  modify from "type: /SimpleLicensing/AnyLicenseInfo" to "type: /SimpleLicensing/LicenseExpression"
    1.2 add "id: "<SPDX base prefix for Individuals>NoAssertionLicense""
    1.3 under "Property Values" modify from "name: "NOASSERTION"" to "licenseExpression: "NOASSERTION""
2. In model/ExpandedLicensing/Individuals/NoneLicense.md 
    1.1  modify from "type: /SimpleLicensing/AnyLicenseInfo" to "type: /SimpleLicensing/LicenseExpression"
    1.2 add "id: "<SPDX base prefix for Individuals>NoneLicense""
    1.3 under "Property Values" modify from "name: "NONE"" to "licenseExpression: "NONE""

model/ExpandedLicensing/Classes/IndividualLicense.md
```diff
@@ -1,21 +0,0 @@
-SPDX-License-Identifier: Community-Spec-1.0
-
-# IndividualLicense
```
goneall: IndividualLicenseInfo instead of IndividualLicense

Restore this class

zvr: IndividualLicensingInfo... since it's not info on a license

goneall:
Based on tech call on 23 Jan 2023:

- Restore the "IndividualLicense" but rename it "IndividualLicensingInfo"
- Remove the NoneLicense and NoAssertion license classes from this PR
- Change the individuals to be of the type IndividualLicenseingInfo
- Have the IDs "https://rdf.spdx.org/v3/Licensing/None
https://rdf.spdx.org/v3/Licensing/NoAssertion"
- Have a name "None" and "NoAssertion"

goneall:
@zvr @sbarnum I just updated the PR with the changes we discussed.

I used the names "NONE" and "NOASSERTION" to be compatible with the previous names used in SPDX 2.X.

For the ID's, I didn't include anything in the markdown file.  I figure we would generate the correct ID based on the individual name and profile rather than specify it in the file.

model/ExpandedLicensing/Individuals/NoAssertionLicense.md
```diff
@@ -15,7 +15,7 @@ the SPDX creator has intentionally provided no information (no meaning should be
 ## Metadata
 
 - name: NoAssertionLicense
-- type: /SimpleLicensing/AnyLicenseInfo
+- type: IndividualLicensingInfo
```
maxhbr, line 18: why is [model/ExpandedLicensing/Individuals/NoAssertionLicense.md](https://github.com/spdx/spdx-3-model/pull/588/files#diff-7f1aaba6ef962a460d62c1e2e96b9381dbd86a6a1d0ff06ce26f16966a579b16) not of type [model/ExpandedLicensing/Classes/NoAssertionLicense.md](https://github.com/spdx/spdx-3-model/pull/588/files#diff-2ac21c661dbfd86dbf49893a827f500545da2efb57b35a664a69cb3ba09ac2b7)? I am confused to see it in Individuals and Classes an the same time

model/ExpandedLicensing/Classes/NoAssertionLicense.md
```diff
@@ -0,0 +1,20 @@
...
+## Description
+
+NoAssertionLicense should be used if the SPDX creator has attempted to but cannot reach a reasonable objective determination;
+the SPDX creator has made no attempt to determine this field; or
+the SPDX creator has intentionally provided no information (no meaning should be implied by doing so).
+
+## Metadata
+
+- name: NoAssertionLicense
+- SubclassOf: /SimpleLicensing/AnyLicenseInfo
```
maxhbr, line 19: Why is [model/ExpandedLicensing/Classes/NoAssertionLicense.md](https://github.com/spdx/spdx-3-model/pull/588/files#diff-2ac21c661dbfd86dbf49893a827f500545da2efb57b35a664a69cb3ba09ac2b7) not  a subclass of [model/ExpandedLicensing/Classes/IndividualLicensingInfo.md](https://github.com/spdx/spdx-3-model/pull/588/files#diff-aadc6c742d7ee00d8904c325c153391f148e9cf32f56a5c1340da2f029099929), which is "A concrete subclass of AnyLicenseInfo used by Individuals in the ExpandedLicensing profile."

maxhbr (commented):
Sorry, I am late to this PR and from reading the diff I can not understand why there are files `.../Classes/NoAssertionLicense.md` and `.../Individuals/NoAssertionLicense.md`, but seemingly no connection between them. 

Please ignore my question, if that is clear for everyone else.

zvr:
You are correct, @maxhbr . @goneall simply forgot to delete the two classes (None and NoAssertion).

The idea is to have only the new class `IndividualLicensingInfo` (concrete, from `SimpleLicensing/AnyLicenseInfo`) and the two individuals as instances of that class.

model/ExpandedLicensing/Classes/NoAssertionLicense.md
zvr: This file should be deleted.

model/ExpandedLicensing/Classes/NoneLicense.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Description
+
+NoneLicense should be used if the SPDX creator determines there is no license available for this Artifact.
+
+## Metadata
+
+- name: NoneLicense
+- SubclassOf: /SimpleLicensing/AnyLicenseInfo
+- Instantiability: Concrete
```
zvr, line 18: This file should be deleted.

zvr (changes requested):
If the two class files are deleted, this is good to go.

goneall:
> If the two class files are deleted, this is good to go.

Good catch @zvr - The two files are now deleted

zvr (approved):

sbarnum:
> @zvr @sbarnum I just updated the PR with the changes we discussed.
> 
> I used the names "NONE" and "NOASSERTION" to be compatible with the previous names used in SPDX 2.X.
> 
> For the ID's, I didn't include anything in the markdown file. I figure we would generate the correct ID based on the individual name and profile rather than specify it in the file.

There are only two properties that ANY Individual in ANY ontology MUST define: ID and TYPE.
They MAY define other properties but only have to based on property cardinality restrictions on the TYPE Class.

I strongly believe that the markdown version of the spec needs to include the IDs for ANY Individuals including these two.
When dealing with an Individual the ID **_is_** the actual identifier/name for that object. 
The 'name' property is only a label.

The intent of the markdown version of the spec is for human readability and clarity.
Anyone who wants to reference either of these Individuals in actual SPDX 3.X content MUST use the ID not the 'name'.
Any code written to deal with the concepts of NONE and NOASSERTION MUST use the ID not the 'name'.

I believe including a 'name' but not the one and only one ID to use for each Individual is very highly likely to lead to confusion and invalid content.

zvr:
@sbarnum yes, the individuals must have an id (and they will, in the generated docs and RDF ontology).

@goneall yes, if you put it right now it will break the spec-parser. I will make and push the change on Tue 13 Feb at the latest.

I propose we merge this one and we know we have to add the IRIs at a later stage. If you both remember, I had proposed to use `https://rdf.spdx.org/v3/Licensing/None` and `.../NoAssertion` (or something like that, I will check my notes once I'm back).

goneall:
I'll go ahead and merge - we have another issue to track the creation of the IDs.

PR #593, opened 2024-01-09
Fix type for SpdxDocument AnyLicenseInfo

goneall:


JPEWdev (approved):

swinslow (approved):
I think this makes sense, LGTM.

PR #596, opened 2024-01-11
Updating the IRIs used in RDF

zvr:
This simply updates the documentation on RDF,
to correctly match the IRIs used in the model,
as present in each namespace file.

### DIFF
--- a/serialization/rdf.md
+++ b/serialization/rdf.md
@@ -2,10 +2,19 @@
 
 SPDX data can be serialized in RDF. This can be saved in a variety of formats, like XML, JSON-LD, Turtle, etc.
 
-1. The namespace for SPDX v3 is `http://spdx.org/rdf/v3`
+1. The namespace for SPDX v3 is `https://rdf.spdx.org/v3`
 
-1. IRIs for a class are of the form: `http://spdx.org/rdf/v3/{Namespacename}/{Classname}`
+1. IRIs for a namespace/profile are of the form: `https://rdf.spdx.org/v3/{Namespacename}`
 
-1. IRIs for a property are of the form: `http://spdx.org/rdf/v3/{Namespacename}/{Propertyname}`
+1. IRIs for a class are of the form: `https://rdf.spdx.org/v3/{Namespacename}/{Classname}`
 
-1. IRIs for an enumerated value are of the form: `http://spdx.org/rdf/v3/{Namespacename}/{Classname}/{Valuename}`
+1. IRIs for a property are of the form: `https://rdf.spdx.org/v3/{Namespacename}/{Propertyname}`
+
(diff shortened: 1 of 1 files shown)


serialization/rdf.md
```diff
@@ -2,10 +2,19 @@
 
 SPDX data can be serialized in RDF. This can be saved in a variety of formats, like XML, JSON-LD, Turtle, etc.
 
-1. The namespace for SPDX v3 is `http://spdx.org/rdf/v3`
+1. The namespace for SPDX v3 is `https://rdf.spdx.org/v3`
```
goneall, line 5: In the current spec (version 2.X) the namespace also is a URL pointing to the documentation.

If we change it to `rdf.spdx.org`, we would need to register the subdomain through the LF.

Is this something we want to do? It make take a bit of work and time to get through the LF IT - but I can make an attempt at it if we want to change it.

goneall (changes requested):

zvr:
The purpose of this PR is simply to harmonize the doc about RDF with the addresses actually used in the model files, so that we have a consistent set that can be presented without raising questions. The alternatives were either change the single doc file or change each and every namespace file. I opted for the easiest path.

We can discuss/decide the actual format at a later stage; I don't expect anyone to start using these for a while.

Regarding what they should be, I agree that they should point to the ontology documentation. There are two implementaion alternatives:
- create a new DNS entry and have it pointing to somewhere where static web content is server; or
- carve out a subtree from the current WordPress setup to serve static content from somewhere.

No predictions on which of the two will be easier with LF... ;-)  We can ask and decide later.

But again, this is only to have everything aligned.

serialization/rdf.md
goneall: I opened a separate issue to track

goneall (approved):

goneall:
> The purpose of this PR is simply to harmonize the doc about RDF with the addresses actually used in the model files, so that we have a consistent set that can be presented without raising questions. The alternatives were either change the single doc file or change each and every namespace file. I opted for the easiest path.
> 
> We can discuss/decide the actual format at a later stage; I don't expect anyone to start using these for a while.
> 
> Regarding what they should be, I agree that they should point to the ontology documentation. There are two implementaion alternatives:
> 
> * create a new DNS entry and have it pointing to somewhere where static web content is server; or
> * carve out a subtree from the current WordPress setup to serve static content from somewhere.
> 
> No predictions on which of the two will be easier with LF... ;-) We can ask and decide later.
> 
> But again, this is only to have everything aligned.

I opened issue #596 to track resolving this before release of 3.0.  I'll go ahead and approve this change.

kestewart (approved):
Approved pending securing of web namespace for rdf.spdx.org

kestewart:
Discussed in tech call,  no issues raised, so going ahead and merging.

PR #597, opened 2024-01-11
Update SWHID description

zvr:
Updates the description of `swhid` as an entry in the vocabulary `Core/ExternalIdentifierType`


kestewart (changes requested):
Can you add linkage to the identification parsing specification in there, so it is clear where folks can go to look up more information.     I'm fine with putting the additional wording in, but probably should point to https://www.swhid.org/specification/v1.1/4.Syntax/ or something similar.

zvr:
How's that, @kestewart ?

goneall (approved):
LGTM

PR #598, opened 2024-01-11
Alternative integrity methods for packages 

goneall:
This PR is an alternative to PR #498.

Compared to the code in PR #498, this PR changes the following in responses to [comments made](https://github.com/spdx/spdx-3-model/pull/498#discussion_r1332675795) in PR #498:
- Removes the abstract class restriction on `verifiedUsing` on the `Package` class
- Removes the abstract class `PackageVerification`
- Remove the concrete class `PackageHash`
- Add statements in the description for `Package` how to verify a single file package distribution

Compared to the code in PR #498, this PR changes the following in responses to commented made in issue #345:
- Create a new subclass of `IntegrityMethod` `IdentifierIntegrityMethod` with a single property to reference an external identifier.  I took this approach rather than creating a new enumeration since it duplicates an existing vocabulary.
- Added a statement in the description of `Package` how this integrity method is to be used.
- Removed the concrete classes `PackageGitoid` and `PackageSwid`.

NOTE: This alternative does not allow for restricting a package verification to any specific methods.

It does, however, remove an abstract class layer that seemed to cause confusion and concern.  It also removes possible duplication of data for external identifiers.


goneall:
Based on the package verification meeting (part of the Security profile team) discussion, we decided that the model does support package verification codes as is.  A separate PR will be opened to better document this in the spec.

Closing this PR as it is no longer needed.

PR #600, opened 2024-01-13
Create Readme.md

VenkatTechnologist:


goneall:
I agree with having a README as long as it doesn't trip up the spec parser CI scripts - which I don't think it will.

@zvr - any thoughts?

@VenkatTechnologist - if @zvr doesn't object (he maintains the spec parser which uses the contents of this directory as input), it would be great if you could put in proposed content once you're up to speed.

zvr:
No, please let's keep the `model` directory with actual files of the model, automatically processed. People should not be looking at the directory via GitHub interface, so the proposed `README` will not be automatically visible.

No objections to document it in the top-level `README`.

zvr (changes requested):
Let's not add stuff in the automatically processed `model` directory.

VenkatTechnologist:
> No, please let's keep the `model` directory with actual files of the model, automatically processed. People should not be looking at the directory via GitHub interface, so the proposed `README` will not be automatically visible.
> 
> No objections to document it in the top-level `README`.

Thanks for your input, @zvr . I'm okay either way - with having the model sub-directories related information in the main level README or in the model directory. Just fyi, when I search for SPDX in the Internet, the second top page in the results happens to be the github link of the model, so I think people will land at model often; in fact, I did :). But if you feel strongly that model directory should have files only for processing, that's fine with me.

@goneall , your thoughts?

goneall:
> > No, please let's keep the `model` directory with actual files of the model, automatically processed. People should not be looking at the directory via GitHub interface, so the proposed `README` will not be automatically visible.
> > No objections to document it in the top-level `README`.
> 
> Thanks for your input, @zvr . I'm okay either way - with having the model sub-directories related information in the main level README or in the model directory. Just fyi, when I search for SPDX 3.0 in the Internet, the second top page in the results happens to be the github link of the model, so I think people will land at model often; in fact, I did :). But if you feel strongly that model directory should have files only for processing, that's fine with me.
> 
> @goneall , your thoughts?

I'll go with Alexios on this. I have a feeling once we publish the SPDX 3.0 spec in the form of HTML, the search results will point to that rather than the model directory.

VenkatTechnologist:
> > > No, please let's keep the `model` directory with actual files of the model, automatically processed. People should not be looking at the directory via GitHub interface, so the proposed `README` will not be automatically visible.
> > > No objections to document it in the top-level `README`.
> > 
> > 
> > Thanks for your input, @zvr . I'm okay either way - with having the model sub-directories related information in the main level README or in the model directory. Just fyi, when I search for SPDX 3.0 in the Internet, the second top page in the results happens to be the github link of the model, so I think people will land at model often; in fact, I did :). But if you feel strongly that model directory should have files only for processing, that's fine with me.
> > @goneall , your thoughts?
> 
> I'll go with Alexios on this. I have a feeling once we publish the SPDX 3.0 spec in the form of HTML, the search results will point to that rather than the model directory.

Alrighty. So, I will remove this file from model directory and add in the model directory content details in the README of the main directory. Please let me know if this is a satisfactory resolution. Thanks.

goneall:
> Alrighty. So, I will remove this file from model directory and add in the model directory content details in the README of the main directory. Please let me know if this is a satisfactory resolution. Thanks.

Sounds good to me.  That's the solution I was thinking of.

PR #602, opened 2024-01-15
improve software integrity and verification documentation

jeff-schutt:
Improves documentation in the SPDX 3 model to direct users towards existing support for package verification.

RE: https://github.com/spdx/spdx-3-model/issues/345#issuecomment-1889843573

CC: @goneall

### DIFF
--- a/model/Core/Classes/Hash.md
+++ b/model/Core/Classes/Hash.md
@@ -8,12 +8,11 @@ A mathematically calculated representation of a grouping of data.
 
 ## Description
 
-A hash is a grouping of characteristics unique to the result
-of applying a mathematical algorithm
-that maps data of arbitrary size to a bit string (the hash)
-and is a one-way function, that is,
-a function which is practically infeasible to invert.
-This is commonly used for integrity checking of data.
+A hash is a grouping of characteristics unique to the result of applying a mathematical algorithm
+that maps data of arbitrary size to a bit string (the hash) and is a one-way function, that is,
+a function which is practically infeasible to invert. This is commonly used for integrity checking of data.
+
+The recommended method to verify the integrity of `SoftwareArtifacts` Elements (including `Files`, `Snippets`, and `Packages`) is to use the SoftwareArtifact’s `contentIdentifier` property. 
 
 ## Metadata
 
--- a/model/Core/Classes/IntegrityMethod.md
+++ b/model/Core/Classes/IntegrityMethod.md
@@ -13,6 +13,8 @@ of a specific Element that correlates to the data in this SPDX document. This id
 a recipient to determine if anything in the original Element has been changed and eliminates
 confusion over which version or modification of a specific Element is referenced.
 
+The recommended method to verify the integrity of `SoftwareArtifacts` Elements (including `Files`, `Snippets`, and `Packages`) is to use the SoftwareArtifact’s `contentIdentifier` property. 
+
 ## Metadata
 
 - name: IntegrityMethod
--- a/model/Core/Properties/verifiedUsing.md
+++ b/model/Core/Properties/verifiedUsing.md
@@ -10,6 +10,8 @@ Provides an IntegrityMethod with which the integrity of an Element can be assert
 
 VerifiedUsing provides an IntegrityMethod with which the integrity of an Element can be asserted.
 
+The recommended method to verify the integrity of `SoftwareArtifacts` Elements (including `Files`, `Snippets`, and `Packages`) is to use the SoftwareArtifact’s `contentIdentifier` property.
+
 ## Metadata
 
 - name: verifiedUsing
--- a/model/Software/Properties/contentIdentifier.md
+++ b/model/Software/Properties/contentIdentifier.md
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provides a place to record the canonical, unique, immutable identifier for each software artifact using the artifact's gitoid.
+Used by SPDX producers to record the artifact’s gitoid: a canonical, unique, immutable identifier that can be used for software integrity verification.
+
+Used by SPDX consumers to verify the integrity of a software artifact they received.
 
 ## Description
 
-The contentIdentifier provides a canonical, unique, immutable artifact identifier for each software artifact. SPDX 3.0 describes software artifacts as Snippet, File, or Package Elements. The ContentIdentifier can be calculated for any software artifact and can be recorded for any of these SPDX 3.0 Elements using Omnibor, an attempt to standardize how software artifacts are identified independent of which programming language, version control system, build tool, package manager, or software distribution mechanism is in use.  
+### SPDX Producers
+The contentIdentifier is a canonical, unique, immutable artifact identifier for each software artifact. The ContentIdentifier for any software artifact can be calculated and recorded in SPDX 3.0 Snippet, File, or Package Elements. For additional information, see [OmniBOR](https://omnibor.io): an attempt to standardize how software artifacts are identified independent of which programming language, version control system, build tool, package manager, or software distribution mechanism is in use.  
 
 The contentIdentifier is defined as the [Git Object Identifier](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) (gitoid) of type `blob` of the software artifact. The use of a git-based version control system is not necessary to calculate a contentIdentifier for any software artifact.
 
@@ -18,7 +21,12 @@ The gitoid is expressed in the ContentIdentifier property by using the IANA [git
 Scheme syntax: gitoid":"<git object type>":"<hash algorithm>":"<hash value>
 ```
 
-The OmniBOR ID for the OmniBOR Document associated with a software artifact should not be recorded in this field. Rather, OmniBOR IDs should be recorded in the SPDX Element's ExternalIdentifier property. See [https://omnibor.io](https://omnibor.io) for more details.
+The OmniBOR ID for the OmniBOR Document associated with a software artifact must NOT be recorded in this field. Rather, OmniBOR IDs should be recorded in the SPDX Element's ExternalIdentifier property. See [https://omnibor.io](https://omnibor.io) for more details.
+
+### SPDX Consumers
+The integrity of software objects can be verified by calculating the gitoid(s) (`git hash-object foo`) of the object(s) and comparing the results to the value stored in the SPDX contentIdentifier field. ContentIdentifiers are canonical: Omnibor specifies a reproducible algorithm for anyone with the software object to perform this calculation. ContentIdentifiers are unique: the gitoid value is the result of a specific implementation of a one-way hash function. If the calculated gitoid value is the same as the gitoid value stored in SPDX, you can be sure it’s the same software. ContentIdentifiers are immutable: if a software object changes the resulting contentIdentifier will differ. These properties enable the verification of software integrity between producer and consumer using SPDX.
+
+
 
 ## Metadata
 
--- a/model/Software/Properties/contentIdentifier.md
+++ b/model/Software/Properties/contentIdentifier.md
@@ -6,7 +6,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 Used by SPDX producers to record the artifact’s gitoid: a canonical, unique, immutable identifier that can be used for software integrity verification.
 
-Used by SPDX consumers to verify the integrity of a software artifact they received.
+Used by SPDX consumers to verify the identity and integrity of a software artifact they received.
 
 ## Description
 
--- a/model/Software/Properties/contentIdentifier.md
+++ b/model/Software/Properties/contentIdentifier.md
@@ -10,8 +10,7 @@ Used by SPDX consumers to verify the identity and integrity of a software artifa
 
 ## Description
 
-### SPDX Producers
-The contentIdentifier is a canonical, unique, immutable artifact identifier for each software artifact. The ContentIdentifier for any software artifact can be calculated and recorded in SPDX 3.0 Snippet, File, or Package Elements. For additional information, see [OmniBOR](https://omnibor.io): an attempt to standardize how software artifacts are identified independent of which programming language, version control system, build tool, package manager, or software distribution mechanism is in use.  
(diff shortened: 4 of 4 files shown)


model/Software/Properties/contentIdentifier.md
```diff
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provides a place to record the canonical, unique, immutable identifier for each software artifact using the artifact's gitoid.
+Used by SPDX producers to record the artifact’s gitoid: a canonical, unique, immutable identifier that can be used for software integrity verification.
+
+Used by SPDX consumers to verify the integrity of a software artifact they received.
```
goneall, line 9:
```suggestion
Used by SPDX consumers to verify the identity and integrity of a software artifact they received.
```

goneall (changes requested):
One suggestion and one question for @zvr on spec parser support for the subheadings.

model/Software/Properties/contentIdentifier.md
```diff
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Description
 
-The contentIdentifier provides a canonical, unique, immutable artifact identifier for each software artifact. SPDX 3.0 describes software artifacts as Snippet, File, or Package Elements. The ContentIdentifier can be calculated for any software artifact and can be recorded for any of these SPDX 3.0 Elements using Omnibor, an attempt to standardize how software artifacts are identified independent of which programming language, version control system, build tool, package manager, or software distribution mechanism is in use.  
+### SPDX Producers
```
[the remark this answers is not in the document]
goneall: It looks like the CI failed in the spec parser.  I can't be sure, but I strongly suspect it is the `###`.  @jeff-schutt - you may want to change this to a plain text format like `For SPDX producers: `.

model/Software/Properties/contentIdentifier.md
```diff
@@ -18,7 +21,12 @@ The gitoid is expressed in the ContentIdentifier property by using the IANA [git
 Scheme syntax: gitoid":"<git object type>":"<hash algorithm>":"<hash value>
 ```
 
-The OmniBOR ID for the OmniBOR Document associated with a software artifact should not be recorded in this field. Rather, OmniBOR IDs should be recorded in the SPDX Element's ExternalIdentifier property. See [https://omnibor.io](https://omnibor.io) for more details.
+The OmniBOR ID for the OmniBOR Document associated with a software artifact must NOT be recorded in this field. Rather, OmniBOR IDs should be recorded in the SPDX Element's ExternalIdentifier property. See [https://omnibor.io](https://omnibor.io) for more details.
+
+### SPDX Consumers
```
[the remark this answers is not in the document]
goneall: Same comment as above

kestewart:
Bob and Dick to take a pass through this, per call today.

model/Software/Properties/contentIdentifier.md
bobmartin3000: makes sense

model/Software/Properties/contentIdentifier.md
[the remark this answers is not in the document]
bobmartin3000: This is clearer.

rjb4standards:
I concur with Gary's proposed language. It may be worth noting that component ordering can effect OmniBor results. SBOM's have no implied ordering requirement and may not be easily "verifiable" using OmniBor due to this mismatch in ordering requirements.

model/Software/Properties/contentIdentifier.md
```diff
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provides a place to record the canonical, unique, immutable identifier for each software artifact using the artifact's gitoid.
+Used by SPDX producers to record the artifact’s gitoid: a canonical, unique, immutable identifier that can be used for software integrity verification.
```
zvr, line 7: Obviously is not only a "gitoid", but any content identifier.

zvr, line 7:
```suggestion
Used to record a canonical, unique, immutable identifier of the artifact, that can be used for verification its identity and integrity.
```

model/Software/Properties/contentIdentifier.md
zvr, line 9:
```suggestion
```
No need to have it separate for producers/consumers.

model/Software/Properties/contentIdentifier.md
```diff
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
...
+### SPDX Producers
+The contentIdentifier is a canonical, unique, immutable artifact identifier for each software artifact. The ContentIdentifier for any software artifact can be calculated and recorded in SPDX 3.0 Snippet, File, or Package Elements. For additional information, see [OmniBOR](https://omnibor.io): an attempt to standardize how software artifacts are identified independent of which programming language, version control system, build tool, package manager, or software distribution mechanism is in use.  
 
 The contentIdentifier is defined as the [Git Object Identifier](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects) (gitoid) of type `blob` of the software artifact. The use of a git-based version control system is not necessary to calculate a contentIdentifier for any software artifact.
```
zvr, line 16: I disagree that `contentIdentifier` has to be a gitoid. We should leave it for any kind of unique identifier.

zvr (changes requested):

model/Software/Properties/contentIdentifier.md
rjb4standards: It may be more appropriate to include an identifier tag in the syntax, like IANA uses to qualify identifier types - similar to what SPDX does today : dns:www.foo.com  SHA256:.......

model/Software/Properties/contentIdentifier.md
goneall: Note that not all content identifiers support use in verification.  We would also need to somehow specify which content identifier is used (e.g. the enumeration of supported content identifiers).

rjb4standards: That appears similar to how SPDX handles this today with the SHA* identifying tags accompanying a hash value in an SBOM

model/Software/Properties/contentIdentifier.md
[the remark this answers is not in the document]
jeff-schutt: I've made this change to the headings.

model/Software/Properties/contentIdentifier.md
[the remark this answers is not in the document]
jeff-schutt: I've made this change to the headings.

model/Software/Properties/contentIdentifier.md
goneall: @zvr - Let's move forward with this definition and open a separate issue to include better support for SWHID after RC2 - just trying to get RC2 released.

goneall (approved):
With the updated changes - all looks good

model/Software/Properties/contentIdentifier.md
kestewart: There are other content identifiers out there, as Alexios points out.    Let's handle the broadening as a separate PR though, and move this forward.    The original had gitoid in it, and is just restructured here.

bobmartin3000:
Agree that the changes addressed the concerns for RC2.

kestewart (approved):

model/Software/Properties/contentIdentifier.md
zvr: We can remove gitoid and leave generic "unique, immutable artifact identifier" or we can mention both gitoid and swhid that we already know of.

zvr: I'm not objecting just because I don't want SWHIDs to be forgotten:
The most important part is, given that we know there might be more than one format, modeling it as a single value (of type URI) will not work. There has to be a type defined somewhere (maybe with a vocabulary), etc.

zvr (changes requested):

goneall:
> I'm not objecting just because I don't want SWHIDs to be forgotten:
The most important part is, given that we know there might be more than one format, modeling it as a single value (of type URI) will not work. There has to be a type defined somewhere (maybe with a vocabulary), etc.

@zvr - Since the `contentIdentifier` property defined as a GitOID has been there for a while and making this field accept other formats and/or changing the model for `verifiedUsing` to a yet to be specified alternative will take some time - do you think we should delay RC2 to resolve this or should we merge this in and discuss as a post RC2 issue?

zvr:
This should be resolved before RC2, since it most probably _will_ result in changes in the model.
If it was only the description text, we could leave it.

zvr:
In an effort to resolve the different discussions here, I've broken down the different parts and submitted a number of PRs:

* #609, to add the wording to the classes and property of Core

and then, an option of two approaches:

* #610 makes `gitoid` a property to `SoftwareArtifact` (and we can add others later); **or**
* #611 to make `contentIdentifier` more general, supporting different types by adding properties and a vocabulary

Note that there is no mention of "preferred", "recommended", "suggested", etc. anywhere. Policy is something different than model definition and can be decided later -- and expressed with various conformance tests, profile definitions, etc.

@goneall , @jeff-schutt , @kestewart and everyone else, please review these new PRs.

goneall:
> In an effort to resolve the different discussions here, I've broken down the different parts and submitted a number of PRs:
> 
> * [Updates integrity verification docs in Core #609](https://github.com/spdx/spdx-3-model/pull/609), to add the wording to the classes and property of Core
> 
> and then, an option of two approaches:
> 
> * [Use gitoid for SoftwareArtifact integrity verification #610](https://github.com/spdx/spdx-3-model/pull/610) makes `gitoid` a property to `SoftwareArtifact` (and we can add others later); **or**
> * [Generalize contentIdentifier for SoftwareArtifact integrity verification #611](https://github.com/spdx/spdx-3-model/pull/611) to make `contentIdentifier` more general, supporting different types by adding properties and a vocabulary
> 
> Note that there is no mention of "preferred", "recommended", "suggested", etc. anywhere. Policy is something different than model definition and can be decided later -- and expressed with various conformance tests, profile definitions, etc.
> 
> @goneall , @jeff-schutt , @kestewart and everyone else, please review these new PRs.

Thanks @zvr - having the specific pull requests definitely makes it easier to compare / review / decide.

I'm OK with either of your two choices in the model changes.  I'm also OK with the current state, but you do raise a good point about `contentIdentifier` as a name implies a broader range than gitoid, which leans me towards changing the property name as the easiest path forward.

I do like PR #602 as a better starting point for the documentation for reasons noted in the review comments.  We should probably decide on the 3 choices of the model -1) keep as is; 2) PR #610 rename the property or 3) PR #611 generalize `contentIdentifier`.

@jeff-schutt @zvr Since I'm OK with all three approaches, perhaps you can connect and decide which works OK for both of you.

model/Software/Properties/contentIdentifier.md
sbarnum: My recollection of the discussions on this topic from last year were that for v3.0 we would specify gitoid as the default but that the structures and definitions would be generalized (as @zvr is suggesting above) to enable addition of other possible types post v3.0 without breaking backward compatibility.

In other words, I agree that the definitions should be generalized and not locked to one approach.

model/Core/Classes/Hash.md
```diff
@@ -8,12 +8,11 @@ A mathematically calculated representation of a grouping of data.
...
-of applying a mathematical algorithm
-that maps data of arbitrary size to a bit string (the hash)
-and is a one-way function, that is,
-a function which is practically infeasible to invert.
-This is commonly used for integrity checking of data.
+A hash is a grouping of characteristics unique to the result of applying a mathematical algorithm
+that maps data of arbitrary size to a bit string (the hash) and is a one-way function, that is,
+a function which is practically infeasible to invert. This is commonly used for integrity checking of data.
+
+The recommended method to verify the integrity of `SoftwareArtifacts` Elements (including `Files`, `Snippets`, and `Packages`) is to use the SoftwareArtifact’s `contentIdentifier` property.
```
sbarnum, line 15: I do not think that these sort of case specific details should be placed here.
This is the general structure to be used across SPDX.
Placing case specific variations here is likely to lead to an eventual laundry list of such specializations.

I propose that it would be more appropriate to add this note to the markdown files for the case specific structures (SoftwareArtifact, File, Package, Snippet).

model/Core/Classes/IntegrityMethod.md
```diff
@@ -13,6 +13,8 @@ of a specific Element that correlates to the data in this SPDX document. This id
 a recipient to determine if anything in the original Element has been changed and eliminates
 confusion over which version or modification of a specific Element is referenced.
 
+The recommended method to verify the integrity of `SoftwareArtifacts` Elements (including `Files`, `Snippets`, and `Packages`) is to use the SoftwareArtifact’s `contentIdentifier` property.
```
sbarnum, line 16: I do not think that these sort of case specific details should be placed here.
This is the general structure to be used across SPDX.
Placing case specific variations here is likely to lead to an eventual laundry list of such specializations.

I propose that it would be more appropriate to add this note to the markdown files for the case specific structures (SoftwareArtifact, File, Package, Snippet).

model/Core/Properties/verifiedUsing.md
```diff
@@ -10,6 +10,8 @@ Provides an IntegrityMethod with which the integrity of an Element can be assert
 
 VerifiedUsing provides an IntegrityMethod with which the integrity of an Element can be asserted.
 
+The recommended method to verify the integrity of `SoftwareArtifacts` Elements (including `Files`, `Snippets`, and `Packages`) is to use the SoftwareArtifact’s `contentIdentifier` property.
```
sbarnum, line 13: I do not think that these sort of case specific details should be placed here.
This is the general structure to be used across SPDX.
Placing case specific variations here is likely to lead to an eventual laundry list of such specializations.

I propose that it would be more appropriate to add this note to the markdown files for the case specific structures (SoftwareArtifact, File, Package, Snippet).

goneall:
This is a bit stale with the renaming to gitoid - retargeting to post RC2 for discussion

goneall:
This is now quite stale and there are other PR's which overlap.

I'm going to go ahead and close this PR.

@jeff-schutt @zvr - Please feel free to open a fresh PR if you think the documentation needs to be updated.

PR #603, opened 2024-01-15
Remove hash algorithms with no description

goneall:
The hash algorithms have only "TODO" for comments. I could not find any record of these in the minutes and no one has volunteered to fill in the description, so this PR deletes these algorithms from the model.

If anyone would like to add them back, please add them with a description as a separate pull request.


kestewart (changes requested):
For PVC - Package Verification Code include both SHA1 & SHA256.

goneall:
> For PVC - Package Verification Code include both SHA1 & SHA256.

So these are for implementing the package verification code?  If so, we should remove them for now as it conflicts with another PR and we haven't decided if PVC is needed.  Something we can do after RC2.

@kestewart - let me know if you agree we can delete these for now and consider adding them back (with a description) after RC2.

maxhbr (approved):
lgtm

PR #604, opened 2024-01-17
Validating PRs via new parser

zvr:
Adds an action to validate proposed PRs by using the new spec-parser

Presumably the previous workflow can be deleted, but this is not done here.


goneall (approved):

goneall:
@zvr Thanks for adding this.  Do we want to create a separate PR to fix or remove the other action or add it to this PR?

PR #605, opened 2024-01-23
run shacl2code in ci

maxhbr:
The script is from https://github.com/JPEWdev/shacl2code/blob/main/pyproject.toml

this should place the generated code in `./shacl2code` in gh-pages and attach it as an artifact to the ci run.

### DIFF
--- a/.github/workflows/spec-parser.yml
+++ b/.github/workflows/spec-parser.yml
@@ -59,6 +59,17 @@ jobs:
           name: context.json
           path: /tmp/spec-parser.out/context.json
 
+      - name: run JPEWdev/shacl2code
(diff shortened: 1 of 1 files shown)


goneall:
We probably should wait to merge this until after the other action PR's are resolved

kestewart:
Need to get update from Jeff Licquia; before this one can move forward.

kestewart:
@JPEWdev to review.

maxhbr:
I think this is stale and outdated as there seems to be the decision to throw generated output into another repository.

goneall:
Closing as it is now outdated

PR #606, opened 2024-01-23
GitHub actions improvements

puerco:
This PR is a follow-up to #604 to update the workflows improving the way actions are setup in the repo, the summary of the changes is the following:

* Actions are now pinned to hashes instead of tags for security purposes
* Fixes a bug on the validation workflow trigger
* ~Simplifies the artifact upload in spec-parser to upload everything in one go~
* Dropped the hard coded `/tmp` path to use the runner-specific temp directory
* ~Bump python version in spec-parser to 3.12 to unify it with the version used in valida-pr~
* Adds a basic dependabot config to keep the actions up to date
* Normalizes the license boiler plates of the code in the actions config
* ~Fixes a bug in the directory where the spec parser was being cloned~
* ~Fixes the invocation of the parser in `spec-parser`~

**Update:** This PR now deletes the entire `spec-parser` workflow as per the discussion below.

/cc @zvr @goneall 

Signed-off-by: Adolfo García Veytia (Puerco) <adolfo.garcia@uservers.net>


maxhbr:
the file [.github/workflows/spec-parser.yml](https://github.com/spdx/spdx-3-model/pull/606/files#diff-da81fe6083fdcf8fe39a61ebb4257c51c22a53694b4933938ae2ebbd3a199f27) is about to be deleted. Your changes will soon conflict

puerco:
@maxhbr do you want me to delete it in this PR?

puerco:
Opened https://github.com/spdx/spec-parser/pull/90 to fix an unhandled attribut error breaking the check

maxhbr:
> @maxhbr do you want me to delete it in this PR?

I think that this was the agreement. But it is sad to loose the ontospy generation and the other stuff. Maybe we need to create a ticket to bring that back.

puerco:
The workflow has been broken for a while and hasn't run for a while (see [this run](https://github.com/spdx/spdx-3-model/actions/runs/7631558311/job/20789764251) for example). I'll fix it so that it can start pushing updates again.

zvr:
@puerco , I did #604 for a workflow, and that file should be the only one remaining.

Remember, the only action needed is to check the PRs before merging. All generations happen in the spdx-spec repo.

.github/workflows/validate-pr.yml
```diff
@@ -1,21 +1,24 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright 2024 The SPDX Contributors
+
 name: validate PR
 
 on:
-  pull-request
+  pull_request:
```
zvr, line 7: I don't think the final `:` is actually needed, if there is nothing underneath in inner levels.

zvr (changes requested):
Remove the old files and only keep the new workflow.

puerco:
@zvr got it. The question is should we keep the ontospy generation jobs?

If yes, I can fix the current workflow that is broken.

If not, I can simply delete the spec-parser job (keeping only validate-pr).

zvr:
No, it will run on the spdx-spec repo and publish the results in rdf.spdx.org (or whatever our URI will be).

maxhbr:
> No, it will run on the spdx-spec repo and publish the results in rdf.spdx.org (or whatever our URI will be).

I disagree slightly and think that some artifacts which are purely built from the model should be generated in the spdx-3-model. But that is more relevant for the shacl and related stuff. Ontospy is an edge case and maybe soon no longer helpful.

.github/workflows/spec-parser.yml
```diff
@@ -17,11 +20,11 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@0a5c61591373683505ea898e09a3ea4f39ef2b9c # v5.0.0
         with:
-          python-version: "3.11"
+          python-version: "3.12"
       - name: get spec-parser
```
goneall, line 28: This section where we run the spec-parser should be deleted - the current output formats don't currently work with the latest version.

@licquia will add this back to the spec-parser repo

.github/workflows/spec-parser.yml
```diff
@@ -31,33 +34,21 @@ jobs:
       - name: run spec-parser
```
goneall, line 34: Same with this section

.github/workflows/spec-parser.yml
```diff
@@ -31,33 +34,21 @@ jobs:
...
-              --gen-rdf \
-              --out-dir /tmp/spec-parser.out \
-              ../model
+          mkdir -p  ${{ runner.temp }}/spec-parser.out
+          python main.py ../model ${{ runner.temp }}/spec-parser.out 
 
       - name: validate that context is valid
         run: |
           python3 -m pip install rdflib
-          python3 ./.github/workflows/test_context.py /tmp/spec-parser.out/context.json
```
goneall, line 45: I guess we have to remove this as well

.github/workflows/spec-parser.yml
```diff
@@ -31,33 +34,21 @@ jobs:
...
-          name: model.ttl
-          path: /tmp/spec-parser.out/model.ttl
-
-      - uses: actions/upload-artifact@v3
-        with:
-          name: model.jsonld
-          path: /tmp/spec-parser.out/model.jsonld
+          python3 ./.github/workflows/test_context.py ${{ runner.temp }}/spec-parser.out/context.json
 
-      - uses: actions/upload-artifact@v3
```
goneall, line 57: And remove this

.github/workflows/spec-parser.yml
```diff
@@ -68,28 +59,28 @@ jobs:
       - name: generate docs with Ontospy
         run : |
           cd Ontospy
-          mkdir -p /tmp/auto-generated
```
goneall, line 71: I think we should keep this for now until we get the spdx-spec repo up and running

goneall (changes requested):
Let's keep the onotospy for now - no need to remove it until we get everything setup in the new spdx-spec repo.

@licquia - you may want to review updated spec-parser.yml file that @puerco created for us - it has some security improvements we should carry forward to the spdx-spec repo actions.

maxhbr:
> Let's keep the onotospy for now - no need to remove it until we get everything setup in the new spdx-spec repo.


@goneall : we can't keep it, when we drop the generation. To keep that we would need to pin the spec-parser in the old action to an old version.

maxhbr:
I think that this PR should for now drop `.github/workflows/spec-parser.yml` and lets create a Ticket to rebuild its functionality

goneall:
> I think that this PR should for now drop .github/workflows/spec-parser.yml and lets create a Ticket to rebuild its functionality

Agree - @puerco - if you could make that change, I can do a quick final review and merge

puerco:
OK, sounds good :+1: 

I've deleted the spec-parser workflow in a new commit. This should let us rescue the fixed version if we need it.

puerco:
I think we should delete `test_context.py` as well ?

maxhbr:
> I think we should delete `test_context.py` as well ?

lets ignore that file for now and just get it merged. I think now everything is green -> merge

maxhbr (approved):
LGTM

goneall (approved):
LGTM

PR #611, opened 2024-01-26
Generalize contentIdentifier for SoftwareArtifact integrity verification

zvr:
Adds a general contentIdentifier type system to support various types.

Note that the actual value of an identifier (`contentIdentifierValue` in a `ContentIdentifier`) is defined to be of type `xsd:string`. The two current available types both produce `xsd:anyURI` but maybe we want to keep our options open for the future.

The description of the types was copied from `/Core/ExternalIdentifierType`; I simplifies the `swhid` entry since we only consider content in this context.

### DIFF
--- a/model/Software/Classes/ContentIdentifier.md
+++ b/model/Software/Classes/ContentIdentifier.md
@@ -1,0 +1,30 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# ContentIdentifier
+
+## Summary
+
+A canonical, unique, immutable identifier
+
+## Description
+
+A ContentIdentifier is a canonical, unique, immutable identifier of the content of a software artifact, such as a package, a file, or a snippet.
+It can be used for verifying its identity and integrity.
+
+## Metadata
+
+- name: ContentIdentifier
+- SubclassOf: /Core/IntegrityMethod
+- Instantiability: Concrete
+
+## Properties
+
+- contentIdentifierValue
+  - type: xsd:string
+  - minCount: 1
+  - maxCount: 1
+- contentIdentifierType
+  - type: ContentIdentifierType
+  - minCount: 1
+  - maxCount: 1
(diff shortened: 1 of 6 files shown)


goneall (commented):
Overall, this looks like a consistent approach to contentIdentifier.  The only downside I can see over the current state is with a choice, consumers of the SBOMs would need to be able to calculate all of the (two) methods to verify since we wouldn't know which one is provided.

One possible solution to the above downside would be to create an "Integrity" profile which specifies the property as required and limits the range to only one value.  This could be a post RC2 improvement.

sbarnum:
If the presumption that content identifiers can ALWAYS be reduced to a single serialized string then this approach seems to be the best solution to this issue.

goneall:
based on discussions in the tech call and in the security meeting week of 12 Feb, we are going to go with the gitoid renaming for RC2 and retarget the more general discussion for after RC2.

Primarily due to the amount of thought and review needed for the more general approach and lack of availability of key individuals for discussion.

kestewart:
@zvr - is this stale?  or something still being requested.

kestewart:
@zvr to fix up conflicts,  but we need discussion with Sean & Jeff before merging.

sbarnum (commented):
I believe this PR has been overcome by other decisions.

While significant discussion occurred on this topic which resulted in a modified version of the concept in this PR that was clear, simple and preferred by the majority of those involved in the discussions a decision was made to add an explicitly dedicated 'gitoid' property to SoftwareArtifact for SPDX 3.0.

I believe this PR should be closed without approval.

zvr:
I will re-work it and probably make a new one without conflicts, but there was never any decision that only `gitoid` will be the verification mechanism.
We had it for RC2 and left the discussion for 3.0.

kestewart:
See #701

PR #616, opened 2024-01-29
Fix delegatedTo Relationship

JPEWdev:
In practice, the definition of the delegatedTo relationship was flawed. Since it linked one agent to another, it was impossible to determine which delegatedTo relationships applied to which invokedBy relationships.

To solve this, reword the delegatedTo relationship to indicated that the `to` end must point to the specific invokedBy relationships to which it applies.

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -32,7 +32,7 @@ name completes the sentence:
 - contains: The `from` Element contains each `to` Element
 - coordinatedBy: (Security) The `from` Vulnerability is coordinatedBy the `to` Agent(s) (vendor, researcher, or consumer agent)
 - copiedTo: The `from` Element has been copied to each `to` Element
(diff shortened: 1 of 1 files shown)


JPEWdev:
It should be noted that this is an unorthodox relationship in that it relates an Actor to another Relationship, which IIRC has not been done before. However, I believe in this context is correctly conveys the information required.

sbarnum:
I have no issue with constraining the `to` or `from` of a Relationship to another Relationship. You are correct that this is completely valid to do.

I am finding it hard to follow the scenario and intent here proposed in this PR.
I think this would benefit very well from a simple diagram of an example.

JPEWdev:
Here is an example that hopefully explains it a little better: https://drive.google.com/file/d/1F6MiE48PAuz8uhk8Jy-g2jCOaAdVolFX/view?usp=sharing

kestewart (approved):
Thanks for providing the diagram Josh, that helped explain it.   Change looks fine to me.

PR #619, opened 2024-01-29
rename begin and end to beginIntegerRange and endIntegerRange

### LINKED ISSUE #567: The "begin" and "end" properties on PositiveIntegerRange are far to generically named and will almost certainly cause semantic collision. They need renamed something more specific.
The PositiveIntegerRange class currently defines an integer range using a "begin" and an "end" property both of datatype positiveInteger.
These names are far to generic for use within the model.
The concept of 'begin' and 'end' could be desired/needed to characterize time windows, substrings, memory regions, etc.
Any other use would conflict with the current properties given their specific typing.
To avoid such semantic collisions these two properties should simply be renamed "beginIntegerRange" and "EndIntegerRange".

goneall:
Fixes #567

### DIFF
--- a/model/Core/Classes/PositiveIntegerRange.md
+++ b/model/Core/Classes/PositiveIntegerRange.md
@@ -9,7 +9,7 @@ A tuple of two positive integers that define a range.
 ## Description
 
 PositiveIntegerRange is a tuple of two positive integers that define a range.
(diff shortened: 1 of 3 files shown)


kestewart (approved):
Thanks Gary,  making this explicit will avoid problems.

PR #620, opened 2024-01-29
Change range of issueingAuthority from anyURI to string

### LINKED ISSUE #568: The "issuingAuthority" property on the ExternalIdentifier class should be of type String rather than anyURI.
The "issuingAuthority" property on the ExternalIdentifier class is currently typed as anyURI.
However many if not most issuing authorities for concept identifiers are likely to be general strings such as organization names (IETF, NIST, etc.) rather than URIs.
Changing the type of "issuingAuthority" to String would solve for the general case and also support specifiying URIs as strings.
This type should be changed to String in the spec.

goneall:
Fixes #568

### DIFF
--- a/model/Core/Classes/ExternalIdentifier.md
+++ b/model/Core/Classes/ExternalIdentifier.md
@@ -34,6 +34,6 @@ that uniquely identifies an Element.
   - type: xsd:anyURI
   - minCount: 0
 - issuingAuthority
-  - type: xsd:anyURI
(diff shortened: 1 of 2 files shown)


kestewart (approved):
Change looks reasonable based on the discussion.  URI would be machine readable, but if it's going to cause too many problems, per the issue.   This should sort it.

zvr (approved):

PR #621, opened 2024-01-29
Change MediaType to conform to RFC 2046

### LINKED ISSUE #569: Defined string constraint for MediaType simple data type should reference RFC2046 rather than RFC6838.
The string constraint for the MediaType simple data type was originally specified as conformance to RFC2046 (the IETF standard for Media Types).
At some point it was changed to be conformance to RFC6838 (the IETF standard for Media Type Specifications and Registration Procedures).
RFC6838 is a procedural standard rather than a structural or representation standard like RFC2046 is.
RFC6838 is NOT intended to cover the same ground as RFC2046 or replace it in any way.
RFC6838 is a relatively new standard (published Jan 2023). It does describe how to specify new media types to support intended registration. However, the format constraints of RFC6838 are narrower than RFC2046 meaning that defining the MediaType simple data type in SPDX 3.0 as conformance to RFC6838 would invalidate many existing media types that are conformant to RFC2046 but not to RFC6838.

RFC2046 is the appropriate standard to conform to for the intended purpose of the MediaType simple data type in SPDX.

goneall:
Fixes #569

This basically reverts the MediaType back to the definition before commit cd13828

### DIFF
--- a/model/Core/Datatypes/MediaType.md
+++ b/model/Core/Datatypes/MediaType.md
@@ -4,14 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Standardized way of indicating the type of content of an Element. A String constrained to the RFC 6838 specification.
+Standardized way of indicating the type of content of an Element. A String constrained to the RFC 2046 specification.
 
 ## Description
 
-A MediaType is a string constrained to the RFC 6838 specification, and includes
-`type-name`, `subtype-name` and optional `parameters`.
-It provides a standardized way of indicating the type of content of an Element.
-A list of all possible media types is available at https://www.iana.org/assignments/media-types/media-types.xhtml.
+A MediaType is a string constrained to the RFC 2046 specification. It provides a standardized
+way of indicating the type of content of an Element.
 
 ## Metadata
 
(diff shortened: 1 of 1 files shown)


model/Core/Datatypes/MediaType.md
```diff
@@ -4,14 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
-Standardized way of indicating the type of content of an Element. A String constrained to the RFC 6838 specification.
+Standardized way of indicating the type of content of an Element. A String constrained to the RFC 2046 specification.
 
 ## Description
 
-A MediaType is a string constrained to the RFC 6838 specification, and includes
-`type-name`, `subtype-name` and optional `parameters`.
-It provides a standardized way of indicating the type of content of an Element.
-A list of all possible media types is available at https://www.iana.org/assignments/media-types/media-types.xhtml.
```
kestewart, line 14: Any reason why the list of possible media types is not being carried forward?    Looks like useful information to me.

goneall: I just copied over the previous commit which refers to RFC 2046 - see issue #569 for the reason for this

kestewart: I understand the reversion, based on the issue history.    I just think that the phrase:  "A list of all possible media types is available at https://www.iana.org/assignments/media-types/media-types.xhtml."  is useful to add.  However, I'm fine with it being added later.

kestewart (approved):

model/Core/Datatypes/MediaType.md
goneall: Got it - I thought your comment was referring to a different line.  I'll add that in.

goneall:
Added the statement back.

@sbarnum - could you do a quick review?

PR #629, opened 2024-02-05
New Element Individuals to handle NONE and NOASSERTION scenarios for the 'to' property of Relationships

sbarnum:
Addresses #527

### DIFF
--- a/model/Core/Classes/Relationship.md
+++ b/model/Core/Classes/Relationship.md
@@ -16,7 +16,7 @@ should contain the 'NONE' individual and no other elements.
 A relationship that contains 'NONE' and other elements in the `to` property is not valid.
 
 To explicitly assert that no assertions are being made regarding the 
-existence of such relationships, the `to` property should contain the 'NOASSERTION' individual.
+existence of such relationships, the `to` property should also contain the 'NOASSERTION' individual.
 
 ## Metadata
 
--- a/model/Core/Classes/Relationship.md
+++ b/model/Core/Classes/Relationship.md
@@ -11,6 +11,12 @@ Describes a relationship between one or more elements.
 A Relationship is a grouping of characteristics unique to an assertion
 that one Element is related to one or more other Elements in some way.
 
+To explicitly assert that no such relationships exist, the `to` property 
+should contain the 'NONE' individual.
+
+To explicitly assert that no assertions are being made regarding the 
+existence of such relationships, the `to` property should contain the 'NOASSERTION' individual.
+
 ## Metadata
 
 - name: Relationship
@@ -25,7 +31,7 @@ that one Element is related to one or more other Elements in some way.
   - maxCount: 1
 - to
   - type: Element
-  - minCount: 0
+  - minCount: 1
 - relationshipType
   - type: RelationshipType
   - minCount: 1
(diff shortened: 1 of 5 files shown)


kestewart (approved):
Thanks for pulling this together Sean.  Looks good to me.

goneall (approved):
This looks good to me - The description of the NoAssertion captured all the cases I was looking for.

Thanks @sbarnum

model/Core/Classes/Relationship.md
```diff
@@ -11,6 +11,12 @@ Describes a relationship between one or more elements.
 A Relationship is a grouping of characteristics unique to an assertion
 that one Element is related to one or more other Elements in some way.
 
+To explicitly assert that no such relationships exist, the `to` property 
+should contain the 'NONE' individual.
```
maxhbr, line 15:
```suggestion
should contain the 'NONE' individual and no other elements.
A relationship that contains 'NONE' and other elements in the `to` property is not valid.
```

model/Core/Classes/Relationship.md
```diff
@@ -11,6 +11,12 @@ Describes a relationship between one or more elements.
...
+
+To explicitly assert that no assertions are being made regarding the 
+existence of such relationships, the `to` property should contain the 'NOASSERTION' individual.
```
maxhbr, line 18:
```suggestion
existence of such relationships, the `to` property should also contain the 'NOASSERTION' individual.
```

maxhbr (changes requested):
Added some more explanation, especially that `[Element1, NONE]` is not allowed. We should make that even more explicit so that validation also catches that.

There is also the case of a Relationship with `from`=NOASSERTION, `relationshipType`="ancestorOf", and `to`=Element1. This should also be explained or should be disallowed. Same for NONE.

model/Core/Classes/Relationship.md
sbarnum: This looks like a good added clarification.

model/Core/Classes/Relationship.md
sbarnum: This change does not look appropriate to me. Maybe I am misunderstanding its intent?
If the `to` property contains a value of NOASSERTION it should contain no other values similar to NONE.
If it does contain contain ANY other values then assertions are being made and NOASSERTION would not hold true.

I believe the clarification made for NONE ("and not other elements")should also be made here.

kestewart: Discssed with Max on call, and we'll not apply this suggestion from Max.

We'll go ahead and merge this as is now.

maxhbr:
I disagree that this is the right direction. I think a to relationship towards an empty list, together with relationship completeness, is a better way to express that information.

kestewart:
Discussion is we should document from migration guide, how to go this for 3.0.   Concern is this with completeness is going to cause more confusion than it resolves.   Let's document it, and then decide which PR to apply.   ( Documenting Empty list is ok with completeness indicator..  vs.  adding this in again)

goneall:
Decided to move this to post RC2 and go with the existing solution using the `completness` property for RC2.

kestewart:
Discussed 623 with @JPEWdev and agreed we're going to merge it.   We're going to raise this in next tech call.

goneall:
From tech call on 2 April 2024: agreed to rename `None` and `NoAssertion` to `NoneElement` and `NoAssertionElement` resp.

goneall (changes requested):

goneall (changes requested):
Per tech call on 2 April 2024 -

goneall (approved):

zvr (approved):

Issue #387, opened 2023-06-14
Standards Fields


matthewhjcrawford:
I think the following four fields should be considered when labelling a Standard

StandardName
StandardVersion
StandardURL
StandardUseCase 

Hopefully the first three are self explanatory, but the last one might need explaining. When a Standard is involved we should consider whether it is Implemented (i.e. company A uses it in their deliverable and ships to company B), or Supported (i.e. company A does not use it, but provides support for company B to maybe use it). The different use cases may have difference risks for company A and may carve the second option out of their indemnity where the first might not be. This distinction might be helpful when discussing Standards

goneall:
@matthewhjcrawford I could use a bit more context - can you include a reference to the property / class / vocabulary that should be updated?

A pull request would also provide the context.

goneall:
Gary to ask Matthew and @kestewart to see if we can resolve

goneall:
@matthewhjcrawford @kestewart - could you take a look at the request and see if a pull request could be created?

goneall:
Moving to 3.1 - discussed at Member Summit - this will be part of the newly created Operations Profile

goneall:
@kestewart to work on a draft for version 3.0 RC2

kestewart:
Talked to Matthew in Cambridge, and there's been some evolution on this.   I've got an email from him, and I'll put in the PR if he doesn't beat me to it.    They will all be optional properties - but will give us a better foundation than just a single string, and it's something that Armin wanted to see changed anyway.

kestewart:
Here's a recent example of what
StandardName: IEEE Std 754-2008 for Floating point  arthrimatic
StandardVersion: 2008
StandardSupplier: IEEE
StandardHomePage: IEEE.org

From discussion with Gary,  this needs to be a class rather than individual properties.

matthewhjcrawford:
Hopefully this is all self explanatory, let me know if you need any commentary on the fields.

kestewart:
@matthewhjcrawford - yes they are.   Gary and I discussed this in person last night, and were thinking that some of the agent info is probably appropriate for some of the properties.    To that end,  he's convinced me we should set up a class call "Standard" and then have StandardName and StandardVersion added now.   The other properties we can sort out adding in 3.1 if we can't use agents.     That work for you?

goneall:
[...]

PR #631, opened 2024-02-06
Change name of 'standard' property to be 'standardName'

sbarnum:
This PR implements the consensus decision of the Feb 06, 2024 Tech call to  address immediate concerns under Issue #387.

More structured approaches to fully resolving Issue #387 will be pursued post RC2.

### DIFF
--- a/model/Core/Classes/Artifact.md
+++ b/model/Core/Classes/Artifact.md
@@ -38,6 +38,6 @@ such as an electronic file, a software package, a device or an element of data.
   - type: DateTime
   - minCount: 0
   - maxCount: 1
-- standard
+- standardName
   - type: xsd:string
   - minCount: 0 
--- a/model/Core/Properties/standardName.md
+++ b/model/Core/Properties/standardName.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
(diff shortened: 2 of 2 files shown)


goneall (approved):
I agree with the solution as well - and the PR LGTM.

Thanks @sbarnum and the participants on the tech team call.

goneall (changes requested):
I just noticed that the CI is failing - there is a class referencing standard that also needs updating.  I believe it is `Artifact`.

@sbarnum - if you could add that change in as well.

maxhbr:
this fails with `AssertionError: Property name standard does not match metadata standardName`

maxhbr:
this also needs to be changed in:
- the title of the file
-  the places where it is used, e.g. in `.../Core/Classes/Artifact.md`
   - should the standard property in Artifact be using this?

goneall:
@sbarnum @maxhbr - Let me know if you'd like me to update the PR with the required changes

kestewart:
@goneall - discussed this in AI call, and we're good with it as is.   For future, Karen volunteers to coordinate with IEEE/ISO for future standard fields.

goneall (approved):
LGTM - CI failures are now fixed

kestewart (approved):
Thanks Gary for fixing the remaining refs.

PR #633, opened 2024-02-12
Initial updates for use case 14

goneall:
Describes service delivery dependencies


model/Service/Properties/provider.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Description
+
+The Agent that provides a SoftwareService to one or more consumers.
+
+## Metadata
+
+- name: provider
+- Nature: ObjectProperty
+- Range: ServiceProvider
```
maxhbr, line 17: is `ServiceProvider` already defined

goneall: Yes - it is already in the service-profile branch (previous commit)

maxhbr: Ah, I tried to grep for it and failed to find it.

goneall: You're probably on the wrong branch - I just grepped the `service-profile` branch with the following results:

```
$ egrep -ri serviceprovider ./
./Service/Classes/ServiceProvider.md:# ServiceProvider
./Service/Classes/ServiceProvider.md:A ServiceProvider is an entity that provide a software service.
./Service/Classes/ServiceProvider.md:- name: ServiceProvider
```

model/Service/Classes/SoftwareService.md
```diff
@@ -0,0 +1,24 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SoftwareService
```
nishakm, line 3: Why not just "Service"?

nishakm (commented):
I'm trying to figure out how to model a situation when a ServiceProvider IS AN Organization or IS A Person, which is what happens for the most part. Maybe include a property "operatedBy" which is a Person or Organization?

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -80,6 +80,7 @@ name completes the sentence:
 - republishedBy: (Security) Designates a `from` Vulnerability's details were tracked, aggregated, and/or enriched to improve context (i.e. NVD) by a `to` Agent(s)
 - serializedInArtifact: The `from` SPDXDocument can be found in a serialized form in each `to` Artifact
 - testedOn: (AI, Dataset) The `from` Element has been tested on the `to` Element
+- serviceDependsOn (Service) The `from` SoftwareService depends on each `to` SoftwareService.
```
nishakm, line 83: I'm wondering if `dependsOn` is enough to describe this relationship.

nishakm (commented):
One more thought :)

model/Service/Classes/SoftwareService.md
goneall: That was my original thought when I was writing this up.

After looking up different definitions of Software as a Service, I decided that we should qualify the name with "Software".

In the future, we could add other services that are not software if there ever is a use case.

That being said, I don't feel strongly about using SoftwareService over Service.

goneall:
> I'm trying to figure out how to model a situation when a ServiceProvider IS AN Organization or IS A Person, which is what happens for the most part. Maybe include a property "operatedBy" which is a Person or Organization?

Really good thought.  We could use a property as suggested.  Another thought is to use a relationship - the advantage of a relationship is it can be added later after the original service BOM is created.  Personally I'm leaning toward property, but I need to think about it some more.

model/Core/Vocabularies/RelationshipType.md
goneall: This would simplify the spec, so I'm +1 on just using `dependsOn`.

Looking at the definition of `dependsOn`: 

"The from Element depends on each to Element during a LifecycleScopeType period."

We may need to change the definition of `LifecycleScopeType` to include service related lifecycles, like deployed.  May be a good idea to do this anyway.

goneall: Discussed on 26 Feb 2024 - agreed to remove `serviceDependsOn` - `dependsOn` is enough.

goneall:
> > I'm trying to figure out how to model a situation when a ServiceProvider IS AN Organization or IS A Person, which is what happens for the most part. Maybe include a property "operatedBy" which is a Person or Organization?
> 
> Really good thought. We could use a property as suggested. Another thought is to use a relationship - the advantage of a relationship is it can be added later after the original service BOM is created. Personally I'm leaning toward property, but I need to think about it some more.

Discussed on 26 Feb 2024 - Change provider property to be of type Agent - this should resolve this.  We also removed the `providerUrl` property since it is redundant with the `extenalIdentifier`

goneall:
Reviewed during the service profile calls - I'm going to go ahead and merge this before working on the second update

PR #634, opened 2024-02-13
Adds IRIs to individual licenses

zvr:
This adds specific IRIs to the individuals in the licensing domain.

This is following up on https://github.com/spdx/spdx-3-model/pull/588#issuecomment-1917924971.

Please note that every class, property, individual, etc. already has an IRI (automatically generated, present in the generated files). The addition of explicit IRIs for Individuals allows us to choose custom ones, which are simpler.


maxhbr (approved):
LGTM

maxhbr:
Spec-parser needs to catch up

goneall:
@zvr - Please take a look at the run failure - looks like we need an update to the spec parser for this issue.

PR #635, opened 2024-02-14
moves PVC to core class

jeff-schutt:
This change moves the PackageVerificationCode.md to the Core Class.

PackageVerificationCode is a Subclassof: model/Core/IntegrityMethod, which, (as with all Non-Element Classes) is a class defined in model/Core/Classes rather than model/Software/Classes


goneall (approved):
LGTM - Thanks @jeff-schutt

kestewart (approved):
Thanks Jeff.   This matches the discussions.

PR #636, opened 2024-02-15
Fix  syntax errors...

zvr:
The newly added vocabulary was written in a different syntax, which is not automatically parsed.


goneall (approved):
LGTM

goneall (approved):

kestewart (approved):
Thanks for catching these Alexios!   LGTM

PR #645, opened 2024-02-17
In EpssVulnAssessmentRelationship.md, remove 'from', 'to', and 'suppliedBy' in syntax

VenkatTechnologist:
I propose that we remove the following three lines from the syntax:

"from": "urn:spdx.dev:vuln-cve-2020-28498",
  "to": ["urn:product-acme-application-1.3"],
  "suppliedBy": ["urn:spdx.dev:agent-jane-doe"],

The 'from' and 'to' do not seem to indicate meaningful semantics here. On fields about vulnerability, they could indicate the time period between which the assessment is valid, but not sure what is being indicated in EPSS context.

'suppliedBy' would always be the EPSS group of FIRST.org, so I don't see a need for explicitly calling it out.


goneall:
From and to are required - closing issue

PR #648, opened 2024-02-20
Update VexVulnAssessmentRelationship.md

VenkatTechnologist:
To be precise, VEX as such does not have a version of its own at least as of now. The version (1.0.0 released in April 2023) is for minimum requirements of data elements. Hence suggest this field to reflect precisely what the version field means.

### DIFF
--- a/model/Security/Classes/VexVulnAssessmentRelationship.md
+++ b/model/Security/Classes/VexVulnAssessmentRelationship.md
@@ -38,7 +38,7 @@ the document's date must be considered as authoritative.
 
 ## Properties
 
-- vexVersion
+- vexMinRequirementsVersion
   - type: xsd:string
   - minCount: 0
   - maxCount: 1


VenkatTechnologist:
There is a key error, and it looks like there are places in code to change if this change is accepted.

zvr (changes requested):
If there is need for the property to be renamed, the changes must also be made in the [property definition](https://github.com/spdx/spdx-3-model/blob/main/model/Security/Properties/vexVersion.md) as well.

VenkatTechnologist:
> If there is need for the property to be renamed, the changes must also be made in the [property definition](https://github.com/spdx/spdx-3-model/blob/main/model/Security/Properties/vexVersion.md) as well.

Updated properties too, and created a separate pull request for that. Note that default version should be "1.0.0", not zero. Mentioned the same in the property description.

goneall:
@VenkatTechnologist - if PR #649 is still valid - please add this change to PR #649 and close this one

jeff-schutt (changes requested):
IMO we should not change the name of this property. Reason explained [here](https://github.com/spdx/spdx-3-model/pull/649#pullrequestreview-1987271049).

puerco:
+1 for not changing this property is not the version of vex.

rnjudge:
This PR should be merged with https://github.com/spdx/spdx-3-model/pull/649. Closing this one so we can continue discussion on 649.

PR #649, opened 2024-02-20
Update vexVersion.md

VenkatTechnologist:
'vexVersion' meant here stands for minimum requirements for data elements version released by CISA. Hence it is suggested that the field be updated to 'vexMinRequirementsVersion'. The other elements in the field like description and metadata are also being updated.

### DIFF
--- a/model/Security/Properties/vexVersion.md
+++ b/model/Security/Properties/vexVersion.md
@@ -1,17 +1,19 @@
 SPDX-License-Identifier: Community-Spec-1.0
 
-# vexVersion
+# vexMinRequirementsVersion
 
 ## Summary
 
-Specifies the version of the VEX document.
+Specifies the minimum requirements of data elements version that the VEX document adheres to.
 
 ## Description
 
-The document version default value is zero. When any VEX-related content changes, the version must be incremented.
+The version default value is "1.0.0". When minimum requirements of data elements
+version of the VEX document is updated, this field of that VEX document should be
+updated to the same value.
 
 ## Metadata
 
-- name: vexVersion
+- name: vexMinRequirementsVersion
 - Nature: DataProperty
 - Range: xsd:string


goneall:
I noticed the CI is failing - I think this is due to some references to the old name `vexVersion` which have not been updated to the proposed new name `vexMinRequirementsVersion`.

VenkatTechnologist:
Hi Gary,

Would you like me to withdraw this change?
We can track this change as part
of the broader scope of VEX related observations
and incorporate it if needed at that time.

Thanks,
Venkat.

goneall:
> Would you like me to withdraw this change?

Let's just switch it to "Draft" mode.  We can switch it back after the VEX discussions.

goneall:
@VenkatTechnologist - is this PR still valid based on your analysis?

jeff-schutt (changes requested):
> 'vexVersion' meant here stands for minimum requirements for data elements version released by CISA. 

From what I recall this was not the intended use of this field. 

This PR appears to change the definition from supporting one of the data fields that is required `2.2.2 Document version [doc_version]` to support one that is optional `2.0 VEX data elements`, both described [here](https://www.cisa.gov/sites/default/files/2023-04/minimum-requirements-for-vex-508c.pdf). Note the differences between use of MUST and SHOULD.

model/Security/Properties/vexVersion.md
```diff
@@ -1,17 +1,19 @@
 SPDX-License-Identifier: Community-Spec-1.0
 
-# vexVersion
+# vexMinRequirementsVersion
 
 ## Summary
 
-Specifies the version of the VEX document.
+Specifies the minimum requirements of data elements version that the VEX document adheres to.
```
puerco, line 7: This change is not correct, but note that `vexVersion` defines the VEX statement version (not the document). It is used in the vex relationships.  This is the version detailed in 2.4.2 of the vex Minimum Requirements linked by Jeff above. We named it `vexVersion` (as opposed to just `version`) because there was a name clash.

Essentially, the Minimum Requirements doc states the fields all vex implementations need to define. It is a spec of specs but the source of truth is the SPDX spec itself, not the CISA document. In other words, this is a field defined by SPDX not by the Minimum Reqs.

puerco (changes requested):

model/Security/Properties/vexVersion.md
puerco: To clarify, we were relying on another field to serve as the document version and the summary of `vexVersion` should read "Specifies the version of a VEX statement."

rnjudge:
@VenkatTechnologist @jeff-schutt @puerco What I'm gleaning from reading through the comments on this PR and https://github.com/spdx/spdx-3-model/pull/648 is that we should **not** merge this PR as is and, instead, change the Summary definition of `vexVersion` in `model/Security/Properties/vexVersion.md` to be: "Specifies the version of a VEX statement." This would mean also closing https://github.com/spdx/spdx-3-model/pull/648. 

If everyone is agreeable, I can open a PR with the change proposed above.

puerco:
@rnjudge sounds good to me :+1: 

I can also fix, no problem. But I thought Venkat can update this PR and we're done :)

VenkatTechnologist:
@puerco, which SPDX field is used for doc version?

VenkatTechnologist:
If vexVersion is meant to be the statement version, recommend it to be
clearly named as vexStatementVersion so that there is no confusion.

puerco:
> which SPDX field is used for doc version?

That's the catch with SPDX docs: They are supposed to be immutable, so there are no versions.

> recommend it to be clearly named as vexStatementVersion 

So, we don't have a vex statement in SPDX. A VEX statement is assembled by a triad of (at least):

   a software package + a vex assessment relationship + a vulnerability

The `vexVersion` field marks is a property of the assessment. This means that naming it `vexStatementVersion` is not correct because there is no vex statement element. It would have been named `VexVulnAssessmentRelationship.version` but there was a naming collision and we had to look for an alternative.

VenkatTechnologist:
which SPDX field is used for doc version?

***@***.***: "That's the catch with SPDX docs: They are supposed to be
immutable, so there are no versions."*

Then how do we implement doc versions of VEX implementation? Each VEX
document can be updated
(and its version incremented) when there is an update to any of the
statements within. How would this be
implemented in SPDX?

***@***.***: "So, we don't have a vex statement in SPDX. A VEX statement is
assembled by a triad of (at least):*

*a software package + a vex assessment relationship + a vulnerability*


*The vexVersion field marks is a property of the assessment. This means
that naming it vexStatementVersion is not correct because there is no vex
statement element. It would have been
named VexVulnAssessmentRelationship.version but there was a naming
collision and we had to look for an alternative."*
As I mentioned above, a VEX statement does not stand on its own. It is
embedded in a VEX document
because of the way the VEX structure is designed. Bypassing this structure
does not seem to be a
good idea in terms of consumption perspective. Consumers will be clearly
confused.

To me, the vex assessment relationship is what the 'doc' level is supposed
to be. It makes sense to have
the 'vexVersion' as 'vexDocVersion', and then have the 'vulnerability'
point to the statement level details.

If what I said makes sense, we can look at this change in SPDX 3.1 and
leave things as is for SPDX 3.0.
This will also enable us to collect feedback from consumers on the VEX
implementation in SPDX 3.0.
(I personally feel there will be a LOT of explanation to do.)

rnjudge:
@VenkatTechnologist @puerco can you both plan to join the security call this Wed to discuss and come to a conclusion on this?

VenkatTechnologist:
I am in the midst of a travel. My views are already recorded here.

My recommendation is to defer this to 3.1 and look at it in detail when we
have time.
Thanks.

rnjudge:
Decision in the April 10th security call to defer this to 3.1 per Venkat's comment. I will open a PR to update the Summary definition of vexVersion in model/Security/Properties/vexVersion.md to be: "Specifies the version of a VEX statement." per puerco's recommendation and to clear up confusion in the short term.

goneall:
This looks like it was fixed with #709

PR #661, opened 2024-03-08
model: Core: DateTime: Inherit from xsd:dateTime

JPEWdev:
Inherit from xsd:dateTime instead of xsd:string. The XSD definition is compatible with the SPDX definition, and this will allow code generation tools to map the value to a code-native datatype.

### DIFF
--- a/model/Core/Datatypes/DateTime.md
+++ b/model/Core/Datatypes/DateTime.md
@@ -15,7 +15,7 @@ The specific format is one of the most commonly used ISO-8601 formats.
 ## Metadata
 
 - name: DateTime
-- SubclassOf: xsd:string
+- SubclassOf: xsd:dateTimeStamp
 
 ## Format
 


zvr:
👎 I disagree.

All our "Datatypes" are `xsd:string` with a `pattern` defining the valid format.

Note that `Core/DateTime` does *not* allow everything that `xsd:dateTime` does.

JPEWdev:
The SPDX DateTime pattern is a strict subset of the allowed xsd:dateTime string, so it's not incompatible, it would just be a further restriction.

The nice part about using xsd:dateTime is that it means language bindings can know to use a native datetime type which is convenient for users; otherwise it just looks like another string.

zvr:
Very briefly:

- I don't even know how `dateTime` values are represented in JSON-LD (asthey are not strings). I know that in other serializations, e.g. TTL, you have to always use something like `"2024-03-14T12:34:56"^^xsd:dateTime` and this is not very convenient.
- I am also not sure whether we can have SHACL `pattern` restrictions on non-string values. The SHACL online playground does not seem to support it, but I don't know whether this is a restriction of the implementation or the SHACL spec. On the other hand, if there are no validators that support it, the point is moot...

goneall:
I did a bit of research on `xsd:dateTime`:

- It is a subtype of `xsd:anySimpleType` which is the base type for both `dateTime` and `string` - so it would be incompatible to switch this - probably a decision we need to make in 3.0
- You are allowed to further restrict `xsd:dateTime` with a pattern
- Based on several stackoverflow articles, I concluded that it is not common to restrict `xsd:dateTime` with a pattern, so we may get somewhat mixed results from some XML and RDF parsers

Based on the above I'm leaning towards changing the type inheritance to `xsd:dateTime` as proposed.

References:
- http://www.datypic.com/sc/xsd/t-xsd_anySimpleType.html
- https://www.w3.org/TR/xmlschema11-2/#dateTime

kestewart:
Mostly agreed to in 3/19 meeting,  if @zvr  doesn't comment further by 3/28 - will be merged.

kestewart:
Alexios joined, and agreed that decision made last week can stand.   Merging.

PR #662, opened 2024-03-08
Core: Change SpdxId to IdProperty

JPEWdev:
This field is intended to be implemented by the underlying ID property in the serialized format (e.g. "@id" in JSON-LD), so change it to the special "IdProperty" nature so that the model parsing code can recognize this and make the mapping.

### DIFF
--- a/model/Core/Properties/spdxId.md
+++ b/model/Core/Properties/spdxId.md
@@ -16,6 +16,6 @@ so that relationships between Elements can be clearly articulated.
 ## Metadata
 
 - name: spdxId
-- Nature: DataProperty
+- Nature: IdProperty
 - Range: xsd:anyURI
 


JPEWdev:
See https://github.com/spdx/spec-parser/pull/105

zvr:
I thought we had agreed that we will handle the `spdxId` property in a special way, not that we'll create a new value for property nature.

And that we would add in its description that it gets mapped into `@id` in JSON-LD.


I don't mind implementing it this way, but I fear we'll lose the explanations. `Object Property` and `Data Property` are well-understood RDF terms, nothing introduced by us. If we allow `Nature` to take other arbitrary values, we would have to explain them somewhere -- and we don't have any explanation of the format in the spec.

JPEWdev:
I wasn't aware that the terms in `Nature` were specific to RDF; I assumed they were only things that the spec parser cared about. Anyway, having them be RDF terms I think makes we want to do it this way more though. The ID is certainly not a "Data Property" in the RDF sense; in some encodings like JSON it kinda looks like one because of the @id field, but in RDF you can't have an "object" without an ID because it has to be the subject of the triplet (AFAIK, but I did just finally wrap my head around RDF a few days ago).

Maybe a better name would make that more clear; perhaps `Subject` or `SubjectProperty`?

zvr:
Actually... this is nothing new here.

Let's have a look at how SPDXv2 handled the exact same situation and see whether we agree it was a correct approach that we can repeat (can't really remember right now).

goneall:
> Let's have a look at how SPDXv2 handled the exact same situation and see whether we agree it was a correct approach that we can repeat (can't really remember right now).

In RDF representations, the ID is the RDF ID (e.g. in RDF/XML represented by the `about` attribute - see https://github.com/spdx/spdx-spec/blob/8a595028b0386138c84f7188237787c374f6a6cf/examples/SPDXRdfExample-v2.3.spdx.rdf.xml#L7 for an example).

In all non-RDF formats, we translate the SPDX ID into the full ID by concatenating the document namespace to the SPDX ID field.  This is only supported for the classes that have an explicit SPDX ID tag in the tag/value spec.  For classes that do not have an explicit SPDX ID, the RDF ID information is lost - when I translate non-RDF formats to RDF, I just create blank nodes.

JPEWdev:
I think it's also important to keep in mind that this model is used to generate the documentation, so having the `spdxId` field documented would actually be useful. While I appreciate that it's not a "pure" use of the model to document a field that is implementation specific, the documentation _is_ the thing that end users will want to read to figure out how to write SPDX documents; if it's too disconnected from the reality of serializing something, or omits key things (like the fact that a `spdxId` is required and must be a valid IRI!). The vast majority of users (especially if we are actually expecting SPDX to become popular) using SPDX won't care about the purity of the model; they will care if they can _very specifically_ read the documentation and write a correct JSON-LD serialized document from said documentation. IMHO, if we are too strict about the purity of the "model", we will be in danger of making it's usage incomprehensible (or at least unnecessarily difficult) for users and SPDX will have a hard time gaining traction.

If it were up to me, I'd put a JSON-LD fragment for each object _in_ the documentation to describe how to write it; _that's_ user friendly :)

JPEWdev:
Abandon in favor of #673

PR #666, opened 2024-03-12
Change isDirectory to fileKind enumeration

### LINKED ISSUE #630: Consider extensibility for package "file" types
In issue #83 and PR #618, there appears to be an implicit assumption that packages are either files or directories.  However there are potentially many other storage formats for packages, including things like S3 buckets, artifact management system identifiers/URIs, even git coordinates (git is occasionally used for BinOps / binary management).

The preferred solution for RC2 described in PR #618 (add a `isDirectory` flag to the `File` class) makes support for other forms of storage by the SPDX specification in future more difficult and may require backwards-compatibility breaking changes, and there may be better alternatives, including but not limited to:
* instead of an `isDirectory` flag, use a `type` enumeration that might initially have a controlled vocabulary of `"file" | "directory"`, but might be extended in future with additional values (`"s3bucket"`, for example)
* use a class hierarchy for package files that might initially only have two subclasses: `File` and `Directory`
	* note: there may also be value in having virtual / "marker" classes in the hierarchy with names such as `Singleton` and `Collection`, and then have `File` and `Directory` subclass the relevant marker class (and in future, a hypothetical `S3Bucket` class would also subclass `Collection`, for example)

**Meta context**
In the `spdx-tech` meeting on 2024-02-06, the decision was made to continue with PR #618 as written in the interest of progressing RC2, and to revisit this bigger issue as a separate issue in a post-RC2 discussion.  This issue is intended to be a placeholder for that discussion.

goneall:
Fixes #630

Note that fileKind was choosen since fileType is already in use


model/Software/Classes/File.md
```diff
@@ -24,7 +24,7 @@ If the isDirectory property is specified and set to true, then the file represen
   - type: /Core/MediaType
   - minCount: 0
   - maxCount: 1
-- isDirectory
+- fileKind
   - type: xsd:boolean
```
zvr, line 28: Surely not `boolean` ?!?

zvr (changes requested):

model/Software/Classes/File.md
goneall: Good catch @zvr - fixed

model/Software/Vocabularies/FileKindType.md
```diff
@@ -0,0 +1,20 @@
...
+
+An SPDX file may represent a file on disk or a directory of files.  In the future, this may be extended to other kinds (e.g. network based files).
+
+## Metadata
+
+- name: FileKindType
+
+## Entries
+
+- file: The file represents a single file.
```
kestewart, line 19: Can we make it explicit, that if the field is not specified, the assumption is that fileKind is "file".

model/Software/Properties/fileKind.md
```diff
@@ -0,0 +1,18 @@
...
+
+# fileKind
+
+## Summary
+
+Describes if a given file is a directory or non-directory kind of file.
+
+## Description
+
+An SPDX file may represent a file on disk or a directory of files.  In the future, this may be extended to other kinds (e.g. network based files).
```
kestewart, line 11: file could be in memory not just "disk".... probably want to rephrase to something like: 
"represent a file on disk" to "represent a specific file"

kestewart (changes requested):
Couple of ambiguities should be cleared up.  See comments.

goneall:
@zvr @kestewart - updated per comments - please review or request changes

kestewart (approved):
Thanks Gary.  Both issues are resolved. 
Good to merge from my perspective.

PR #670, opened 2024-03-21
DO NOT MERGE: ID example for Sean

JPEWdev:
@sbarnum These are the changes I made to the model for the SPDX 3 playground example. The datetime change is already captures in #661 .


kestewart:
Example is present.  Ok'd by Josh to close.

PR #673, opened 2024-03-22
Remove SpdxId

JPEWdev:
Removes the SpdxId property as it's not a real property in the model and confuses the spec parser

Per discussion in the serialization working group with @sbarnum


kestewart:
This doesn't make sense to me, in that how are we able to refer to elements between serialized documents without it.   Let's discuss in Tech call.

kestewart:
This was discussed in the call, and strong concerns about this, and what it will break for referring to between serializations.    We need to have Gary, Alexios, Maximillian, Sean, Bob, Kate, Josh in the serialization meeting - to reconcile this.  Ideally like to have the tool handle rather than revisit.

kestewart:
Lack of critical mass in the serialization meeting to discuss this issue,  as planned tuesday.  As a result, we're going to close this, as it's going to be way too confusing for folks using the field to refer between documents and elements without it, in other serializations (other than rdf).    Discussed this with Joshua and he is fine with closing it.

PR #680, opened 2024-03-29
Glossary.md: Sort terms alphabetically (A-Z)

bact:
- Sort terms A-Z
- Replace abbreviations like "json-ld", "xml" with "JSON-LD", "XML"
- Break few long paragraphs into smaller ones
- No change in content


kestewart (approved):
Thanks for pulling together this cleanup.    Much appreciated.

goneall (approved):

Issue #676, opened 2024-03-26
[3.1] AI: Add inferenceEnergyConsumption and totalEnergyConsumption


bact:
Per discussion in https://github.com/spdx/spdx-3-model/pull/671#discussion_r1534622099 with @bennetkl, also related to #677 and #682, the next version of `energyConsumption` may also include the energy consumption at the inference time.

It is also possible to include the estimation of the energy consumption for the entire life cycle of an AI system.

This issue will be use to collect
- Definitions
- Estimation methods
- Links to relevant technical/legal requirements or documents

## Proposed properties and descriptions

A free-form text that captures the amount of energy needed to train and operate the AI model.

- `trainingEnergyConsumption` for energy consumption during the training of an AI model
  - Training energy consumption is the amount of energy consumed during the training phase of an AI model. This includes the energy used to power the hardware and software systems used for training, such as GPUs, CPUs, and data centers.
  - Renamed from `energyConsumption` #677
  - [EU AI Act](https://www.europarl.europa.eu/meetdocs/2014_2019/plmrep/COMMITTEES/CJ40/AG/2024/02-13/1296003EN.pdf) refs: see Points (2)(d) and (2)(e) in Annex IXa Section 1 and Point (c) in Annex IXc

- `inferenceEnergyConsumption` for energy consumption per one inference
  - Inference energy consumption is the amount of energy consumed per one inference of an AI model. This includes the energy used to power the hardware and software systems used for making predictions or decisions based on the model, such as edge devices, servers, or cloud services.

- `totalEnergyConsumption` for energy consumption during the entire life cycle of the system
  - Total energy consumption is the total amount of energy consumed by an AI model over its entire lifecycle, including training, inference, and any other related activities.

bact:
Another way to record the energy consumption is to record the compute (FLOPs and FLOPs per second)
https://epochai.org/blog/estimating-training-compute 
https://blog.montrealanalytics.com/predicting-and-reducing-energy-consumption-of-machine-learning-models-7dc44e197225

bact:
AI team meeting 2024-04-03 agrees to change `energyConsumption` property range from `xsd:string` to a `EnergyConsumption` class. 

The `EnergyConsumption` class will have these properties: `trainingEnergyConsumption`, `finetuningEnergyConsumption`, `inferenceEnergyConsumption` - all with `EnergyConsumptionDescription` type.

[...]

PR #682, opened 2024-03-29
[WIP] AI: Rename energyConsumption to trainingEnergyConsumption (+suggested unit?)

### LINKED ISSUE #677: [3.0] AI: Rename: energyConsumption to trainingEnergyConsumption
From its description, the `energyConsumption` is intended for being a training energy consumption.

This is confirmed by a conversation here https://github.com/spdx/spdx-3-model/pull/671#discussion_r1535291356

The proposal is to rename `energyConsumption` to `trainingEnergyConsumption` to make it more explicit.

This is also to avoid conflict and make rooms for possible future properties like `inferenceEnergyConsumption` and `totalEnergyConsumption` (see #676).

bact:
- To fix #677 
  - See also Points (2)(d) and (2)(e) in Annex IXa Section 1 and Point (c) in Annex IXc in the [provisional agreement of the EU AI Act](https://www.europarl.europa.eu/meetdocs/2014_2019/plmrep/COMMITTEES/CJ40/AG/2024/02-13/1296003EN.pdf).
- Do we need a suggested unit of energy in the description?
  - For example,
    - Kilowatt-hour (kW-h) -- a non-SI unit, a common billing unit for electrical energy
    - Megajoule (MJ) -- a SI unit, 1 J is 1 Watt-second, 1 KW-h is 3.6 MJ.

### DIFF
--- a/model.drawio
+++ b/model.drawio
@@ -1719,7 +1719,7 @@
         <mxCell id="96wljPxh_fQPzhBJDJxU-11" value="+ Core/Artifact/releaseTime: DateTime[1]" style="text;strokeColor=none;fillColor=#E1D5E7;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;" parent="96wljPxh_fQPzhBJDJxU-1" vertex="1">
           <mxGeometry y="146" width="413.51" height="30" as="geometry" />
         </mxCell>
-        <mxCell id="96wljPxh_fQPzhBJDJxU-7" value="+ energyConsumption: String[0..1]" style="text;strokeColor=none;fillColor=none;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;" parent="96wljPxh_fQPzhBJDJxU-1" vertex="1">
+        <mxCell id="96wljPxh_fQPzhBJDJxU-7" value="+ trainingEnergyConsumption: String[0..1]" style="text;strokeColor=none;fillColor=none;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;" parent="96wljPxh_fQPzhBJDJxU-1" vertex="1">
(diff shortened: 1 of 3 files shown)


bact:
@bennetkl @rgopikrishnan91 please kindly review. Thank you

bennetkl:
Yeah, I saw the suggestion to change Kilowatt-hour (kW-h) to Watts per minutes.  I wasn't at the meeting where this was discussed, but consideration that training time / energy is hug, I would think that minutes might be too small a unit, but will let Gopi add his comments, since I don't have a feel for why we wanted to change it.

rgopikrishnan91:
I think KwH might be a better unit too. But is there a reason we can't allow it to be expressed in both ways and ask them to mention the unit? That way the ppl are not restricted and the parser is still happy? or much like date format we should pick one and stick to it. 

I understand the reason for using an SI unit, but I worry if we will be forcing ppl to do some conversions which might affect the uptake and hence my hesitation to pick a unit. 
@kestewart @bact @bennetkl

bact:
Currently the filed is a string. So the maintainer can express it in anyway they like and just mention the unit.

In the description, we can ask them to mention the unit - and, if you have a preference/preferences, we can suggest that in the property description as well. Or we can just mention some common units with no preference as well.

And, yes, the Kw-H is much more common.

bact:
There's also another way/unit to record the energy as well, to record the compute. See https://github.com/spdx/spdx-3-model/issues/676#issuecomment-2030949971

rgopikrishnan91:
So two thoughts. 1. Inference energy consumption and 2. training energy consumption should be seperately noted?  I vote for giving a few options and specifying the format in which they should be specified. For example, allowed options could be (kwh, joules) and it should be mentioned as x.xx unit. (just a representative example, not a list of all the units we should consider

kestewart:
Discussing this on the call, and agreement to make this a bit more future proof so evolving to have a Class - EnergyConsumption

{
  "energyConsumption": [
    {
      "type": "inference",
      "units": "kW/J",
      "quantity": 10
    },
    {
      "type": "learning",
      "units": "kW/J",
      "quantity": 5
    }
  ]
}

bact:
I will close this PR. Please see instead these @rgopikrishnan91 PRs:
- #694
- #695
- #696
- #697

Summary:

AI team meeting 2024-04-03 agrees to change `energyConsumption` property range from `xsd:string` to a `EnergyConsumption` class. 

The `EnergyConsumption` class will have these properties: `trainingEnergyConsumption`, `finetuningEnergyConsumption`, `inferenceEnergyConsumption` - all with `EnergyConsumptionDescription` type.

`EnergyConsumptionDescription` will contains a unit information. The unit will have a type of`EnergyConsumptionDescriptionUnitType`.


More refs about compute and energy estimation are avail in #676

PR #686, opened 2024-04-02
Doc: Add SHACL examples

zvr:
This adds examples of generated SHACL to represent the information of the model.

Discussed on the 2023-03-19 Tech call.

### DIFF
--- a/serialization/shacl.md
+++ b/serialization/shacl.md
@@ -1,0 +1,75 @@
+# RDF/SHACL model
+
+Quick notes on how the model description is translated into SHACL
+
+# Class example
+
+```ttl
+<https://rdf.spdx.org/v3/Software/Snippet> a rdfs:Class,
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM - I realize there are some related enhancements that may change the example, but I think we should go ahead and merge after one more review

sbarnum (changes requested):
I think there are a few tweaks and simplifications needed before this PR is accepted and merged.

goneall:
Let's move this to 3.1 to make sure the SHACL generated matches the documentation.  Any of the above suggested changes will not be a breaking change to the SHACL document itself.

zvr:
So, this PR was about adding quick notes from the meeting we had about SHACL.

It seems we want better documentation, so I'll just close and abandon this.

I've opened #788 to document the need.

PR #688, opened 2024-04-02
Clarify meaning of empty copyrightText string

### LINKED ISSUE #655: How should a blank / empty `copyrightText` be interpreted?
In the context of [ORT parsing SPDX files](https://github.com/oss-review-toolkit/ort/issues/7222) the question arose how a given `"copyrightText": ""` for `files` should be interpreted. None of the specs seem to be crystal clear on this. IMO all of these options have some merit:

- interpret a blank / empty `copyrightText` as `NONE`
- interpret a blank / empty `copyrightText` as `NOASSERTION`
- interpret a blank / empty `copyrightText` as a liternal copyright declaration that is empty (even if that does not really add any value)

Currently, ORT simply refuses to parse SPDX files with such `copyrightText` to avoid ambiguity.

Any advice how to handle this case?

PS: The same question basically arises for a set (i.e. present) `licenseConcluded` that is set to an empty / blank string.

swinslow:
This PR just clarifies that an "empty" copyrightText string should be interpreted as equivalent to `NOASSERTION`.

Fixes #655 

### DIFF
--- a/model/Software/Properties/copyrightText.md
+++ b/model/Software/Properties/copyrightText.md
@@ -27,6 +27,11 @@ following applies:
 * the SPDX data creator has intentionally provided no information (no
   meaning should be implied by doing so).
 
+If a copyrightText is present, but consists of solely an empty string or a
+string with no substantive content (e.g., a string that contains only
+whitespace), then this should be interpreted as equivalent to a "NOASSERTION"
+value as described above.
+
 ## Metadata
 
 - name: copyrightText


swinslow:
@sschuberth @zvr @goneall I think this addresses the intended outcome of the consensus in #655, but feel free to revise or reject if I've gotten it wrong.

model/Software/Properties/copyrightText.md
```diff
@@ -27,6 +27,11 @@ following applies:
 * the SPDX data creator has intentionally provided no information (no
   meaning should be implied by doing so).
 
+If a copyrightText is present, but consists of solely an empty string or a
+string with no substantive content (e.g., a string that contains only
```
sschuberth, line 31: IMO "no substantive content" is a bit too fluffy (as this commit intends to clarify things). For example, some people could interpret the literal copyright holder "E.G." as non-substantive as they might interpret it as "exempli gratia", i.e. a placeholder, whereas other might interpret it as initials of a name.

Thus, I propose to clearly say

> If a copyrightText is present but empty or solely consists of whitespace, it should be interpreted as equivalent to a "NOASSERTION" value as described above.

sschuberth (changes requested):

zvr (approved):

kestewart (approved):
Reviewed in tech call, and change generally looks good to participants.

model/Software/Properties/copyrightText.md
zvr: I disagree that it should only cover empty or blank strings; "no substantive content" is correct.

sschuberth: > "no substantive content" is correct.

I'm sorry, but this is exactly the problem we have with SPDX when it comes to real-world implementations: Legal language being used as a basis for technical specification. I'd have no idea how a validation tool should check for "no substantive content" concretely.

PR #696, opened 2024-04-04
Create EnergyConsumptionDescription

rgopikrishnan91:
@kestewart @goneall Can you please review this PR. Part 3 of the PR

### DIFF
--- a/model/AI/Classes/EnergyConsumptionDescription
+++ b/model/AI/Classes/EnergyConsumptionDescription
@@ -1,0 +1,27 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# EnergyConsumptionDescription
+
+## Summary
+
+The Class that helps note down the actual energy consumption value and the unit in which it is recorded.
+
+## Description
+
+The Class that helps note down the actual energy consumption value and the unit in which it is recorded. Value property describes the energy consumption and the unit field describes the units in which it was recorded. 
+
+## Metadata
+
+- name: EnergyConsumptionDescription
+- Instantiability: Concrete
+
+## Properties
+
+- Value
+  - type: xsd:string
+  - minCount: 1
+- unit
+  - type: EnergyConsumptionDescriptionUnitType
+  - minCount: 1
+
+


model/AI/Classes/EnergyConsumptionDescription
```diff
@@ -0,0 +1,27 @@
...
+## Metadata
+
+- name: EnergyConsumptionDescription
+- Instantiability: Concrete
+
+## Properties
+
+- Value
+  - type: xsd:string
+  - minCount: 1
```
bact, line 22: may need `maxCount: 1` as well?

model/AI/Classes/EnergyConsumptionDescription
```diff
@@ -0,0 +1,27 @@
...
+- unit
+  - type: EnergyConsumptionDescriptionUnitType
+  - minCount: 1
```
bact, line 25: may need `maxCount: 1` as well?

bact:
Actual vs Estimated value

The proposed summary and description mentioned "actual energy consumption".

But a lot of models didn't recorded the actual energy consumption. Instead, they use an estimation.

For example, from Meta LLaMA paper https://arxiv.org/abs/2302.13971 :
"We follow a formula for Wu et al. (2022) to estimate the Watt-hour, Wh, needed to train a model,"

--

I propose to change the summary to:
- "The Class that helps note down the energy consumption value and its unit."

and the description to:
- "The Class that helps note down the energy consumption value and its unit. The energy consumption can be an actual recorded value or an estimation."

--

Additionally, if we like, we can also introduce another Class property `actualOrEstimated` (can't thinking of a better name) to note if the value is an actual value or an estimated value. (This actual vs estimated is also common in an electricity bill).

We can also decided to have `actualOrEstimated` as an optional property. Missing it will imply that the value is actual.

Something along this line:
```json
{
  "energyConsumption": [
    {
      "type": "inference",
      "unit": "kWh",
      "value": 10,
      "actualOrEstimated": "estimated"
    },
    {
      "type": "learning",
      "unit": "kWh",
      "value": 5
    }
  ]
}
```

--

(Also btw sorry for not being in the yesterday meeting from the beginning to join the discussion.
I'm now in UTC+7 and it's a bit difficult during these weeks to attend for me.)

model/AI/Classes/EnergyConsumptionDescription
```diff
@@ -0,0 +1,27 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# EnergyConsumptionDescription
+
+## Summary
+
+The Class that helps note down the actual energy consumption value and the unit in which it is recorded.
```
kestewart, line 7: Remove word "actual" - see comments from @bact below.  I think that in some cases "actual" may be hard to establish - so best estimate may be all that is possible.

model/AI/Classes/EnergyConsumptionDescription
```diff
@@ -0,0 +1,27 @@
...
+
+## Description
+
+The Class that helps note down the actual energy consumption value and the unit in which it is recorded. Value property describes the energy consumption and the unit field describes the units in which it was recorded.
```
kestewart, line 11:
```suggestion
The Class that helps note down the energy consumption value and the unit in which it is recorded. Value property describes the energy consumption and the unit field describes the units in which it was recorded. 
```

model/AI/Classes/EnergyConsumptionDescription
kestewart, line 7:
```suggestion
The Class that helps note down the energy consumption value and the unit in which it is recorded.
```

kestewart (changes requested):
Please merge changes with others, and add in "Signed off by"

rgopikrishnan91:
Closed since its now merged as a part of #697

Issue #646, opened 2024-02-17
Update PackageVerificationCode to have algorithm explicit


kestewart:
In https://github.com/spdx/spdx-3-model/pull/643 the PackageVerificationCode model had the algorithm, but the text for the class had it hardcoded as SHA1.   The model was updated to match, but the algorithm should be put back into the diagram & the description of packageVerificationCode rewritten to include use of algorithms other than SHA1.

kestewart:
Need PVR to be added to HashAlgorithm (in model and diagram).

kestewart:
<img width="623" alt="Screenshot 2024-03-28 at 11 04 03 AM" src="https://github.com/spdx/spdx-3-model/assets/13152682/888e6ed0-a775-4ff5-827b-f5c60b950587">

Need to get example to see how this is able to express with different HASH algorithms - @sbarnum can you help here

kestewart:
DIscussion from the call,  preference is to replicate algoirthm in PackageVerificaitonCode.

PR will be submitted.

goneall:
Resolved with PR #699

PR #699, opened 2024-04-06
Update PackageVerificationCode.md

kestewart:
Add in Algorithm, and rework algorithm example to use example fields.

### DIFF
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -18,25 +18,18 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
-verificationcode = 0
-
-filelist = templist = ""
-
-for all files in the package {
-
-    if file is an "excludes" file, skip it /* exclude SPDX analysis file(s) */
-
-        append templist with "SHA1(file)/n"
+    verificationcode = 0
+    filelist = templist = ""
+    for all files in the package {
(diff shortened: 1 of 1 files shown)


kestewart:
This addresses #646

maxhbr:
I thought that the algorithm [was fixed to `SHA1` and is not parametric](https://spdx.github.io/spdx-spec/v2.3/package-information/#79-package-verification-code-field)? Is there a reason to change this compared to v2? I thought that package verification code is also basically deprecated and discourage from being used. Making it more capable feels weird.

goneall:
@maxhbr you raise a good point.  I'm OK either way as this is just for legacy purposes.

It may be easier to migrate if we don't have the property and just define it as SHA1 since we don't have an algorithm property in 2.3.

@kestewart - thoughts?

kestewart:
It had been talked about using SHA256 as this will be more common than SHA1.   At this point, prefer to leave it flexible with "algorithm", and can restrict it if necessary going forward.

maxhbr:
we describe the PackageVerificationCode with

> This verification method is provided for compatibility with SPDX 2.X.
> 
> Use of this verification code method is discouraged except for scenarios where the `gitoid` property on `Artifact` can not be used.

In my opinion, this makes complicated code that seldomly gets implemented correctly more complex.

maxhbr (approved):
The code looks fine, but I think this should not be merged.

Also the V2->V3 migration guide would need to be adopted for this change.

goneall (approved):
LGTM - consistent with discussions on tech call

PR #707, opened 2024-04-09
Add `/Extension/` prefix to two `Extension` in diagram

bact:
- For labelling consistency
- Also add a space between a colon and a type
  - between "extension:" and "/Extension/"
  - between "useSensitivePersonalInformation:" and "/Core/PresenceType[0..1]"


bact:
I'm actually not sure which one to choose between `/Extension/Extension` and `Extension`,
but we probably need to choose one and apply the same convention to all those four instances in the diagrams.

kestewart:
@bact - given Sean's updates - is this still needed?

bact:
I have checked the latest diagram - @kestewart we still need this.

sbarnum:
This is fixed in #718

bact:
Thank you

PR #709, opened 2024-04-10
Small Security profile corrections

rnjudge:
This PR fixes two small Security profile issues that were discovered on the April 10th call.

See https://github.com/spdx/spdx-3-model/pull/649


puerco (approved):
Thanks Rose!

kestewart (approved):
Good catches.   Thanks.

PR #710, opened 2024-04-10
Clarify ExternalIdentifier and ExternalRef

rnjudge:
This commit attempts to clarify the differences between ExternalIdentifier and ExternalRef classes.

Closes #705

PR #711, opened 2024-04-10
Add cwe as ExternalRefType

rnjudge:
Per discussion in the April 10 Security call, this commit adds cwe (common weakness enumeration) as an ExternalIdentifierType.

Closes https://github.com/spdx/spdx-spec/issues/102


puerco (changes requested):
I think this is not correct. 

A CWE string is not an identifier like a CVE number is, for example. A vulnerability can have as its root cause more than one CWE. 

The CWE strings are a reference to a catalog of vulnerability categories. As per your definitions in #710, I think an ExternalRef fits better.

rnjudge:
Thanks @puerco! PR updated.

puerco (approved):
Awesome, thanks!

goneall (approved):
LGTM

PR #703, opened 2024-04-09
Updated model diagrams for changes 02-15-24 to 04-11-24

sbarnum:
This includes all changes in the diff for 02-15-24 to 04-08-24.

The diagram set aligns to the submission to OMG for ISO fast track.
This includes breaking out Core, Core Enumerations, Software, and Software Enumerations diagrams in addition to the preexisting combined Core+Software diagram.
It also includes both .png (for easy and portable web and document inclusion) and .svg (for cleanly scalable use)versions of each document.

PR #712, opened 2024-04-11
Update IRI in diagram for NoAssertionLicense and NoneLicense

bact:
In Licensing model diagram, update IRIs of two Individuals in ExpandedLicensing to new IRIs, as decided on the [2024-04-02 tech call](https://github.com/spdx/meetings/blob/main/tech/2024-04-02.md):
- `https://rdf.spdx.org/v3/ExpandedLicensing/NoAssertionLicense` -> `https://spdx.org/rdf/3.0.0/terms/Licensing/NoAssertion`
- `https://rdf.spdx.org/v3/ExpandedLicensing/NoneLicense` -> `https://spdx.org/rdf/3.0.0/terms/Licensing/None`

--

Also, does there a reason of using the term `IRI` instead of `id` in the Metadata section of these two specific Individuals `NoAssertionLicense` and `NoneLicense`?

If there's no specific reason, I can change that two "IRI:" text labels to "id:", to keep them consistent with others.

Please look at these snippets:

From [model/ExpandedLicensing/Individuals/NoneLicense.md](https://github.com/spdx/spdx-3-model/blob/main/model/ExpandedLicensing/Individuals/NoneLicense.md):

```markdown
## Metadata

- name: NoneLicense
- type: IndividualLicensingInfo
- IRI: https://spdx.org/rdf/3.0.0/terms/Licensing/None
```

From [model/ExpandedLicensing/Individuals/NoAssertionLicense.md](https://github.com/spdx/spdx-3-model/blob/main/model/ExpandedLicensing/Individuals/NoAssertionLicense.md):

```markdown
## Metadata

- name: NoAssertionLicense
- type: IndividualLicensingInfo
- IRI: https://spdx.org/rdf/3.0.0/terms/Licensing/NoAssertion
```

As for every other places, they use `id`:

From [model/Core/Core.md](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Core.md):

```markdown
## Metadata

- id: https://spdx.org/rdf/3.0.0/terms/Core
- name: Core
```

See [more of other instances](https://github.com/search?q=repo%3Aspdx%2Fspdx-3-model+%22%3A+https%3A%2F%2Fspdx.org%2Frdf%2F3.0.0%2F%22&type=code).

### DIFF
--- a/model.drawio
+++ b/model.drawio
@@ -1199,13 +1199,13 @@
         <mxCell id="KA1xjPNPVZ9xSUQX3SqY-5" value="&lt;&lt;Individual&gt;&gt;&#xa;NoAssertionLicense : IndividualLicensingInfo" style="swimlane;fontStyle=1;childLayout=stackLayout;horizontal=1;startSize=26;fillColor=#f5f5f5;horizontalStack=0;resizeParent=1;resizeParentMax=0;resizeLast=0;collapsible=1;marginBottom=0;strokeColor=#666666;fontColor=#333333;" parent="1" vertex="1">
           <mxGeometry x="2160" y="126" width="350" height="56" as="geometry" />
         </mxCell>
-        <mxCell id="_3iI9CYwvOLBjie_8vdV-1" value="https://rdf.spdx.org/v3/ExpandedLicensing/NoAssertionLicense" style="text;strokeColor=none;fillColor=none;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;" parent="KA1xjPNPVZ9xSUQX3SqY-5" vertex="1">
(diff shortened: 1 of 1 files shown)


zvr:
@bact thanks for asking, but there **is** a reason.

Specifically for Individuals, our input format allows specifying an `IRI`, so that we can put one, if we want so.

Only Namespaces have `id`, and everything else (classes, properties, vocabularies, datatypes) gets an `id` automatically based on the namespace they are in.

But for individuals we allow overwriting this, in order to have something more user-friendly, if desired. This is exactly the case with the two individual licenses, where we specify the `IRI` to be `.../Licensing/None` instead of the automatically generated `.../ExpandedLicensing/NoneLicensing`.
If you want the gritty details, the RDF ontology actually generates both and specifies an `owl:sameAs` predicate between them.

Therefore please don't change it.

bact:
Thank you @zvr, that is very clear.

bact:
Close. As it's already got fixed by updates in #703.

PR #713, opened 2024-04-12
AI and Dataset: Sort entries A-Z

bact:
As some lists are getting long and it is inconvenient to find an entry in an unsorted list. 

Sorting rules:
- Do not alphabetically sort the list that is already semantically sorted, such as by level of intensity (as in SafetyRiskAssessmentType, ConfidentialityLevelType, DatasetAvailabilityType)
- Sort A-Z
- Keep `other` and `noAssertion` at the end of the list

### DIFF
--- a/model/AI/Classes/AIPackage.md
+++ b/model/AI/Classes/AIPackage.md
@@ -18,29 +18,36 @@ Metadata information that can be added to a package to describe an AI applicatio
 
 ## Properties
 
+- autonomyType
+  - type: /Core/PresenceType
+  - minCount: 0
+  - maxCount: 1
+- domain
+  - type: xsd:string
+  - minCount: 0
 - energyConsumption
   - type: EnergyConsumption
   - minCount: 0
   - maxCount: 1
-- standardCompliance
-  - type: xsd:string
+- hyperparameter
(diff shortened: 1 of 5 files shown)


bact:
@rgopikrishnan91 @bennetkl would you mind reviewing this please. Thank you.

zvr:
I don't object, as long as it is understood what is being done.

For reference reasons sorted is better for the use case of finding something. It also saves us all the trouble about deciding which should be the "correct" order, to never show any preference between options (e.g. algorithms), etc.
That's why the order of the input files is ignored and generated files are _always_ sorted: in the contents classes, properties, etc. appear in alphabetical order, and inside each one properties, vocabulary entries, etc. (essentially every list/table) are also in alphabetical order.

Therefore, this PR is OK if you want the _model_ files to be in this order for editing, but realize that this will not be reflected in the generated spec.

bact:
Thank you @zvr . Yes, it's for the editing purpose.

I came up to this after I was trying to check the diagrams and found that the unsorted lists in Markdown files are not ready facilitating the job.

kestewart (approved):
Thanks @bact for adding the ordering.

goneall (approved):
LGTM

PR #728, opened 2024-04-23
Add SupportRelationship

JPEWdev:
Adds a new SupportRelationship that uses a derived Relationship to indicate support, since this allows associating the support with a specific entity.

### DIFF
--- a/model/Core/Classes/SupportRelationship.md
+++ b/model/Core/Classes/SupportRelationship.md
@@ -1,0 +1,37 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SupportRelationship
+
+## Summary
+
+Describes how an Agent Provides Support for an Artifact
+
+## Description
+
+Specifies how an Agent supports a given Artifact. The Relationship Type must be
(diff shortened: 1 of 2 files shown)


kestewart (approved):
Thanks Josh.   This matches what was discussed in the call earlier.

model/Core/Classes/SupportRelationship.md
```diff
@@ -0,0 +1,37 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SupportRelationship
+
+## Summary
+
+Describes how an Agent Provides Support for an Element
```
goneall, line 7: Should we have the "to" of the relationship be an `Artifact` rather than an `Element`?  It would seem that only artifacts would be something that would require support.

goneall (changes requested):
One suggestion to consider

goneall:
Thanks @JPEWdev for creating the PR.

goneall:
Ping @JPEWdev - Note review comment above and please rebase.

goneall (approved):
LGTM

PR #739, opened 2024-04-28
Add .editorconfig and .gitattributes to enforce newline and whitespace consistency

bact:
Copying configurations from `spdx/spdx-spec` repo (sorted):

- https://github.com/spdx/spdx-spec/blob/development/v3.0.1/.editorconfig
- https://github.com/spdx/spdx-spec/blob/development/v3.0.1/.gitattributes

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# See also http://editorconfig.org/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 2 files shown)


zvr:
Why shouldn't people in Windows have their native line endings?

Sorry, I may be misunderstanding something...

bact:
We can make it auto (`crlf` for Windows, `lf` for the rest), just try keeping both `spdx-spec` and `spdx-3-model` repos using the same convention.

goneall (approved):
LGTM

kestewart (approved):
This is a very good idea.   Thank you!

PR #741, opened 2024-04-29
AI: autonomyType: Indicate what yes and no mean

### LINKED ISSUE #719: AI: meaning of PresenceType in autonomyType
[`autonomyType`](https://github.com/spdx/spdx-3-model/blob/main/model/AI/Properties/autonomyType.md) in `AIPackage` has a type of [`/Core/PresenceType`](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/PresenceType.md).

`/Core/PresenceType` has 3 possible values:
- yes: Indicates presence of the field.
- no: Indicates absence of the field.
- noAssertion: Makes no assertion about the field. 
 
This is a summary of `autonomyType`:

    States if a human is involved in the decisions of the AI software.

If a human is involved, should `autonomyType` be "yes" or "no"?

In other words, does "Yes" means:
1. "Yes, there is human involved" (refer to the summary: "if a human is involved ...") or
2. "Yes, the system is autonomous" (refer to the name of the field: "autonomyType") ?
 
May need a further clarification in the property description. ("Yes is..., No is ...")

Similar question for:
- [/AI/Properties/useSensitivePersonalInformation](https://github.com/spdx/spdx-3-model/blob/main/model/AI/Properties/useSensitivePersonalInformation.md)
- [/Dataset/Properties/hasSensitivePersonalInformation](https://github.com/spdx/spdx-3-model/blob/main/model/Dataset/Properties/hasSensitivePersonalInformation.md)

bact:
Rewrite the summary and description to centered around automation,
to lead to a direct binary yes/no value (yes = automated, no = not automated).

To resolve #719 (see more discussion in that issue, also suggestions for 3.1)


kestewart (approved):
This helps clarify.  Thanks.

bennetkl (approved):
It's much clearer now, approved for merger

bact:
2024-06-12 AI team meeting approved this change.

--
Note: For next major version (that will allow breaks), we probably need to reconsider if `xsd:boolean` (`true` and `false`), together with a semantic of no assertion, is already suffice for this property. Without the need to use `PrecenseType`. See #719 for full discussion.

PR #742, opened 2024-04-29
Core: Update ProfileIdentifierType entries

### LINKED ISSUE #735: ProfileIdentifierType list
ProfileIdentifierType lists the following profiles:
* ai: the element follows the AI profile specification
* build: the element follows the Build profile specification
* core: the element follows the Core profile specification
* dataset: the element follows the Dataset profile specification
* expandedLicensing: the element follows the expanded Licensing profile specification
* extension: the element follows the Extension profile specification
* security: the element follows the Security profile specification
* simpleLicensing: the element follows the simple Licensing profile specification
* software: the element follows the Software profile specification
* usage: the element follows the Usage profile specification

Questions:
* where is the Usage profile described?
* the Lite profile is missing

bact:
- Remove non-existing `usage` profile (for 3.0, it got pushed to 3.1 see: #191)
- Add missing `lite` profile
- <s>Sort entries (will make it easier to spot anything missing in the future)</s>

This will fix #735 as reported by @vargenau

### DIFF
--- a/model/Core/Vocabularies/ProfileIdentifierType.md
+++ b/model/Core/Vocabularies/ProfileIdentifierType.md
@@ -17,7 +17,7 @@ The profile may also contain additional restrictions on existing properties and
 classes defined in other profiles.
 
 If the creator of an SPDX collection of elements includes a profile in the list
-of conformanceProfiles, they are claiming that all contained elements conform
+of profileConformance, they are claiming that all contained elements conform
 to all restrictions defined for that profile.
 
 ## Metadata
@@ -33,7 +33,6 @@ to all restrictions defined for that profile.
 - expandedLicensing: the element follows the expanded Licensing profile
   specification
 - extension: the element follows the Extension profile specification
-- licensing: the element follows the Licensing profile specification
 - lite: the element follows the Lite profile specification
 - security: the element follows the Security profile specification
 - simpleLicensing: the element follows the simple Licensing profile
--- a/model/Core/Vocabularies/ProfileIdentifierType.md
+++ b/model/Core/Vocabularies/ProfileIdentifierType.md
@@ -26,15 +26,15 @@ to all restrictions defined for that profile.
 
 ## Entries
 
-- ai: the element follows the AI profile specification
-- build: the element follows the Build profile specification
 - core: the element follows the Core profile specification
-- dataset: the element follows the Dataset profile specification
-- expandedLicensing: the element follows the expanded Licensing profile
-  specification
-- extension: the element follows the Extension profile specification
-- lite: the element follows the Lite profile specification
-- security: the element follows the Security profile specification
+- software: the element follows the Software profile specification
 - simpleLicensing: the element follows the simple Licensing profile
   specification
-- software: the element follows the Software profile specification
+- expandedLicensing: the element follows the expanded Licensing profile
+  specification
+- security: the element follows the Security profile specification
+- build: the element follows the Build profile specification
+- ai: the element follows the AI profile specification
+- dataset: the element follows the Dataset profile specification
+- extension: the element follows the Extension profile specification
+- lite: the element follows the Lite profile specification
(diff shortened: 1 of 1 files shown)


model/Core/Vocabularies/ProfileIdentifierType.md
```diff
@@ -4,28 +4,38 @@ SPDX-License-Identifier: Community-Spec-1.0
...
-- build: the element follows the Build profile specification
 - ai: the element follows the AI profile specification
+- build: the element follows the Build profile specification
+- core: the element follows the Core profile specification
 - dataset: the element follows the Dataset profile specification
-- usage: the element follows the Usage profile specification
+- expandedLicensing: the element follows the expanded Licensing profile
+  specification
 - extension: the element follows the Extension profile specification
+- licensing: the element follows the Licensing profile specification
```
zvr, line 36: No, `licensing` is not a profile that should be used in a list of conformant profiles. Only `simpleLicensing` and `expandedLicensing` should be allowed.

model/Core/Vocabularies/ProfileIdentifierType.md
```diff
@@ -4,28 +4,38 @@ SPDX-License-Identifier: Community-Spec-1.0
...
-If the creator of an SPDX collection of elements includes a profile in the list of conformanceProfiles, they are claiming that all contained elements conform to all restrictions defined for that profile.
+
+A profile consists of a namespace that may add properties and classes to the
+core profile unique to the domain covered by the profile.
+
+The profile may also contain additional restrictions on existing properties and
+classes defined in other profiles.
+
+If the creator of an SPDX collection of elements includes a profile in the list
+of conformanceProfiles, they are claiming that all contained elements conform
```
zvr, line 20: The property is called `profileConformance`

zvr (changes requested):

model/Core/Vocabularies/ProfileIdentifierType.md
bact: Thank you

model/Core/Vocabularies/ProfileIdentifierType.md
bact: Thank you, it is now removed.

zvr:
I personally prefer the order that we had, since this is the order they were initially created and we keep referring to (e.g. in presentations).
The point is moot, anyway, since all Vocabularies are presented in alphabetical order in the generated spec.

bact:
The original order makes more logical sense to me as well, for it having some level of semantic grouping (license-related profiles are together, AI and data are together).

My general feeling is that

- if the list is small, semantic grouping/semantic ordering works much better
  - for example, in [SafetyRiskAssessmentType](https://github.com/spdx/spdx-3-model/blob/main/model/AI/Vocabularies/SafetyRiskAssessmentType.md) and [ConfidentialityLevelType](https://github.com/spdx/spdx-3-model/blob/main/model/Dataset/Vocabularies/ConfidentialityLevelType.md) - alphabetical order works really bad
- but once the list gets longer, the positive cognitive effect from semantic grouping will start to diminish
  - due to: too many groups + more different ways of grouping + visual distance on the screen between groups
  - in this case, alphabetical order is probably one of the few than can be applied consistently.

([HashAlgorithm](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/HashAlgorithm.md) is probably the exception here by the nature of its entries which have name prefixes. With that, it can have both semantic grouping and (almost) alphabetical order at the same time.). 

We currently have 10 entries here in ProfileIdentifierType which is not that very long yet (unlike [RelationshipType](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/RelationshipType.md) which is truly long).

I can make the list back to the original order if it is more preferred.

bact:
Tech team call 2024-04-30* agrees to keep the original order (starting with "core") for the time being as the list is not that big. The call also open that in the future when the list is growing we can consider to have it alphabetically.

(*the date in the commit comment was wrong)

kestewart (approved):
Thanks for catching this.

goneall (approved):
LGTM

goneall:
It looks like all of the requested changes have been made - merging

PR #748, opened 2024-04-29
root Element(s) of a tree of elements contained in a BOM (not SBOM)

### LINKED ISSUE #745: rootElement: SBOM or BOM?
In `rootElement`it is written:

This property is used to denote the root Element(s) of a tree of elements contained in an SBOM.

Is it really SBOM, or can it be a BOM?

BOM is defined in Core, SBOM is defined in Software only.

vargenau:
Fixes #745


goneall (approved):
LGTM

PR #750, opened 2024-04-29
Use consistent case.

vargenau:
"Bill of Materials" is always written with "of" in the spec.


zvr (approved):

kestewart (approved):
Yes, we need to be consistent.   Thank you.

PR #751, opened 2024-04-29
Make URL clickable.

vargenau:


zvr:
In general we need to have 

```
 see https://spdx.org
```

and not 

```
 see [SPDX](https://spdx.org)
```

because I think it was a restriction of the ISO standard formatting (IIRC).

But maybe things have evolved; I'll ask around.

bact:
Nice catch. Another different behavior between MkDocs and GitHub.

GitHub can automatically render this to a clickable link:

```markdown
https://www.iana.org/
```

For the same result, MkDocs needs this:

```markdown
[https://www.iana.org/](https://www.iana.org/)
```

See an example in https://spdx.github.io/spdx-spec/v3.0/model/Core/Datatypes/MediaType/

vargenau:
@zvr @kestewart @bact 

So I understand I should replace things like:
```
[https://spdx.org/licenses/exceptions-index.html](https://spdx.org/licenses/exceptions-index.html)
```
by
```
<https://spdx.org/licenses/exceptions-index.html>
```

What should I do with things like:
```
[RFC 2046](https://www.ietf.org/rfc/rfc2046.txt) 
```
Should I remove the link?

We have such kind of links in e.g.
https://spdx.github.io/spdx-spec/v3.0/serializations/

(file docs/serializations.md in https://github.com/spdx/spdx-spec)

zvr:
I'd say we pause this one till we know (and test) the exact production process, so that we know what the requirements will be. It makes no sense to make changes now and have to make other changes later.

bact:
Agree with @zvr on the pause, the knowledge of production process, and the test.

With that, we should make the knowledge and the test available and accessible more to the spec contributors.

More documentation on how spdx-3-model, spdx-spec, spec-parser, and mkdocs are working together to generate the final spec website are welcome. https://github.com/spdx/spdx-spec/pull/946 is part of the attempt.

bact:
Btw, it is totally understandable that we were not given the priority to such the documentation before this, since we were all focused on the release of 3.0. Now we have less pressure on that.

bact:
> So I understand I should replace things like:
> 
> ```
> [https://spdx.org/licenses/exceptions-index.html](https://spdx.org/licenses/exceptions-index.html)
> ```
> 
> by
> 
> ```
> <https://spdx.org/licenses/exceptions-index.html>
> ```

@vargenau From the machine point of view, both with get rendered perfectly with mkdocs.

From the human point of view, the later is more friendly for the human editor. It is less error prone as well, because you don't have to worry about misspelling another URL.

> What should I do with things like:
> 
> ```
> [RFC 2046](https://www.ietf.org/rfc/rfc2046.txt) 
> ```

If it is the ISO requirement to have the URL in clear (to make the URL visible when printed on paper), then we should expand it.

It can be something like:

```markdown
RFC 2046 <https://www.ietf.org/rfc/rfc2046.txt>
```

or in a style according to our style guide (which I'm not sure if we have one).

zvr:
So, the final outcome is that we can (and must) have links not showing, like:

> see [SPDX](https://spdx.org/)

and the production process will translate it to footnotes in the ISO standard.

model/Core/Datatypes/MediaType.md
```diff
@@ -8,10 +8,11 @@ Standardized way of indicating the type of content of an Element. A String const
 
 ## Description
 
-A MediaType is a string constrained to the RFC 2046 specification. It provides a standardized
+A MediaType is a string constrained to the [RFC 2046](https://www.ietf.org/rfc/rfc2046.txt) specification. It provides a standardized
 way of indicating the type of content of an Element.
 
-A list of all possible media types is available at https://www.iana.org/assignments/media-types/media-types.xhtml.
+A list of all possible media types is available at
+[https://www.iana.org/assignments/media-types/media-types.xhtml](https://www.iana.org/assignments/media-types/media-types.xhtml).
```
zvr, line 15:
```suggestion
[IANA](https://www.iana.org/assignments/media-types/media-types.xhtml).
```

zvr (changes requested):

kestewart (approved):
Changes look reasonable to me.    Thanks for adding these.

bact:
All links are already got covered.
As time goes by, a change set in this PR eventually got reduced to one,
and that link is also got covered, see: https://spdx.github.io/spdx-spec/v3.0.1-draft/model/Core/Datatypes/MediaType/

We can close this PR.

bact:
Thank you @vargenau for the PR, as we learn along the process about subtle differences between Markdown renderers.

kestewart:
Closing PR, as links have been handled by other PRs.

PR #723, opened 2024-04-21
Doc: Fix list and nested list rendering issues in spec webpages + other Markdown format cleanup

bact:
## Background

Apart from minor incompatibility in Markdown-flavours, GitHub Markdown rendering engine tends to be more robust/error forgiving (can render things that are not a “well-formed” Markdown) but mkdocs (a tool that we use to generate HTML for the spec website) is not so. So some spec description that looks good on GitHub may looks not as good on our spec website.

Compare
- GitHub: https://github.com/spdx/spdx-3-model/blob/main/model/Software/Properties/copyrightText.md and
- Spec website: https://spdx.github.io/spdx-spec/v3.0/model/Software/Properties/copyrightText/

You can see that the list in NOASSERTION paragraph is rendered badly on the spec website.

This is because the .md file does not have a blank line between the list and the paragraph before the list. GitHub can tolerate this, but mkdocs cannot.

This PR is trying to avoid the issue (and any issue in the future) by formatting our Markdown files to be well-formed as much as possible.

## Changes

Cleanup Markdown formatting to fix document rendering issues on https://spdx.github.io/spdx-spec/v3.0/ , especially on lists and nested lists.

- Keep one blank line before and after list (markdownlint [MD032](https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md032.md))
  - This fixes an issue of the list is not rendering as list (but shown as plain text, like "* ... * ...") in:
    - https://spdx.github.io/spdx-spec/v3.0/model/Licensing/Licensing/
      - Reported in #720
    - https://spdx.github.io/spdx-spec/v3.0/model/Software/Properties/copyrightText/
- Only use `-` for list item (removing the mix of `*` and `-` in one file) ([MD004](https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md004.md))
  - `-` is preferred, as `*` is also used by emphasis (`*emphasis*`) and in cardinality (`1..*`)
- Keep list indentation spaces as expected ([MD007](https://github.com/DavidAnson/markdownlint/blob/v0.34.0/doc/md007.md))
  - This fixes nested list rendering issue in https://spdx.github.io/spdx-spec/v3.0/model/Extension/Classes/Extension/
    - Currently the nested list one level under Point 3 is displayed as Point 4, 5, 6, 7, which is wrong. See [source](https://github.com/spdx/spdx-3-model/blob/main/model/Extension/Classes/Extension.md).
[...]

PR #758, opened 2024-04-29
Fix list syntax.

vargenau:
Use "-" instead of "*", add empty line to be more compatible with mkdocs.


zvr:
Are these not part of #723 ?

bact:
With [this recent merge](https://github.com/spdx/spdx-3-model/commit/802137f6a3156d831cffe0f51b62e11c0de0eb50), this issue is fixed already.

PR #764, opened 2024-05-07
Relationships: Note class contraints on security types

puerco:
This commit adds language to note the class constraints on security relationship types.

/cc @rnjudge 

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@stacklok.net>


puerco:
@rnjudge I added slightly different verbiage than what we had before trying to avoid mixing "class" and "type". Let me know if you think this is better than the previous one and I can even all security relationships.

goneall (approved):
LGTM

PR #777, opened 2024-05-31
Add link SPDX License List

bact:
Make it more convenient for the spec readers

- For "SPDX License List", add link to https://spdx.org/licenses/
- For "SPDX Exceptions List", add link to https://spdx.org/licenses/exceptions-index.html
- For "custom-defined license", add link to https://spdx.github.io/spdx-spec/v3.0/model/ExpandedLicensing/Classes/CustomLicense/
- Add `<`..`>` around bare URLs in License profile to make it clickable on mkdocs


swinslow:
Hi @bact, thanks for this! 

One comment: Please note that a "custom-defined license" is different from a license expression:

* A license expression is a string representation of one or more licenses, potentially combined using "AND" / "OR" and similar operators.
* A "custom-defined license," in the sense used here, is a single license that is _not_ on the SPDX License List. That custom-defined license might be represented in the SPDX 3.0 object model as a [`CustomLicense`](https://github.com/spdx/spdx-3-model/blob/main/model/ExpandedLicensing/Classes/CustomLicense.md); or it might be represented in the license expression syntax as a `LicenseRef-` license (as described in some parts of the Annex D that you linked to).

So I probably wouldn't suggest linking to the license expression syntax annex for that item; probably best to link to the `CustomLicense` class.

Other than that, I'm generally +1 to this PR (though I note that it includes a couple of edits to serialization/rdf.md which are beyond the scope of the license-related profiles, so @zvr or @goneall may want to take a quick look at that also).

bact:
Thanks @swinslow for clarification. I will revise accordingly based on your suggestions.

bact:
Updated per @swinslow suggestions.

- "custom-defined license" now links to https://spdx.github.io/spdx-spec/v3.0/model/ExpandedLicensing/Classes/CustomLicense/
- Also links "SPDX Exceptions List" to https://spdx.org/licenses/exceptions-index.html

model/SimpleLicensing/Classes/AnyLicenseInfo.md
```diff
@@ -12,7 +12,8 @@ according to the SPDX license expression syntax.
 
 An AnyLicenseInfo is used by licensing properties of software artifacts.
 It can be a NoneLicense, a NoAssertionLicense,
-single license (either on the SPDX License List or a custom-defined license);
+single license (either on the [SPDX License List](https://spdx.org/licenses/)
+or [a custom-defined license](https://spdx.github.io/spdx-spec/v3.0/model/ExpandedLicensing/Classes/CustomLicense/));
```
zvr, line 16: Links should never include version number, since we don't want to update them continuously. (and we're at 3.0.1 now).

Please always use relative links.

model/ExpandedLicensing/Properties/deprecatedVersion.md
```diff
@@ -9,8 +9,9 @@ identifier was deprecated.
 
 ## Description
 
-A deprecatedVersion for a ListedLicense or ListedLicenseException on the SPDX
-License List specifies which version release of the License List was the first
+A deprecatedVersion for a ListedLicense or ListedLicenseException on the
+[SPDX License List](https://spdx.org/licenses/)
```
zvr, line 13: To be more accurate, licenses are on the license list and exceptions are on the exception list.

zvr (changes requested):

model/SimpleLicensing/Classes/AnyLicenseInfo.md
bact: Thank you. Will fix all instances to relative ones.

zvr (approved):

goneall (approved):
Looks like all the requested changes have been made.  LGTM.

PR #789, opened 2024-06-16
Remove Software/contentType (duplicative)

### LINKED ISSUE #781: Do we want / need to have contentType defined in Software and Core?
It looks like we have the property `contentType` defined in both the Core and the Software profile.

The definitions look similar.

From doing a bit of research, it looks like both were added with this commit: https://github.com/spdx/spdx-3-model/commit/b1988bfd98f77972f01740556e916461d61237ec

Was this intentional?

If not, should we fix it in the 3.0.1 release?

bact:
- Update the `Core/Properties/contentType.md` with wording from `Software/Properties/contentType.md`, adjusted to accommodate both Element and Property
- Update `Core/Datatypes/MediaType.md` to accommodate both Element and Property, also add string examples
- Remove `Software/Properties/contentType.md` file
- Update `Software/Classes/File.md` to refer to `/Core/contentType` instead of `contentType`
- Update examples: `file1.json` and `spdx_document2.json`

This should resolve #781


goneall (approved):
LGTM - Thanks @bact

goneall:
@JPEWdev - This would be a breaking change for existing 3.0.0 users.  We're thinking of merging this into 3.0.1 - would this be a problem for your Yocto work?

JPEWdev:
No, we aren't using that field. Is it possible to "hide" it in the spec, but leave it in the RDF so it's backward compatible maybe?

goneall:
> No, we aren't using that field. Is it possible to "hide" it in the spec, but leave it in the RDF so it's backward compatible maybe?

@JPEWdev - good to know.

For the RDF, we could keep the `Software/contentType` and use the `owl:sameAs` to associate it with the `core/contentType`.  This would imply a bit of a hack to the spec parser to capture this - or a more extensive change to the markdown spec and the spec parser.

My opinion is if we get this changed quickly  and release 3.0.1 with the changes, it won't impact many folks since we're in the early phases of 3.0 adoption.

I'm going to go ahead and merge this change.

@zvr if you have time - feel free to update the spec parser.

PR #791, opened 2024-06-24
trainedOn relationship type desc: "trained by" -> "trained on"

bact:
Change from:

```text
The `from` Element has been trained by the `to` Element(s)
```

to:

```text
The `from` Element has been trained on the `to` Element(s)
```

- This will make the description consistent to the name of the type.
  - `testedOn` type description also use the phrase "tested on"
- Also to avoid confusion. "by" may signify that the `to` is an Agent, which is not the case here.


bact:
Approved during AI Team meeting 2024-06-26

kestewart (approved):
LGTM

kestewart:
Approving, since documented in AI meeting there was approval.

PR #794, opened 2024-06-29
Fix types in Security example snippets

bact:
Revise example snippets ("Syntax" section) in *VulnAssessmentRelationship classes:

- Put `xsd:decimal` value inside a quote
- Put `CvssSeverityType` entry value to lower case, per https://spdx.github.io/spdx-spec/v3.0/model/Security/Vocabularies/CvssSeverityType/
- Change `to` value to array (cardinality ..*) (in CvssV4VulnAssessmentRelationship)
- `externalRefs` -> `externalRef`
- `externalIdentifiers` -> `externalIdentifier`
- `"externalIdentifierType": "securityAdvisory"` -> `"externalIdentifierType": "securityOther"`
  - There is no `securityAdvisory` entry in `ExternalIdentifierType` vocab (It is in `ExternalRefType` vocab)
- Add `security_` prefix to security properties
- `@id` -> `spdxId`
- `@type` -> `type`


bact:
- See also #799

goneall (approved):
LGTM - Verified that the severity types should be lower case

kestewart (approved):
Thanks for cleaning this up.

PR #795, opened 2024-06-29
Update EpssVulnAssessmentRelationship properties

### LINKED ISSUE #783: publishedTime redefined in EpssVulnAssessmentRelationship
Within the Security profile, both [EpssVulnAssessmentRelationship](https://github.com/spdx/spdx-3-model/blob/e542684fc292183227bd03c87e068a8a52578f19/model/Security/Classes/EpssVulnAssessmentRelationship.md?plain=1#L55) and its parent class [VulnAssessmentRelationship](https://github.com/spdx/spdx-3-model/blob/e542684fc292183227bd03c87e068a8a52578f19/model/Security/Classes/VulnAssessmentRelationship.md?plain=1#L26) define publishedTime.

Looks like publishedTime within EpssVulnAssessmentRelationship should be an external property restriction instead as the only difference is minCount: 1.

bact:
- In `EpssVulnAssessmentRelationship`,
  move `publishedTime` to "External properties restrictions" section,
  as it is already defined in `VulnAssessmentRelationship`
- In `VulnAssessmentRelationship`, 
  delete `/Core/Relationship/to` in "External properties restrictions" section,
  as the declared cardinality (1..*) is the same as one in `/Core/Relationship`

This will resolve #783


goneall (approved):
LGTM

Thanks @bact

bact:
There was a mistake earlier.

- `publishedTime` should now be `/Security/VulnAssessmentRelationship/publishedTime` in the External properties restrictions section

Fixed with the most recent commit.

goneall:
> There was a mistake earlier.
> 
> * `publishedTime` should now be `/Security/VulnAssessmentRelationship/publishedTime` in the External properties restrictions section
> 
> Fixed with the most recent commit.

I missed that too - good catch @bact

kestewart:
@rnjudge - can you please review.

rnjudge:
Discussed in July 16th call. LGTM.

PR #800, opened 2024-07-18
Change URIs to 3.0.1

zvr:
This updates the URIs of the model terms to refer to 3.0.1.

As decided in the tech call of 16 July 2024.


kestewart:
We need to make sure this is not going break the publishing before applying.  
Discuss in Tech call.

goneall (approved):
LGTM

goneall:
Approved on tech call 30 July 2024

PR #801, opened 2024-07-18
Add SPDX in glossary as its meaning has changed between SPDX 2 and 3

vargenau:
That was suggested in a comment in https://github.com/spdx/spdx-spec/pull/954


kestewart (approved):

goneall (approved):
LGTM

PR #802, opened 2024-07-18
JSON-LD Serialization doc: Fix relationship type name, Add validation info

bact:
- In serialization/README.md
  - Remove serialization information section, as it is already at https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/serializations.md
  - Sort/categorize use cases/serialization examples list
  - Fix name `ExternalReference` -> `ExternalRef`
- In serialization/jsonld.md:
  - Update IRIs to 3.0.1 
  - In Validation section, add link to serialization/jsonld/validation.md
  - Add Examples section and add link to serialization/jsonld/examples.md
- Rename `serialization/json_ld/` directory and `serialization/json-ld.md` to `jsonld`
  - For consistency with `jsonld` in [`spdx-spec`](https://github.com/spdx/spdx-spec/tree/development/v3.0.1/examples/jsonld) and [license-list-data](https://github.com/spdx/license-list-data/tree/main/jsonld) repos
- Make heading capitalization consistent across directory

This is a rework of #766


goneall (approved):
LGTM

serialization/README.md
```diff
@@ -69,12 +71,14 @@ Examples of how to serialize the following cases:
...
-- Adding vulnerability elements with security external reference types including FIXME
-- Adding hasAssociatedVulnerability relationship between a vulnerability element and a software profile element
-- Adding multiple hasAssessmentFor relationships for vulnerability assessment relationships between vulnerability element and package element for VEX, CVSS, etc. to communicate, e.g.,
-  - changes to a vulnerability element’s status affecting a specific package element using VEX (Vulnerability Exploitability eXchange)  (see the serialized examples listed in Syntax under each SPDX-3-MODEL/model/Security Class definition)
-  - how a vulnerability element may be fixed for a particular software package element
-  - results of a vulnerability scan or audit
-- Adding foundBy, publishedBy, reportedBy relationships between an agent element and a vulnerability element
-
-### Build Profile use cases to be added here
+- Adding vulnerability elements with security external reference types including *FIXME*
```
kestewart, line 74: *FIXME* still needs to be resolved.

goneall:
@rnjudge On line 74 of the README, there is a `FIXME` placeholder for external reference types for security.

Can you list which external references should be added?

Here's a link to the vocab: https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/ExternalRefType.md

bact:
I cannot find `spdxDocumentForArtfact` (or `spdxDocumentForArtifact`) relationship type mentioned here:

> A Relationship of type spdxDocumentForArtfact can be used to relate an SpdxDocument to one more serialized for of the SpdxDocument.

From #542, I believe it should be `serializedInArtifact`:

> A Relationship of type serializedInArtifact can be used to relate an SpdxDocument to one more serialized for of the SpdxDocument.

My apology to tag people from #542
@goneall @JPEWdev @sbarnum @zvr can you please confirm this? Thank you.

bact:
Also this one on "XCollection"

> When serializing the physical SpdxDocument, any properties of the logical element which can be represented using native mechanisms in the specific serialization format (e.g. @context prefixes in JSON-LD in place of the namespaceMap) should use these mechanisms and the remaining properties should be serialized within the “XCollection” element itself.

What is "XCollection" ?

JPEWdev:
> Also this one on "XCollection"
> 
> > When serializing the physical SpdxDocument, any properties of the logical element which can be represented using native mechanisms in the specific serialization format (e.g. @context prefixes in JSON-LD in place of the namespaceMap) should use these mechanisms and the remaining properties should be serialized within the “XCollection” element itself.
> 
> What is "XCollection" ?

I believe it was the placeholder name for SpdxDocument, and the find/replace missed this one. It should be `SpdxDocument`

serialization/jsonld.md
```diff
@@ -44,29 +44,37 @@ discussion).
 The SPDX organization provides a global JSON-LD @context file that MUST be used
 universally for all SPDX JSON-LD files of a given SPDX version.
 
-The context is available under https://spdx.org/rdf/3.0.0/spdx-context.jsonld
+The context is available under <https://spdx.org/rdf/3.0.1/spdx-context.jsonld>
```
JPEWdev, line 47: Last I checked, we are not changing this URL for micro version?

goneall: From the [2024-07-16 minutes](https://github.com/spdx/meetings/blob/main/tech/2024/2024-07-16.md), it looks like the decision was made to include the micro version?

I wasn't on the call myself, so I can't confirm the decision.

bact:
> > What is "XCollection" ?
> 
> I believe it was the placeholder name for SpdxDocument, and the find/replace missed this one. It should be `SpdxDocument`

Thank you Joshua. I have replaced XCollection with SpdxDocument here and in the spec doc.

serialization/jsonld.md
bact: I'm sure that the decision is to use micro/patch version (and it is already applied in #800). But I'm not sure about this specific URL for context file.

Maybe the reason we like to keep the URL of context file 3.0.0 because we don't want to create too much burden for users to updating the URL in the future?

- If we change the context file URL every micro/patch version, delivering updated content in the new context file to user will be more difficult/slower for each version -- as the user has to specifically changing the URL.

- If we keep "3.0.0" name for any 3.0.x version, we'll basically use "3.0.0" here the way "3.0" supposed to be. The actual content from this "3.0.0" URL will be the latest 3.0.x.

At the same time, it's quite weird to have the URL that way - because the versioning is not consistent with our IDs (which is now https://spdx.org/rdf/3.0.1/terms/Core, etc. and can be https://spdx.org/rdf/3.0.2/terms/Core, etc. in the future ..)

I know that this is a breaking one, but does it too late to have just `https://spdx.org/rdf/3.0/spdx-context.jsonld` instead ?
- Burden to update context file URL still remain for the next update (from `https://spdx.org/rdf/3.0.0/spdx-context.jsonld` to `https://spdx.org/rdf/3.0/spdx-context.jsonld`), but that will be the last time?
- We can still keep using IDs 3.0.x if we want; things may be less challenging if we can update the content of context file regularly

serialization/README.md
bact: Remove for now

serialization/jsonld.md
bact: 2024-08-06 Tech call decided to go with `3.0.1`  https://github.com/spdx/meetings/pull/757

serialization/README.md
bact: I have now put it as:

> - Adding vulnerability elements with security external reference types
  including `securityFix`, `vulnerabilityDisclosureReport`,
  and `vulnerabilityExploitabilityAssessment`

will that work?

bact:
After the merge of https://github.com/spdx/spdx-spec/pull/1016, the [Model and serializations chapter](https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/serializations.md) in the spec now has the content from the Serialization information section in serialization/README.md, so I will remove them from the file.

kestewart (approved):
Looks good, but could please resolve the conflict, so we can merge.

goneall:
@bact - could please resolve the merge conflicts - once completed, we can merge

bact:
Thank you. @goneall @kestewart conflicts resolved.

PR #806, opened 2024-07-25
Update cpe23 link in ExternalIdentifierType

bact:
- Make cpe23 URL matched with one in the References chapter (see https://github.com/spdx/spdx-spec/pull/1006 )
- Add full names to some entries.
  - For example, `swid` here is specifically a concise SWID one according to the ref. So add "Concise Software Identification tag" to the description.
- Update obsoleted ref
  - `swid` ref to a pre-RFC Internet Draft that expired on 8 Sep 2022, updated to RFC 9393 which replace it
- Make all URLs clickable. Also add text labels to bare URLs, as required by OMG/ISO standard doc


model/Core/Vocabularies/ExternalIdentifierType.md
```diff
@@ -16,14 +16,14 @@ ExteralIdentifierType specifies the type of an external identifier.
...
+- cpe22: Common Platform Enumeration Specification 2.2 <https://cpe.mitre.org/files/cpe-specification_2.2.pdf>
```
zvr, line 19:
```suggestion
- cpe22: [Common Platform Enumeration Specification 2.2](https://cpe.mitre.org/files/cpe-specification_2.2.pdf)
```

model/Core/Vocabularies/ExternalIdentifierType.md
```diff
@@ -16,14 +16,14 @@ ExteralIdentifierType specifies the type of an external identifier.
...
+- cpe23: Common Platform Enumeration: Naming Specification Version 2.3 <https://csrc.nist.gov/publications/detail/nistir/7695/final>
+- cve: Common Vulnerabilities and Exposures identifiers, an identifier for a specific software flaw defined within the official CVE Dictionary and that conforms to the CVE specification as defined by <https://csrc.nist.gov/glossary/term/cve_id>.
```
zvr, line 21:
```suggestion
- cve: Common Vulnerabilities and Exposures identifiers, an identifier for a specific software flaw defined within the official CVE Dictionary and that conforms to the [CVE specification](https://csrc.nist.gov/glossary/term/cve_id).
```

zvr (changes requested):
Links should not be visible (I only did a couple)

bact:
> Links should not be visible (I only did a couple)

Thanks. I have made all of them invisible now.

goneall (approved):
LGTM

zvr (approved):

PR #807, opened 2024-07-25
Remove redundant maxCount of EpssVulnAssessmentRelationship

bact:
Fix spec-parser's "has same maxCount as the parent class" warning.

In class /Security/EpssVulnAssessmentRelationship, property /Security/VulnAssessmentRelationship/publishedTime has same maxCount as the parent class.


zvr (approved):

goneall (approved):
LGTM

PR #812, opened 2024-07-26
[Security] Add text labels to bare URLs

### LINKED ISSUE #792: percentile definition is specific to EPSS
Currently, while [`probability`](https://github.com/spdx/spdx-3-model/blob/main/model/Security/Properties/probability.md) is defined in a way that is not specific to EPSS data, [`percentile`](https://github.com/spdx/spdx-3-model/blob/main/model/Security/Properties/percentile.md) is.

`probability`:

> The probability score between 0 and 1 (0 and 100%) estimating the likelihood of exploitation in the wild in the next 30 days (following score publication).

`percentile`:

>The percentile between 0 and 1 (0 and 100%) of the current probability score, the proportion of all scored vulnerabilities with the same or a lower EPSS score.

As "EPSS score" in the definition of percentile is equivalent to "probability score"
(see "epss" in <https://www.first.org/epss/data_stats> for comparison), does it possible for us to redefine `percentile` as:

>The percentile between 0 and 1 (0 and 100%) of the current probability score, the proportion of all scored vulnerabilities with the same or a lower probability score.

While currently the only use of percentile is for the EPSS-related [`EpssVulnAssessmentRelationship`](https://github.com/spdx/spdx-3-model/blob/main/model/Security/Classes/EpssVulnAssessmentRelationship.md),
but this will make `percentile` definition more generic and open for other use cases as well -- just like how `probability` currently is.

bact:
For the purpose of OMG/ISO standard doc preparation.

This PR will fix #792 too.


bact:
@zvr @goneall @rnjudge if you have some time, please kindly review. Thank you.

model/Security/Properties/vectorString.md
```diff
@@ -17,7 +17,8 @@ Supports vectorStrings specified in all CVSS versions.
 
 String values for the vectorString range must only include the abbreviated form
 of metric names specified in CVSS specifications, e.g.
-[https://www.first.org/cvss/v4.0/specification-document#Vector-String](https://www.first.org/cvss/v4.0/specification-document#Vector-String)
+[Table 23: Base, Threat and Environmental Vectors](https://www.first.org/cvss/v4.0/specification-document#Vector-String)
```
goneall, line 20: I would remove the "Table: 23" since that is local to the Common Vulnerability Scoring System version 4.0: Specification Document.  Alternatively, we could prefix Table 23 with "Common Vulnerability Scoring System version 4.0: Specification Document", however that gets rather long.

goneall (changes requested):
One suggested change, otherwise looks good.

model/Security/Properties/vectorString.md
bact: Thank you. I have removed it.

bact: Oh. Actually. The full sentence is

> e.g. [Table 23: Base, Threat and Environmental Vectors](https://www.first.org/cvss/v4.0/specification-document#Vector-String) of Common Vulnerability Scoring System version 4.0: Specification Document.

bact: It is now read

> String values for the vectorString range must only include the abbreviated form of metric names specified in CVSS specifications, e.g. [Common Vulnerability Scoring System Vector String](https://www.first.org/cvss/v4.0/specification-document#Vector-String).

model/Security/Vocabularies/CvssSeverityType.md
```diff
@@ -8,7 +8,23 @@ Specifies the CVSS base, temporal, threat, or environmental severity type.
...
+CvssSeverityType specifies the Common Vulnerability Scoring System (CVSS)
+severity type, defined in the CVSS specifications as the textual representation
+of the numeric CVSS score.
+
+The severity type entries are inclusive of and applicable to enumerations found
+in
+[Common Vulnerability Scoring System v3.0: Specification Document](https://www.first.org/cvss/v3.0/specification-document#Qualitative-Severity-Rating-Scale)
+and
+[Common Vulnerability Scoring System version 4.0: Specification Document](https://www.first.org/cvss/v4.0/specification-document#Qualitative-Severity-Rating-Scale).
```
rnjudge, line 19: Why are the links on separate lines and `in` and `and` have their own line? This seems like odd formatting.

model/Security/Vocabularies/CvssSeverityType.md
```diff
@@ -8,7 +8,23 @@ Specifies the CVSS base, temporal, threat, or environmental severity type.
...
+
+CvssSeverityType is a mandatory field because baseSeverity is required in the
+[CVSS 3.0 schema](https://www.first.org/cvss/cvss-v3.0.json),
+[CVSS 3.1 schema](https://www.first.org/cvss/cvss-v3.1.json), and
+[CVSS 4.0 schema](https://www.first.org/cvss/cvss-v4.0.json).
```
rnjudge, line 24: Same question about each link having their own line... when this is rendered in markdown this will look odd. I think they can all go on the same line?

model/Security/Vocabularies/SsvcDecisionType.md
```diff
@@ -8,7 +8,9 @@ Specifies the SSVC decision type.
 
 ## Description
 
-SsvcDecisionType specifies the type of decision that's been made according to the Stakeholder-Specific Vulnerability Categorization (SSVC)  system [https://www.cisa.gov/stakeholder-specific-vulnerability-categorization-ssvc](https://www.cisa.gov/stakeholder-specific-vulnerability-categorization-ssvc)
+SsvcDecisionType specifies the type of decision that's been made according to
+the
+[Stakeholder-Specific Vulnerability Categorization (SSVC)](https://www.cisa.gov/stakeholder-specific-vulnerability-categorization-ssvc).
```
rnjudge, line 13: Same comment... why does `the` have it's own line? Should be on the same line as the link.

rnjudge (changes requested):
@bact Some general formatting that should probably be fixed but content looks good.

bact:
Thank you Rose, I will recheck all of them again for formatting.

model/Security/Vocabularies/CvssSeverityType.md
bact: Thank you.

In this case, a newline is introduced for each item in a list, to make it easier to track changes for each item in the list. See for example: https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/index.md (in Raw).

Markdown will combine all lines in the same paragraph together.

For this particular file, it will be rendered like this in the production,
see: https://bact.github.io/spdx-spec/v3.0.1/model/Security/Vocabularies/CvssSeverityType/

Screenshot 2024-08-07 at 11 17 43

A newline can also be generally introduced when a line reach 80 characters limit, this is to accommodate text editor; and for the diffs, shorter lines mean the change set will be smaller and the change log will be easier to track.

For example, consider there's one character changing in a long paragraph. If it is a long one-line, the entire paragraph will be pulled into the change set. If there are a number of smaller lines, only one short line with the changing character will be included in the change set.

[This commit](https://github.com/spdx/spdx-spec/commit/0d6edf527636c60107edd4590eb95b376aa05d00) from the index.md example above can be an extreme example of having newlines vs having long paragraph in one line.

model/Security/Classes/SsvcVulnAssessmentRelationship.md
```diff
@@ -10,7 +10,9 @@ Provides an SSVC assessment for a vulnerability.
 
 An SsvcVulnAssessmentRelationship describes the decision made using the
 Stakeholder-Specific Vulnerability Categorization (SSVC) decision tree as
-defined on [https://www.cisa.gov/stakeholder-specific-vulnerability-categorization-ssvc](https://www.cisa.gov/stakeholder-specific-vulnerability-categorization-ssvc).
+defined by
+[CISA Stakeholder-Specific Vulnerability Categorization Guide](https://www.cisa.gov/stakeholder-specific-vulnerability-categorization-ssvc).
```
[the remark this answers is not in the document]
bact: The entire `[ ... ]( ... )` link is very long, so I put it in its own line.

When rendered, they will stick together, like this:

Screenshot 2024-08-07 at 11 27 35

(live link: https://bact.github.io/spdx-spec/v3.0.1/model/Security/Classes/SsvcVulnAssessmentRelationship/ )

model/Security/Properties/decisionType.md
```diff
@@ -4,11 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provide the enumeration of possible decisions in the Stakeholder-Specific Vulnerability Categorization (SSVC) decision tree [https://www.cisa.gov/sites/default/files/publications/cisa-ssvc-guide%20508c.pdf](https://www.cisa.gov/sites/default/files/publications/cisa-ssvc-guide%20508c.pdf)
+Provide the enumeration of possible decisions in the
```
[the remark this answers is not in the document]
bact: To have the link in its own line and when there's an update in the link, the change set will be limited to the line with the link.

It will be rendered like this.

Screenshot 2024-08-07 at 11 29 43

https://bact.github.io/spdx-spec/v3.0.1/model/Security/Properties/decisionType/

model/Security/Vocabularies/CvssSeverityType.md
bact: Please see a comment below for CvssSeverityType.md. Thank you.

bact:
Thank you Rose for detailed reviews concerning newlines, I have replied them all on why and with screenshots demonstrating how they will be rendered.

goneall (approved):
New updates look good - Thanks @bact

kestewart (approved):
Rationale for line changes and evidence makes sense.  Moving forward with this.

PR #817, opened 2024-07-30
Fix typos: hasPrerequsite -> hasPrerequisite, and others

bact:
- Fix typo in hasAddedFile: "is is" -> "is"
- Fix typo in type name: "hasPrerequsite" -> "hasPrerequisite"
- Make fullstops and commas in of all types in RelationshipType consistent.
  - some "during a LifecycleScopeType period" have comman before it, some don't
  - some description with one sentence have fullstop, some don't

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -25,61 +25,61 @@ name completes the sentence:
 ## Entries
 
 - affects: (Security/VEX) The `from` vulnerability affect each `to` Element. The use of the `affects` type is constrained to `VexAffectedVulnAssessmentRelationship` classed relationships.
-- amendedBy: The `from` Element is amended by each `to` Element
-- ancestorOf: The `from` Element is an ancestor of each `to` Element
-- availableFrom: The `from` Element is available from the additional supplier described by each `to` Element
-- configures: The `from` Element is a configuration applied to each `to` Element during a LifecycleScopeType period
-- contains: The `from` Element contains each `to` Element
-- coordinatedBy: (Security) The `from` Vulnerability is coordinatedBy the `to` Agent(s) (vendor, researcher, or consumer agent)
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

bact:
@goneall I have pushed another commit after your approval to fix: hasPrerequsite -> hasPrerequisite.

There's a typo of this hasPrerequisite relationship type as well in the 2.3->3.0 translation table. I will open another PR to fix that together with the strikethrough formatting.

goneall (approved):
Good catch @bact 

This could be a breaking change if anyone is using the incorrectly spelled 3.0.0 version.

That being said, I believe we should go ahead and merge this in now so we don't propagate the miss-spelled property further.

@zvr @kestewart - agree?

bact:
> This could be a breaking change if anyone is using the incorrectly spelled 3.0.0 version.
> 
> That being said, I believe we should go ahead and merge this in now so we don't propagate the miss-spelled property further.

For anyone using 3.0.0, it will still be the same name with typo, as this PR will apply only to 3.0.1 (and we have separate IRIs for the two). Unless we like to retrospectively change this in 3.0.0 as well.

I think if it is a human author, they may spot the typo and corrected it by themselves (so lesser degree of an issue?). But if it's a machine or machine-assisted, very likely that it will just follow the mis-spelled information in the schema file (which is considered correct).

The Dots tool, for example, is unfortunately affected by this typo:

Screenshot 2024-07-31 at 20 08 10

bact:
For "hasPrerequsite" -> "hasPrerequisite",
if needed, we can also put an errata like one in [2.3.1](https://github.com/spdx/spdx-spec/blob/2326abefbdeb20b995fb9feab8e1a8f6542bf8dd/chapters/package-information.md#7241-description):

> Cardinality 	0..1
NOTE: The cardinality was incorrectly printed as 0..* in the original publishing of the 2.3 specification

zvr (approved):

PR #819, opened 2024-08-01
licenseXml: Link license-list-XML to a specific version (v3.24.0)

bact:
- Link to v3.24.0 of license-list-XML
- Use "SPDX License List XML Schema" name, for ref text label, taken from a heading in [schema/ListedLicense.xsd](https://github.com/spdx/license-list-XML/blob/v3.24.0/schema/ListedLicense.xsd)
  - `<xhtml:h1>SPDX License List XML Schema</xhtml:h1>`

### DIFF
--- a/model/ExpandedLicensing/Properties/licenseXml.md
+++ b/model/ExpandedLicensing/Properties/licenseXml.md
@@ -11,12 +11,11 @@ XML format.
 
 The license XML format is defined and used by the SPDX legal team.
 
-See the XML fields defined at
-[XML template fields](https://github.com/spdx/license-list-XML/blob/main/DOCS/xml-fields.md)
-for a text description.
+The formal schema definition is available at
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM - Thanks @bact

bact:
@swinslow May needs your review for this change. Thank you.

swinslow:
Are we planning to keep updating it to the latest License List release repeatedly? E.g., I'm likely going to be pushing 3.25.0 in a few days, so this would be out of date soon. Does it make sense to just keep it pointing to the main branch?

bact:
@swinslow I'm actually not sure. Looks like GitHub repo links are perceived as volatile/not-fixed, and a permalink is more preferred. That's why we trying to link to a specific version or tag.

Ping @zvr @goneall for thoughts

kestewart:
Historically we've along with the list copied into the spec, so I'm seeing no difference with referring to a fixed version (that does become out of date), but at least it's a known set at a point in time.

bact:
@swinslow once there's a new version, we can have another PR to update this.

PR #796, opened 2024-07-02
NamespaceMap: Remove TODO and fix caps

bact:
- Remove "TODO" as the URLs are already updated. (although they are all pointing directly to the repo: https://github.com/spdx/spdx-3-model/ - is this something we are preferred?)
- Also use 4-space indent for list, to keep the multiple paragraphs rendered correctly.

Issue #61, opened 2023-01-10
How to add profile data to an existing SBOM or Element


goneall:
This scenario came up in the Asia SPDX tech call on Jan 9, 2023.

An entity which is in the middle of a software supply chain receives an SPDX document and wishes to add profile information (e.g. the Build profile) before passing the information on to downstream consumers.

In SPDX 2.X, you would produce an SPDX Document which has an `AMENDS` relationship to the supplied SPDX document.

Since each element has creator information and can be independent of the Document, I would like to document the best practice for this scenario.

A proposal would be that each Element (be it a Package, SBOM, File, etc.) would be updated in the new document with the Profile information.  The Profile would also be added to the Creator information for the element.  An `AMENDS` relationship would be added from the new Element to the previous Element using the ExternalMap to reference the previous Element.

There are a couple of downsides to the above proposal:
- It is difficult to tell exactly what has changed, you would essentially need to "diff" the Elements
- You end up copying all the information from one element to another

I'm open to any other proposals or variations on a theme.

Below is an example:

Original Supplier element:
```
{
  ...
  "elements": [
    "@type": "Package",
    "@id": "urn:originalsupplier.org:packagea#elementid",
    "creationInfo": { ... "profile": ["core", "software"] ...
    "name": "myPackage"
    ...
    ],
   ...
}
```

Intermediate supplier element
```
{
  ...
  "externalMap": [{"elementId": "urn:originalsupplier.org:packagea#elementid"}],
  "elements": [
    "@type": "Package",
    "@id": "urn:intermediateSupplier.org:packagea#uniqueid2",
    "creationInfo": { ... "profile": ["core", "software", "build"] ...
    "name": "myPackage",
    "buildDate": "2022-12-13T18:36:37.793Z",
    ...
    ],
   ...
}
```

davaya:
This solution assumes that multiple inheritance is allowed: Package is defined in the Software profile but the Build profile can add properties to it.

An alternate solution would allow only single inheritance: the Build profile defines a BuildPackage type that contains just properties relevant to Build, including a required "sourcePackage" property that points to the original supplier package.  That eliminates 1) multiple inheritance, 2) an extra AMENDS relationship element, and 3) duplicate properties in both original and build elements.

kestewart:
https://github.com/spdx/spdx-3-model/pull/249 resolves this.   Closing.   Please open a new issue if you spot additional concerns.

Issue #42, opened 2022-11-05
Can one parse a JSON containing Elements extended by an unsupported / unknown profile?


maxhbr:
Hey, I have looked at the following example taken from the [current png](https://github.com/spdx/spdx-3-model/commit/92c25fcb420934c5b32f7809cd46fc12d3d6182d) and I have a question related to it, regarding "one does not need to support every profile and just looses information that is not relevant".
```
{
  "@type": "SBOM",
  "@id": "urn:spdx.dev:null-sbom",
  "creationInfo": {
    "specVersion": "3.0",
    "created": "2022-05-02T20:28:00.000Z",
    "profile": ["core","software"],
    "dataLicense": "CC0",
    "createdBy": ["urn:spdx.dev:iamwillbar"]
  },
  "rootElements": ["urn:spdx.dev:spdx-tools-3.0.1"],
  "externalMap": [ <...> ],
  "elements": [ <...> ]
}
```

For this example lets assume that I do **not** have support for software profile. Since SBOM is basically a BOM I would assume that I still should be able to parse it (despite loosing some information, that is not relevant for me). 

**But how do I know that it is basically a BOM, the JSON does not contain that information?**

(A rather ugly workaround would be to have something like `"@type": "SBOM:BOM:Bundle:Collection:Element"` in the serialization.)

goneall:
@maxhbr Good question.  This is very related to a long running debate on how to treat serialization level information in the document.

Parsing implies that there is a deserialization going on of something that was serialized - let's call that an SPDXDocument for the purpose of this comment.

If the SPDXDocument has creation information and that creation information would represent the intent of the SPDXDocument creator .  The profile field in the SPDXDocument element could be used to guide the deserialization and would give you the answer in a very straightforward and high performance manner (e.g. you don't have to interpret each element and try to reverse engineer the profiles based on type etc.).

In the current model diagram - the "collection -> SpdxCollection?" class could be used for that purpose since it already contains information specific to the serialization.

This approach isn't as powerful as interpreting each individual element since each element could conceivably have a different profile, but I would find this approach to be rather difficult to implement.

maxhbr:
> Parsing implies that there is a deserialization going on of something that was serialized - let's call that an SPDXDocument for the purpose of this comment.

Yes, the assumption is that a SBOM was serialized to a json and its abbreviated version is above.

> If the SPDXDocument has creation information and that creation information would represent the intent of the SPDXDocument creator .

[...]

PR #821, opened 2024-08-09
NamespaceMap: fix TODO; "ElementID" -> "Element ID"

bact:
- Complete TODO on serialization information and NamespaceMap
  - Now link to https://spdx.github.io/spdx-spec/v3.0.1-draft/serializations/ (the path is relative, so it will be "v3.0.1" once v3.0.1 is published)
- Use 4-space indent in the numbered list, to keep the intended multiple-paragraph format
- "ElementID" -> "Element ID"
  - It is not a specific name.
  - "elementId" was found used as a property of an ExternalMap in code examples of these two issues from late 2022/early 2023:
    - https://github.com/spdx/spdx-3-model/issues/42
    - https://github.com/spdx/spdx-3-model/issues/61
    - But that property is more likely being called [`externalSpdxId`](https://spdx.github.io/spdx-spec/v3.0.1-draft/model/Core/Classes/ExternalMap/) now.
  - Another instance of "ElementID" is found in [tools-golang](https://github.com/spdx/tools-golang/blob/8574b91809e949442fa0cbf3174d9ae83cd69f60/spdx/identifier.go)
  - Cannot find "ElementID" anywhere else in the spec and the model. Only found in NamespaceMap and namespaceMap.

This is a rework of #796


goneall (approved):
Changes look good - I did notice that we have a link to the SPDX 3 model repo without a specific version or commit.  We can open a separate PR or issue if this needs to be solved for the ISO submission.

kestewart (approved):

PR #825, opened 2024-08-09
Remove spaces from image filenames, put images into its own folder

bact:
- Remove spaces from image filenames to reduce chance of issues occurred from not escaping the space (as %20, for example, in the URL; or something else in a shell script)
  - Filenames will be the same as in https://github.com/spdx/spdx-spec/tree/development/v3.0.1/docs/images 
- Put images in `images/` directory, so the root is less cluttered and repo visitor can immediately see the README
- There's no reference to these files (from within this repo), so no broken links introduced (within this repo)


zvr (approved):

kestewart (approved):
Appreciate the cleanup.

PR #826, opened 2024-08-10
Add Adler32 to hash algorithms

goneall:
Adler32 was added to SPDX 2.3 based on issue 624 (https://github.com/spdx/spdx-spec/issues/624).  This algorithm is not present in the SPDX 3.0 spec.

I searched the SPDX tech minutes to see if this was discussed for 3.0 and I could not find any record - so I assume this was an oversight.

### DIFF
--- a/model/Core/Vocabularies/HashAlgorithm.md
+++ b/model/Core/Vocabularies/HashAlgorithm.md
@@ -18,6 +18,7 @@ practically infeasible to invert.
 
 ## Entries
 
+- adler32: Adler-32 checksum is part of the widely used zlib compression library as defined in [RFC 1950])(https://www.ietf.org/rfc/rfc1950.txt) Section 2.3.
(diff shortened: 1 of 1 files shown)


model/Core/Vocabularies/HashAlgorithm.md
```diff
@@ -18,6 +18,7 @@ practically infeasible to invert.
 
 ## Entries
 
+- adler32: Adler-32 checksum is part of the widely used zlib compression library as defined in [RFC 1950])(https://www.ietf.org/rfc/rfc1950.txt) Section 2.3.
```
bact, line 21:
```suggestion
- adler32: Adler-32 checksum is part of the widely used zlib compression library as defined in [RFC 1950](https://www.rfc-editor.org/info/rfc1950) Section 2.3.
```
 
- Remove extra `)`
- Use https://www.rfc-editor.org/info/rfc1950 to match the style with other existing RFC URLs

bact (changes requested):

zvr (approved):

bact (approved):

kestewart (approved):

PR #828, opened 2024-08-10
Fix casing: "package URL" -> "Package URL" at the beginning of sentence

bact:
In an ExternalIdentifierType entry.


kestewart (approved):

goneall (approved):
LGTM

PR #829, opened 2024-08-10
Update CHANGELOG 2024-08-10

bact:
Note some changes that can be breaking or change model behavior


kestewart:
@bobmartin3000 - do we want last update date before it goes to OMG or not?

kestewart (approved):
Will remove "last update" after 3.0.1 reviews.

goneall (approved):
LGTM

PR #832, opened 2024-08-10
Add patch information for specVersion

### LINKED ISSUE #733: Document PATCH for specVersion
In specVersion, it is written:

The specVersion provides a reference number that can be used to understand how to parse and interpret an Element.
It will enable both future changes to the specification and to support backward compatibility.
The **major version number** shall be incremented when incompatible changes between versions are made
(one or more sections are created, modified or deleted).
The **minor version number** shall be incremented when backwards compatible changes are made.

I understand that for the spec versioning we are using SemVer, so with a PATCH version.

This should be documented.

bact:
Add this sentence to the description

"The patch version number shall be incremented when backward compatible bug fixes are made."

To fix #733

### DIFF
--- a/model/Core/Properties/specVersion.md
+++ b/model/Core/Properties/specVersion.md
@@ -4,18 +4,26 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Provides a reference number that can be used to understand how to parse and interpret an Element.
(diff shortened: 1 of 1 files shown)


kestewart (approved):

goneall (approved):
LGTM

PR #834, opened 2024-08-10
Fixes for PDF generation

### LINKED ISSUE #816: Add ISO number for SWHID
in Vocabulary / ExternalIdentifierType

### LINKED ISSUE #830: [PDF] Long lines are not rendered
When creating the PDF version of the spec, some lines in pre-formatted content are too long and get truncated on the page.

The solution is to re-word or split them.

zvr:
Fixes everything producing issues in PDF generation.

The main changes are:

- shortening of pre-formatted lines (e.g., having long URL in examples)
- changing 5th-level section from bold to italic
- also shortened lines and re-worded the algorithm for PackageVerification

Closes #816 
Closes #830


zvr:
@kestewart @goneall @bact 

You can see the rendered files at https://zvr.github.io/spdx-3-spec/ 

You can see and downloaded the current PDF at https://github.com/zvr/spdx-3-spec/blob/main/SPDXv3-draft.pdf

Obviously ignore content from "spec", this is about **model**.

Goodnight!

zvr:
Ah, looking at the diffs, I also changed "Syntax" to "Example" in classes where there was some JSON showing how the class is to be used.

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,18 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
```
kestewart, line 19: Why has this algorithm been changed without wider review? 
It's not clear to me that the new version is clearer - or even accurate.   Please revert.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -25,28 +25,28 @@ name completes the sentence:
 
 ## Entries
 
-- affects: (Security/VEX) The `from` vulnerability affects each `to` Element. The use of the `affects` type is constrained to `VexAffectedVulnAssessmentRelationship` classed relationships.
```
kestewart, line 28: I don't remember the discussion of removing the profiles that are infront of relationships.   When did this happen?

model/Software/Properties/packageUrl.md
```diff
@@ -9,30 +9,26 @@ Provides a place for the SPDX data creator to record the package URL string
 
 ## Description
 
-A package URL (commonly pronounced and referred to as "purl") is an attempt to
-standardize package representations in order to reliably identify and locate
-software packages. A purl is a URL string which represents a package in a
+A package URL is an attempt to standardize package representations
```
kestewart, line 12:
```suggestion
A package URL (commonly pronounced and referred to as "purl") is an attempt to standardize package representations
```

kestewart (changes requested):
Most looks good, and is an improvement, but please revert the PackageVerificationCode algorithm changes.   They have not been discussed and I'm not convinced what's written is an improvement.

Also, the profile designators on relationships were not discussed before they were removed.

model/Core/Classes/PackageVerificationCode.md
zvr: Well, let's see about the old one. Here are the issues:

- it starts by assigning `verificationcode = 0` and this is never used or mentioned before, since the final line computes `hashValue`
- it assigns `filelist` to an empty string and then reassigns it without using this empty string, so this assignment is useless
- it has templist as a string (not a list), appends to it values plus `/n` -- without defining this plus, and of course newlines are `\n`
- then it says "sort templist" without explaining how to sort a single string
- then it computes a `hashValue` by calculating `algorithm(filelist)` ... while having a comment that `algorithm()` reads a file and does not take content as argument.

I only went in to fix the very long comment line which did not fit in the page.

But reading the old pseudo-algorithm, I found out it makes no sense. @goneall , what do you think?

model/Core/Vocabularies/RelationshipType.md
zvr: What do these (profile names in vocabulary entries) even mean? We never explain this.

Do we mean to say that you can **not** use `testedOn` unless you also comply with the AI/Dataset profile? (either? both?). Or it was simply introduced there?

Anyway these are not consistent, since for example "(Licensing)" does not appear.

model/Software/Properties/packageUrl.md
zvr: This is already in the Annex defining PURL, but I can add it back if you insist.

You know that I am not very font of this, especially when our RDF uses `purl.org` which means something else...

model/Core/Vocabularies/RelationshipType.md
bact: Agree that the information inside (...) is not explained anywhere inside RelationshipType summary/description.

If to leave the (...)s, may have to add a short line describes what they are, plus adding missing (...)s, like the mentioned "(Licensing)".

kestewart: OK,  I'll go along with making it consistent.   Probably something we need to go into more details with as a section in "Using" at any rate.

model/Software/Properties/packageUrl.md
kestewart: If we don't have it,  it will open up questions, so rather keep the wording that's already been debated, rather than change it at this last minute.

model/Core/Vocabularies/RelationshipType.md
bact: This may related to "Translating from 2.3 to 3.0" table? https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/annexes/diffs-from-previous-editions.md#translating-from-23-to-30-8

The table shows `LifecycleScopeType` expected to be used with a relationship type.

bact: Have reading through the relationship type list. I think it's quite safe to remove the information in (...)s.

For all `(Security)`/`(Security/VEX)` relationship types, their description have the word "Vulnerability" inside, which already give a context. Removing info in (...)s should be fine.

For `hasEvidence: (Dataset)` relationship type, we can say that the type can be use in a wider context than only Dataset. So removing `(Datatset)` is ok.

> - hasEvidence: (Dataset) Every `to` Element is considered as evidence for the `from` Element (`from` hasEvidence `to`).

For `(AI, Dataset)` relationship types, they can be a bit problematic if the context inside (...)s are to be removed.

> - testedOn: (AI, Dataset) The `from` Element has been tested on the `to` Element(s).
> - trainedOn: (AI, Dataset) The `from` Element has been trained on the `to` Element(s).

Maybe not so much for `trainedOn`.
But for `testedOn`, it can be confused with `hasTest` and `hasTestCase`.

If we're going to remove info in (...)s, I suggest we add few more context in the descriptions of `testedOn` (and maybe also `trainedOn`).

bact (changes requested):

goneall (changes requested):
Fix for one typo requested, otherwise look good.

Agree the changes in the pseudocode for PackageVerificationCode is an improvement - but we probably could have kept the old version as well since the field is not encouraged for SPDX 3.0.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -76,11 +76,11 @@ name completes the sentence:
...
+- publishedBy: Designates a `from` Vulnerability was made available for public use or reference by each `to` Agent.
+- reportedBy: Designates a `from` Vulnerability was first reported to a project, vendor, or tracking database for formal identification by each `to` Agent.
+- republishedBy: Designates a `from` Vulnerability's details were tracked, aggregated, and/or enriched to improve context (i.e. NVD) by each `to` Agent.
 - serializedInArtifact: The `from` SPDXDocument can be found in a serialized form in each `to` Artifact.
-- testedOn: (AI, Dataset) The `from` Element has been tested on the `to` Element(s).
-- trainedOn: (AI, Dataset) The `from` Element has been trained on the `to` Element(s).
-- underInvestigationFor: (Security/VEX) The `from` Vulnerability impact is being investigated for each `to` Element. The use of the `underInvestigationFor` type is constrained to `VexUnderInvestigationVulnAssessmentRelationship` classed relationships.
+- testedOn: The `from` Element has been tested on the `to` Element(s).
+- trainedOn: The `from` Element has been trained on the `to` Element(s).
```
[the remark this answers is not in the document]
kestewart: Want to use for safety, so let's not special case it.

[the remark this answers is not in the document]
kestewart: Let's describe this in "Using" chapter at this point.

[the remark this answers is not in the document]
bact: Support the generic approach. Keep it the same way then.

goneall (approved):
LGTM

bact (approved):

goneall:
@zvr - Can you resolve the merge conflicts?  BTW - Not quite sure what got merged in to cause the conflict?

kestewart (approved):
All changes except PackageVerificationCode, are fine.   Based on discussion, we'll merge this and Kate will submit a PR to revert the text of the algorithm back to 3.0.0 version, with deficiencies address.

PR #835, opened 2024-08-11
Make "parameters" singular

zvr:
This makes the property of Build profile `parameter` singular.

Whenever we have a "dictionary" or Key-Value pairs,
we always use the singular form because in RDF
we have multiple such property instances.

Here's a `parameter` with this `key` and this `value`,
here's another,
and the class instance has all these properties (`0..*`).

See, for example https://github.com/spdx/spdx-3-model/blob/main/model/AI/Properties/hyperparameter.md

On the other hand, the plural mirrors the SLSA Provenance definitions,
where they use the plural form.

The point is moot, since this is related to SLSA Provenance v0.2
and already v1.0 is out, so presumably we will be updating our model shortly.

We will be having the same discussion about `externalParameters` and `internalParameters` defined there,
and whether we keep our decisions or follow conventions of others.

So, we can be consistent in our frozen standard version,
or follow external conventions,
or do one thing now and another when we update...
I do not feel very strongly either way,
but I wanted to raise this before freezing.


zvr:
.... and when you write PR comments on your desktop, lines are **not** merged into paragraphs, in contrast to what is happening in the `.md` files. Sorry.

zvr:
Aaaaarrgh... and this was supposed to be only the change in the `parameter` property.

PR #836, opened 2024-08-11
Make "parameters" singular

zvr:
This makes the property of Build profile `parameter` singular.

Whenever we have a "dictionary" or Key-Value pairs, we always use the singular form because in RDF we have multiple such property instances.

Here's a `parameter` with this `key` and this `value`, here's another, and the class instance has all these properties `(0..*)`.

See, for example https://github.com/spdx/spdx-3-model/blob/main/model/AI/Properties/hyperparameter.md

On the other hand, the plural mirrors the SLSA Provenance definitions, where they use the plural form.

The point is moot, since this is related to SLSA Provenance v0.2 and already v1.0 is out, so presumably we will be updating our model shortly.

We will be having the same discussion about `externalParameters` and `internalParameters` defined there, and whether we keep our decisions or follow conventions of others.

So, we can be consistent in our frozen standard version, or follow external conventions, or do one thing now and another when we update...  I do not feel very strongly either way, but I wanted to raise this before freezing.


zvr:
Actually... even if we don't change it to singular, we should definitely at least change the current text from:

> parameters is a key-value map of all build parameters

to something like 

> parameters is a key-value entry of a build parameter

since the property is definitely not a map of all key/values pairs; it only holds one.

kestewart:
Thanks for raising.   I defer to @nishakm ond @lumjjb.   Since we can have multiple "parameter" fields - I can live with either, but if we don't hear back from them before noon tomorrow (or when Bob's deadline is), I'm biasing towards leaving it as is.

kestewart (commented):
I'm fine with adjusting this now, so we don't have a problem down the road, but want to make sure that Nisha & Brandon see it and comment - so we don't have a "surprise"

bact (approved):
I agree with the property name and property description changes proposed here themselves. It makes things more consistent.

lumjjb:
as long as its able to encode multiple parameters in the dictionary. I am fine with the labeling change.

goneall (approved):
Good catch - I prefer to fix this early.  Based on the reply from @lumjjb - I'll assume this is OK to merge.

goneall:
We should add a line to https://github.com/spdx/spdx-3-model/blob/main/CHANGELOG.md noting this change.  @zvr could add the update to this PR or I can add it after merge (since it's getting rather late in Europe)

kestewart (approved):
Based on Brandon's comments,  agree let's update this.

bact:
> We should add a line to https://github.com/spdx/spdx-3-model/blob/main/CHANGELOG.md noting this change. @zvr could add the update to this PR or I can add it after merge (since it's getting rather late in Europe)

The change log PR is here #839

(Note that I has accidentally make a commit previously. It is already got reverted)

PR #842, opened 2024-08-12
Revert PackageVerificationCode.md to original algorithm, and improve readability/formatting.

kestewart:
Add back in line separators, else the sort won't work. 
Use revised property names.   Set default value for hashValue.

### DIFF
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -23,7 +23,7 @@ Algorithm:
 
     for all files in the package {
         if file is a packageVerificationCodeExcludedFile
-            skip it /* exclude SPDX analysis file(s) */
+            skip it  /* exclude SPDX analysis file */
         else
             append "algorithm(file)/n" to templist
     }
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
+    hashValue = 0
     templist = empty list of strings
 
-    for all files in the package
-        if file is an "excludes" file
+    for all files in the package {
+        if file is a packageVerificationCodeExcludedFile
             skip it /* exclude SPDX analysis file(s) */
         else
-            append algorithm(contents of file) to templist
+            append "algorithm(file)/n" to templist
+        }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
      
-    verificationcode = algorithm(valueslist)
+    hasValue = algorithm(valueslist)
 
 where `algorithm(string)` applies a hash algorithm on a string and returns the result in lowercase hexadecimal digits.
 
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -18,7 +18,6 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
-    hashValue = 0
     templist = empty list of strings
 
     for all files in the package {
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -18,7 +18,7 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
-    templist = empty list of strings
+    templist = ""
 
     for all files in the package {
         if file is a packageVerificationCodeExcludedFile
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -32,7 +32,10 @@ Algorithm:
     
     valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
      
-    hashValue = algorithm(valueslist)
+    if valuelist is empty
+       hashValue = 0
+    else
+       hashValue = algorithm(valueslist)
 
 where `algorithm(string)` applies a hash algorithm on a string and returns the result in lowercase hexadecimal digits.
 
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -32,7 +32,7 @@ Algorithm:
     /* remove separators from ordered sequence */
     valueslist = remove "/n"s from templist
      
-    if valuelist is empty
+    if valueslist is empty
        hashValue = 0
     else
        hashValue = algorithm(valueslist)
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -32,7 +32,7 @@ Algorithm:
     
     valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
      
-    hasValue = algorithm(valueslist)
+    hashValue = algorithm(valueslist)
 
 where `algorithm(string)` applies a hash algorithm on a string and returns the result in lowercase hexadecimal digits.
 
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -29,7 +29,8 @@ Algorithm:
 
     sort templist in ascending order by value
     
-    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
+    /* remove separators from ordered sequence */
+    valueslist = remove "/n"s from templist
      
     if valuelist is empty
        hashValue = 0
--- a/model/Core/Classes/PackageVerificationCode.md
+++ b/model/Core/Classes/PackageVerificationCode.md
@@ -26,7 +26,7 @@ Algorithm:
             skip it /* exclude SPDX analysis file(s) */
         else
             append "algorithm(file)/n" to templist
-        }
+    }
 
     sort templist in ascending order by value
     


model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
 
-    for all files in the package
-        if file is an "excludes" file
+    for all files in the package {
+        if file is a packageVerificationCodeExcludedFile
             skip it /* exclude SPDX analysis file(s) */
         else
-            append algorithm(contents of file) to templist
+            append "algorithm(file)/n" to templist
+        }
```
bact, line 29:
```suggestion
    }
```

This is a closing of `for` ?

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+
```
bact, line 32:
```suggestion

```

Remove extra spaces.

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
+        }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
      
-    verificationcode = algorithm(valueslist)
+    hasValue = algorithm(valueslist)
```
bact, line 35:
```suggestion
    hashValue = algorithm(valueslist)
```

bact:
I think the line separator is no longer needed, because in the new algorithm `templist` is a list of strings, not a string.

The initialization of older code:

https://github.com/spdx/spdx-3-model/blob/00045dc506a55f8ea2be87ed16b08c727c4bee1e/model/Core/Classes/PackageVerificationCode.md?plain=1#L22

The initialization of current code:
https://github.com/spdx/spdx-3-model/blob/3c03a8164b00f95751d98e702f99105fa023c2d9/model/Core/Classes/PackageVerificationCode.md?plain=1#L21

bact:
The code should be something like:

```
    hashValue = 0
    templist = empty list of strings

    for all files in the package {
        if file is a packageVerificationCodeExcludedFile
            skip it  /* exclude SPDX analysis file(s) */
        else
            append algorithm(file) to templist
    }

    sort templist in ascending order by value
    
    tempstring = join templist values
     
    hashValue = algorithm(tempstring)
```


Not sure if "join templist values" is obvious enough for the reader that this will return a string, so I rename `valueslist` to `tempstring`

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
     templist = empty list of strings
 
-    for all files in the package
-        if file is an "excludes" file
+    for all files in the package {
+        if file is a packageVerificationCodeExcludedFile
             skip it /* exclude SPDX analysis file(s) */
         else
-            append algorithm(contents of file) to templist
+            append "algorithm(file)/n" to templist
```
zvr, line 28: No, it's a list, not a string.
It has different elements, you don't need "separators".

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
+        }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
```
zvr, line 33: No, there should not be anything to remove. You simply join different elements.

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
+    hashValue = 0
     templist = empty list of strings
 
-    for all files in the package
-        if file is an "excludes" file
+    for all files in the package {
```
zvr, line 24: Using indentation for grouping works fine, why introduce brackets in this pseudo-language?

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
+    hashValue = 0
```
zvr, line 21: Why setting it to an initial value, if you plan to overwrite it with an assignment at the end? This is never used.

model/Core/Classes/PackageVerificationCode.md
zvr, line 29: Not needed, indentation works.

zvr (changes requested):

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
         else
-            append algorithm(contents of file) to templist
+            append "algorithm(file)/n" to templist
+    }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
```
bact, line 33:
```suggestion
    tempstring = join templist values
```

Revert to use list

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
      
-    verificationcode = algorithm(valueslist)
+    hashValue = algorithm(valueslist)
```
bact, line 35:
```suggestion
    hashValue = algorithm(tempstring)
```

Match the new variable name "tempstring" from another suggestion.

bact (changes requested):
I cannot make suggestions over deleted lines, so have to break down the suggestions to small chunks.

These suggestions are from the code in https://github.com/spdx/spdx-3-model/pull/842#issuecomment-2283057878 -- @goneall is this the one you agreed?

model/Core/Classes/PackageVerificationCode.md
kestewart: This introduces variability on how the join is implemented, and in which language, and which underlying runtimes the language is drawing on. 

See: https://stackoverflow.com/questions/914233/what-is-the-underlying-data-structure-for-python-lists, https://www.geeksforgeeks.org/arraylist-in-java/, ...

Since we're hashing  contents - the implementations matter. 

This was the trap we fell into in 1.1 --> 1.2 - and we should not be introducing it again, esp. for a field that is just meant for compatibility with 2.3.  That this was not discussed before implementing, and could have saved everyone time if it was, especially this late in the cycle, when we're all stressed, and folded into a "pdf formatting change".

model/Core/Classes/PackageVerificationCode.md
kestewart: The byte specific implementation of list elements (of strings) varies depending on underlying language (JAVA, Python) as well it may have be impacted by the platform (linux, macOS,windows)  the library is implemented on.  

The goal here was to  is to be compatible with 2.3 - not to create a "better" algorithm.

model/Core/Classes/PackageVerificationCode.md
kestewart: "join" is implemented differently, based on language.   Each language may draw on underlying runtimes with different behaviors at the byte level.   Since we're taking a hash,  byte level matters.  Goal here was to get rid of the "differences" by being very basic.

model/Core/Classes/PackageVerificationCode.md
kestewart: This is tied to the above one with the "join" - see reason why not to use join.

model/Core/Classes/PackageVerificationCode.md
kestewart: Style preference.   Original had {, };   this should have been changed earlier and been in a separate PR.

kestewart:
"join" is implemented differently, based on language (JAVA, C++, Python, etc.). Each language may draw on underlying runtimes with different behaviors at the byte level for the datastructures. Since we're taking a hash, byte level matters. Goal here was to get rid of the "differences" by being very basic with strings, and compatibble with what should have been generated in a 2.X file.   

This is reverting the change that was introduced in the "pdf fix" that was not discussed in advance.    If a change the algorithm was requested, it should have been a separate PR, and not included in a "pdf fix".

bact:
Agree on the point of "join".

model/Core/Classes/PackageVerificationCode.md
kestewart: A hash of a null string gives unpredictable valued, based on implementation environement. 

I'll recode this though to be clearer with an "if" at the end.

kestewart, line 21:
```suggestion
```

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
 
 Algorithm:
 
+    hashValue = 0
     templist = empty list of strings
```
bact, line 22:
```suggestion
    templist = ""
```

To keep the original implementation, this should be an empty string. Not a list.

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,21 @@ This identifier enables a recipient to determine if any file in the original pac
...
+    }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
      
-    verificationcode = algorithm(valueslist)
+    hashValue = algorithm(valueslist)
```
kestewart, line 35:
```suggestion
    if valuelist is empty
       hashValue = 0
    else
       hashValue = algorithm(valueslist)
```

bact:
Btw, does the newline escape character is "typically" `\n` (and not `/n`)?
But this doesn't make any difference to the algorithm. The separator can be anything anyway.

model/Core/Classes/PackageVerificationCode.md
kestewart: Style preference.     Using C style comments, so used C style bracketing to be explicit about multiple statements, esp if indenting/formating gets mucked up.

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,23 @@ This identifier enables a recipient to determine if any file in the original pac
...
+    }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
      
-    verificationcode = algorithm(valueslist)
+    if valuelist is empty
```
bact, line 34:
```suggestion
    if valueslist is empty
```

variable with trailing s is very error prone...

model/Core/Classes/PackageVerificationCode.md
```diff
@@ -18,19 +18,23 @@ This identifier enables a recipient to determine if any file in the original pac
...
         else
-            append algorithm(contents of file) to templist
+            append "algorithm(file)/n" to templist
+    }
 
     sort templist in ascending order by value
-
-    valueslist = join templist values     /* ordered sequence with no separators */
+    
+    valueslist = remove "/n"s from templist  /* ordered sequence with no separators */
```
bact, line 32:
```suggestion
    /* ordered sequence with no separators */
    valueslist = remove "/n"s from templist
```

kestewart, line 32:
```suggestion
    /* remove separators from ordered sequence */
    valueslist = remove "/n"s from templist
```

bact: I removed my suggestion and +1 to this one.

bact (approved):
LGTM

zvr:
As I said when I submitted my fix for this, the issues are:

* it has templist as a string (not a list), appends to it values plus /n -- actually appending "algorithm(file)/n" without explaining that this is not a string with the letters a, l, g, ... but the value of the function algorithm, and appending a newline to each one (and of course newlines are `\n` not `/n`)
* then it says "sort templist" without explaining how to sort a single string
* then it computes the final result by calculating algorithm(filelist) -- so algorithm(file) actually reads a filename named file, but algorithm(valueslist) operates on the string argument.

As Gary said, it's moot and noone should be using this. I'm just annoyed that we publish such sloppy pseudocode.

goneall (approved):
LGTM

kestewart:
DIscussion on Outreach.  Agreement to merge.

PR #843, opened 2024-08-12
Update Lite profile

zvr:
Updates Lite profile, copying all the mandatory restrictions from the Annex.

### DIFF
--- a/model/Lite/Lite.md
+++ b/model/Lite/Lite.md
@@ -4,26 +4,20 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-The SPDX Lite profile defines a subset of the SPDX specification, from the
-point of view of use cases in some industries. SPDX Lite aims at the balance
-between the SPDX standard and actual workflows in some industries.
+The SPDX Lite profile defines a simple view of SPDX data,
+from the point of view of use cases in some industries.
 
 ## Description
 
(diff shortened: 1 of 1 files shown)


zvr:
@noriokobota and this is the Lite profile in the model.  Same content, but only the mandatory stuff.

model/Lite/Lite.md
```diff
@@ -49,3 +47,19 @@ the following has to hold:
...
    `from` property and an `/SimpleLicensing/AnyLicenseInfo` as its `to`
    property.
+
+Finally, for a `/Core/Agent` to be conformant with this profile, the following has to hold:
+
+1. The mincount for `name` is 1
+
+
+
+<!--
```
NorioKobota, line 57: What does this commented out block mean?

zvr: I'm waiting to update the Core classes that they should have at least one element and then I can delete this.

If we don't update the classes, I'll add it here, because SPDX-Lite requires non-empty SpdxDocuments and SBOMs.

:-)

NorioKobota: Got it, Thanks.
This change is also looks great.

kestewart (approved):

bobmartin3000 (approved):
LGTM

kestewart:
Thanks for making this consistent.

goneall (approved):
LGTM

PR #844, opened 2024-08-12
Bom and SpdxDocument cannot be empty

### LINKED ISSUE #841: Change cardinality for element in SpdxDocument and Bom classes
From the SPDX tech call 30 July 2024, we reached consensus that the SpdxDocument and Bom classees should have a minimum of 1 element.

We should update the model to reflect this decision.

zvr:
This makes sure Core/Bom and Core/SpdxDocument are not empty.

Essentially adds minCount:1 for elements in these classes.

Closes #841 

**Incompatible with** #845 (choose one)


kestewart:
Being handled by #845 - so closing this one.

zvr:
Decided to go with #845

PR #847, opened 2024-08-12
Change property name imports to import

goneall:
All property names should be singular

### DIFF
--- a/model/Core/Properties/import.md
+++ b/model/Core/Properties/import.md
@@ -1,6 +1,6 @@
 SPDX-License-Identifier: Community-Spec-1.0
 
-# imports
+# import
 
 ## Summary
 
@@ -8,11 +8,11 @@ Provides an ExternalMap of Element identifiers.
 
 ## Description
 
-Imports provides an ExternalMap of Element identifiers that are used within a
+Import provides an ExternalMap of Element identifiers that are used within a
 document but defined external to that document.
 
 ## Metadata
 
-- name: imports
+- name: import
 - Nature: ObjectProperty
 - Range: ExternalMap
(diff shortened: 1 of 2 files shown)


bact:
@goneall do you also like to update another instance here as well please?
https://github.com/spdx/spdx-3-model/blob/647f94b4bf9cde38bf660c66b6864abd91e5dd1e/serialization/jsonld/examples/converted_from_spdx_2.json#L96

bact:
But for its current form of this PR, I can approve.

kestewart (approved):
Urk, another one.   Find this one from the tooling?

model/Core/Properties/import.md
```diff
@@ -1,18 +1,18 @@
...
+# import
 
 ## Summary
 
 Provides an ExternalMap of Element identifiers.
 
 ## Description
 
-Imports provides an ExternalMap of Element identifiers that are used within a
+Import provides an ExternalMap of Element identifiers that are used within a
```
zvr, line 11:
```suggestion
Import provides an ExternalMap of an Element identifier that is used within a
```

bact (approved):
Approved for the code in spdx-spec.
Need to update a related example in spdx-3-model afterward.

zvr (changes requested):
I agree, but then the description of "import" should refer to only one entry.

bobmartin3000 (approved):
LGTM

goneall:
> Urk, another one. Find this one from the tooling?

Yes - I've implemented the code to translate from 2.X to 3.X - this was found from the unit testing.  Should be done with the unit tests by the end of the week.

bact:
@goneall I have add my suggestions to #848

kestewart (approved):
Re-reviewed.   Changes look good.  Going ahead with merging.

PR #849, opened 2024-08-13
Update URLs in serialization docs to 3.0.1

bact:
Note that these URLs are not available yet.

<s>We can wait until they are publicly accessible, by the setup of https://github.com/spdx/spdx-spec/issues/1108 .</s>

Update 9 Sep 2024: All the relevant URLs are available now. 
- https://spdx.org/rdf/3.0.1/spdx-model.ttl
- https://spdx.org/rdf/3.0.1/spdx-context.jsonld
- https://spdx.org/schema/3.0.1/spdx-json-schema.json


goneall:
We should merge this around the same time we update the website with the referenced documents.

goneall (approved):
The schema and ontology links are now live, we can probably merge them since the remaining IRI changes are not required to link to the documentation.

PR #851, opened 2024-08-13
SPDXDocument -> SpdxDocument

bact:
Use actual spelling of the Class


goneall (approved):
Agree

zvr (approved):

bobmartin3000 (approved):
LGTM

kestewart (approved):
Missed the fact that SPDXDocument went to SpdxDocument at some point in the class.  Would have prefered the class to be SPDXDocument as well, but given this is water under the bridge, agree it's best to be consistent.

PR #853, opened 2024-08-13
Revert PR 845 - allow empty Bom and SpdxDocument element lists

goneall:
This is based on the discussion on the 13 August 2024 tech call.

Adding the restrictions in the patch release would be a breaking change for some of the known tooling and it was agreed that adding the restrictions was not necessary.

The decision on the tech call was to allow empty root elements as well as empty element lists for both Bom and SpdxDocuments.


zvr:
Oh! :-(

That's too bad.

Then the Lite profile should be updated (in both the model and the spec), since they still have this requirement.

goneall:
@zvr - I just added a commit to update the Lite profile in the model (this) repo.  I looked at the spec Annex for lite, and it looks OK as is (already has the minimums defined for root and element).

model/Lite/Lite.md
```diff
@@ -47,19 +47,15 @@ Additionally:
...
    `from` property and an `/SimpleLicensing/AnyLicenseInfo` as its `to`
    property.
+   
+For a `/Core/SpdxDocument` to be conformant with this profile, the following has to hold:
+1. The mincount for `/Core/SpdxDocument/element` is 1
+1. The mincount for `/Core/SpdxDocument/rootElement` is 1
+
+For a `/Software/Sbom` to be conformant with this profile, the following has to hold:
+1. The mincount for `/Software/Sbom/element` is 1
+1. The mincount for `/Software/Sbom/rootElement` is 1
```
zvr, line 57:
```suggestion
1. The mincount for `element` is 1
1. The mincount for `rootElement` is 1

For a `/Software/Sbom` to be conformant with this profile, the following has to hold:
1. The mincount for `element` is 1
1. The mincount for `rootElement` is 1
```

zvr (changes requested):
To be consistent with the restrictions above.

(yes, I know, the old stuff in the comment was in the old style)

zvr (approved):

bact (approved):
Looks good. Implemented what was agreed in the meeting.

PR #854, opened 2024-08-14
RelationshipType: hasInputs/hasOutputs -> hasInput/hasOutput

bact:
Made the RelationshipType singular:

- hasInputs -> hasInput
- hasOutputs -> hasOutput

All other "has" relationship types have a singular "target".

For example, hasAddedFile, hasDataFile, hasDependencyManifest, hasDynamicLink.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## 3.0.1 (under development - last update 2024-08-10)
+## 3.0.1 (under development - last update 2024-08-14)
 
 ### Changes since 3.0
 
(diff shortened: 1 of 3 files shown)


goneall:
Perhaps we should fix it now, since we're fixing a number of other "spelling" issues.

We don't have an explicit rule on the relationships types, like we do on property names - so we could just keep this as is.

@zvr - what do you think?

zvr (approved):

goneall (approved):
LGTM

bobmartin3000:
We can reproduce pdfs up until the September meeting.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: goneall ***@***.***>
Sent: Wednesday, August 14, 2024 11:49:15 AM
To: spdx/spdx-3-model ***@***.***>
Cc: Robert A Martin ***@***.***>; Mention ***@***.***>
Subject: [EXT] Re: [spdx/spdx-3-model] RelationshipType: hasInputs/hasOutputs -> hasInput/hasOutput (PR #854)

@bobmartin3000 Any concerns with adding this change after the ISO PDF was produced? If not concerns, I'll go ahead and merge — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you were mentioned. Message


@bobmartin3000<https://github.com/bobmartin3000> Any concerns with adding this change after the ISO PDF was produced?

If not concerns, I'll go ahead and merge

PR #867, opened 2024-08-28
Add documentation for Markdown headings used in spec generation

bact:
Add a documentation about headings and formatting used in Markdown files for the specification generation.

This will allow contributors to better familiarise with the format that is expected from the spec-parser.

If there is an update in the spec-parser, this document should be updated as well.

--

Current **Table of contents**:

- Overview
- Directory organisation
- File content structure and formatting
  - Model file example
- Syntax
  - Classes
  - Datatypes
  - Individuals
  - Properties
  - Vocabularies
- Writing style
- Localisation
  - Localisation example


bact:
@zvr @goneall does this document best to be inside this spdx-3-model repo or in the spec-parser repo?

I can open new PR in spec-parser if it is more preferred (and close this one here).

goneall (approved):
Good catch @bact

specmark.md
```diff
@@ -0,0 +1,353 @@
...
+Blah.
+
+## Metadata
+
+- name: NoneElement
+- type: Element
+
+## Property Values
+
+- name: "NONE"
```
zvr, line 278: Individuals may define an IRI.

I am not sure "name" is used for anything (or processed at all)

specmark.md
```diff
@@ -0,0 +1,353 @@
...
+  and follows by one blank line.
+- The content immediately after the license information must begin with an
+  H1 heading containing the element's name.
+- Each element type has a predefined set of [allowed H2 headings](#syntax) and
+  labeled lists that must be used to structure its content.
+- There must be a blank line before and after a heading.
+- There must be a blank line before the beginning of a list.
+- There must be a blank line after the end of a list.
+- Any block level elements nested in a list, including paragraphs, sub-lists,
+  blockquotes, code blocks, etc., must always be indented by at least 4
```
zvr, line 48: Why is that?

zvr (dismissed):
I am not sure if we want to spend time trying to define everything.

We definitely need to also address multi-lingual content (maybe not right now, but in the future).

specmark.md
bact: Because the same file will be processed by MkDocs, and Python-Markdown that MkDocs use requires that.

https://python-markdown.github.io/#differences

specmark.md
bact: Thank you. IRI added to Metadata.

--

For the "name" property, I took the example from [/Core/NoneElement](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Individuals/NoneElement.md).

Not sure about their original intent. 

These individuals have name as "NONE":
- [/Core/NoneElement](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Individuals/NoneElement.md)
- [/ExpandedLicensing/NoneLicense](https://github.com/spdx/spdx-3-model/blob/main/model/ExpandedLicensing/Individuals/NoneLicense.md)

These two have name as "NOASSERTION":
- [/Core/NoAssertionElement](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Individuals/NoAssertionElement.md)
- [/ExpandedLicensing/NoAssertionLicense](https://github.com/spdx/spdx-3-model/blob/main/model/ExpandedLicensing/Individuals/NoAssertionLicense.md)

- As a label for human use? -- in [Relationship description](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/Relationship.md), "NONE" and "NOASSERTION" are used as labels to refer to `/Core/NoneElement` and `/Core/NoAssertionElement` individuals.
- As a comparable string for machine use? -- in [copyrightText](https://github.com/spdx/spdx-3-model/blob/main/model/Software/Properties/copyrightText.md), it describes the convention of using "NONE" and "NOASSERTION" string values. It may be possible to compare copyrightText value with the names of ExpandedLicensing individuals?

specmark.md
bact: I have added a sentence clarify that in the doc. Thank you for flagging this up.

goneall:
@zvr - Are you OK with merging this?  From the tech call, we felt there was value in documenting this for those doing localization / internationalization for the model.

specmark.md
```diff
@@ -0,0 +1,361 @@
...
+
+# NoneElement
+
+## Summary
+
+An Individual Value example.
+
+## Description
+
+Blah.
```
kestewart, line 277: Probably should update this.  ;-)

kestewart (approved):
Not sure we should leave a "Blah" in the text, but rest looks fine.

specmark.md
bact: > Probably should update this.  ;-)

@kestewart thank you. Updated.

zvr:
I'll review it in detail later this week, but it would be nice to define what our (long-term?) goal will be, @kestewart @goneall .

**If** we are to produce another PDF version, then my suggestion would be to limit the markdown that can be used, since currently I need to manually change some stuff to produce the PDF. If, on the other hand, we only care about the website (HTML version), then it's perfectly fine to leave "allow everything", as long as it renders fine.

From a quick look, we could add info on:

- links (all have to have text, no bare URLs)
- standardized formats (we have for RFCs, I don't remember whether there is already anything else)
- other languages sections

bact:
Just added a section at the bottom for bare links and RFC links

bact:
All three additions suggested by @zvr is added:

- [x] links (all have to have text, no bare URLs)
- [x] standardized formats (we have for RFCs, I don't remember whether there is already anything else)
  - So far we only have one for IETF RFCs.
- [x] other languages sections
  - With an example by @NorioKobota taken from #898

zvr:
@bact a couple of suggestions:
- change the filename, since it now suggests something about the SPEC benchmark ;-).  Maybe `format.md` ?
- separate the other-language stuff to `translations.md` 
- ideally we would like to have another doc describing our git branching strategy (not necessarily in this PR)

Maybe move everything to a DOCS subdirectory instead of top-level?

bact:
Thank you @zvr. I have updated the document, separating it to two files as suggested.

Also move them to a new `docs/`, together with few other docs.

zvr (approved):

zvr:
We can merge this, and improve it later.

Still missing, for example: the file where translations will provide the headings in their own language. But I haven't fully worked it out yet, so no need to wait and block this.

PR #877, opened 2024-09-09
Standardised RFC links in model Markdown files

bact:
- https://github.com/spdx/spdx-spec/issues/1012 needs two PRs to fix:
  - To spdx-3-model part is this PR
  - The spdx-spec part is in https://github.com/spdx/spdx-spec/issues/1107


model/Core/Vocabularies/HashAlgorithm.md
```diff
@@ -18,24 +18,24 @@ practically infeasible to invert.
...
+- blake2b384: BLAKE2b algorithm with a digest size of 384, as defined in [RFC 7693](https://datatracker.ietf.org/doc/rfc7693) Section 4.
+- blake2b512: BLAKE2b algorithm with a digest size of 512, as defined in [RFC 7693](https://datatracker.ietf.org/doc/rfc7693) Section 4.
 - blake3: [BLAKE3](https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/blake3.pdf)
 - crystalsDilithium: [Dilithium](https://pq-crystals.org/dilithium/)
 - crystalsKyber: [Kyber](https://pq-crystals.org/kyber/)
 - falcon: [FALCON](https://falcon-sign.info/falcon.pdf)
-- md2: MD2 message-digest algorithm, as defined in [RFC 1319](https://www.rfc-editor.org/info/rfc1319/).
-- md4: MD4 message-digest algorithm, as defined in [RFC 1186](https://www.rfc-editor.org/info/rfc1186).
-- md5: MD5 message-digest algorithm, as defined in [RFC 1321](https://www.rfc-editor.org/info/rfc1321).
+- md2: MD2 message-digest algorithm, as defined in [RFC 1319](https://datatracker.ietf.org/doc/rfc1319/).
```
goneall, line 29: Minor - while we're editing, let's remove the trailing `/` for consistency (the link works either way)

bact: Neat. Thank you.

bact: I have checked, the canonical URL has the /.
Will make all the same.

bact: Changed all to https://datatracker.ietf.org/doc/rfcXXXX/ (with trailing / )

Notes:
- https://www.rfc-editor.org/info/rfc3986 is without trailing / (RFC's own Bibtex use this format and the URL will not work with the trailing / )
- on the opposite, https://datatracker.ietf.org/doc/rfc3986/ is its canonical URL (the version without will redirected to the version with)

kestewart (approved):

PR #878, opened 2024-09-10
Lite: Fix conformance lists rendering; mincount -> minCount

bact:
The last two lists in https://spdx.github.io/spdx-spec/v3.0.1/model/Lite/Lite/ (starts at "For a /Core/SpdxDocument"...) don't rendered correctly.

This PR adds a newline before the lists, to make it render correctly.

Also fixes property name mincount to minCount.

### DIFF
--- a/model/Lite/Lite.md
+++ b/model/Lite/Lite.md
@@ -9,7 +9,6 @@ from the point of view of use cases in some industries.
 
 ## Description
 
-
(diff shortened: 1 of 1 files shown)


goneall (approved):
LGTM

kestewart (approved):
This is to fix descriptions in the RDF rendering.   Reviewed in meeting and approved.

PR #880, opened 2024-09-11
Add /Core/Individuals/SpdxOrganization

bact:
- This PR should be reviewed alongside https://github.com/spdx/spec-parser/pull/154
- Add a named individual: `SpdxOrganization`
- This individual is meant to be used for CreationInfo inside named individuals
- <s>Note that currently "## Property Values" is not processed, another PR in spec-parser is needed for spdxId, etc to be included in SpdxOrganization - see: https://github.com/spdx/spec-parser/pull/155</s>

### DIFF
--- a/model/Core/Individuals/SpdxOrganization.md
+++ b/model/Core/Individuals/SpdxOrganization.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-An Individual Value for Organization representing the SPDX Project.
+An Organization representing the SPDX Project.
 
 ## Description
 
--- a/model/Core/Individuals/SpdxOrganization.md
+++ b/model/Core/Individuals/SpdxOrganization.md
@@ -8,7 +8,7 @@ An Organization representing the SPDX Project.
 
 ## Description
 
-SpdxOrganization is used by named individuals for their creationInfo.
+SpdxOrganization is an Organization representing the SPDX Project. It is by definition the creator of all individuals defined by the SPDX Project. These individuals include licences and exceptions defined in the SPDX License List, as well as individuals defined in the specification.
 
 ## Metadata
 
--- a/model/Core/Individuals/SpdxOrganization.md
+++ b/model/Core/Individuals/SpdxOrganization.md
@@ -8,7 +8,10 @@ An Organization representing the SPDX Project.
 
 ## Description
 
-SpdxOrganization is an Organization representing the SPDX Project. It is by definition the creator of all individuals defined by the SPDX Project. These individuals include licences and exceptions defined in the SPDX License List, as well as individuals defined in the specification.
+SpdxOrganization is an Organization representing the SPDX Project.
+It is by definition the creator of all individuals defined by the SPDX Project.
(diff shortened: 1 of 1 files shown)


model/Core/Individuals/SpdxOrganization.md
```diff
@@ -0,0 +1,21 @@
...
+SpdxOrganization is used by named individuals for their creationInfo.
+
+## Metadata
+
+- name: SpdxOrganization
+- type: Organization
+
+## Property Values
+
+- spdxId: https://spdx.org/
```
zvr, line 20: No, there is no `spdxId` property.

Use `IRI` in Metadata as in https://github.com/spdx/spdx-3-model/blob/main/model/ExpandedLicensing/Individuals/NoneLicense.md

model/Core/Individuals/SpdxOrganization.md
```diff
@@ -0,0 +1,21 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# SpdxOrganization
+
+## Summary
+
+An Individual Value for Organization representing the SPDX Project.
```
zvr, line 7:
```suggestion
An Organization representing the SPDX Project.
```

model/Core/Individuals/SpdxOrganization.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Description
+
+SpdxOrganization is used by named individuals for their creationInfo.
```
zvr, line 11:
```suggestion
SpdxOrganization is an Organization representing the SPDX Project. It is by definition the creator of all individuals defined by the SPDX Project. These individuals include licences and exceptions defined in the SPDX License List, as well as individuals defined in the specification.
```

zvr (changes requested):

goneall:
I wonder if we should make the IRI more specific - for example, `https://spdx.org/organization`

I could create a web page with a redirect to https://spdx.dev/about/overview/

zvr:
I am confused... Don't you think that "https://spdx.org" is the right identifier for the SPDX project?

I mean, for another example, Intel has tons of pages describing the company (equivalent to "About us"), but the identifier to be used in such a context will definitely be https://www.intel.com/.

It will also be useful to others who would want to refer to the SPDX project to be able to simply use `<https://spdx.org>` as an Element identifier.

goneall (approved):
LGTM

goneall:
> I am confused... Don't you think that "https://spdx.org" is the right identifier for the SPDX project?
> 
> I mean, for another example, Intel has tons of pages describing the company (equivalent to "About us"), but the identifier to be used in such a context will definitely be https://www.intel.com/.
> 
> It will also be useful to others who would want to refer to the SPDX project to be able to simply use `<https://spdx.org>` as an Element identifier.

My suggest is based on "https://spdx.org" being the prefix for the SPDX RDF ontology which may be somewhat confusing.

That said, I'm OK going with the proposed URI.

I've approved the PR - so feel free to merge if you'd like to go with the current URI.

bact:
`SpdxOrganization` spec is now available at
https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Individuals/SpdxOrganization/

PR #883, opened 2024-09-13
Add missing custom IRIs for NoAssertionElement and NoneElement (for owl:sameAs)

bact:
Find this during diagram update for SpdxOrganization,
do the custom IRI of NoAssertionElement and NoneElement missing in the spec's model description?

Currently these are IRIs of NoAssertionElement, NoneElement, NoAssertionLicense and NoneLicense in the [digrams](https://spdx.github.io/spdx-spec/v3.0.1/annexes/rdf-model/):

Core
Licensing

These are "custom" or "simplified" IRIs for those individuals.

However, there are differences between the spec and the diagram:

- [NoAssertionElement](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Individuals/NoAssertionElement/)
  - Current IRI in spec: https://spdx.org/rdf/3.0.1/terms/Core/NoAssertionElement
  - Current IRI in diagram: https://spdx.org/rdf/3.0.1/terms/Core/NoAssertion
- [NoneElement](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Individuals/NoneElement/)
  - Current IRI in spec: https://spdx.org/rdf/3.0.1/terms/Core/NoneElement
  - Current IRI in diagram:https://spdx.org/rdf/3.0.1/terms/Core/None
- [NoAssertionLicense](https://spdx.github.io/spdx-spec/v3.0.1/model/ExpandedLicensing/Individuals/NoAssertionLicense/)
  - Current IRI in spec: https://spdx.org/rdf/3.0.1/terms/Licensing/NoAssertion
  - Current IRI in diagram: https://spdx.org/rdf/3.0.1/terms/Licensing/NoAssertion
- [NoneLicense](https://spdx.github.io/spdx-spec/v3.0.1/model/ExpandedLicensing/Individuals/NoneLicense/)
  - Current IRI in spec: https://spdx.org/rdf/3.0.1/terms/Licensing/None
  - Current IRI in diagram: https://spdx.org/rdf/3.0.1/terms/Licensing/None

- IRIs in spec and digram for License individuals are matched.
  - But IRIs of Element individuals are not matched.
- From the diagram and the Licensing spec, it is likely that custom IRIs in NoneElement and NoAssertionElement are missing.
  - This PR adds those two IRIs: Core/None and Core/NoAssertion
- Note that: adding these IRIs will not break the model, neither v3.0.1 nor v3.0.0.
  - Core/NoneElement and Core/NoAssertionElement are still valid and accessible


zvr:
No, there is no need for custom IRIs for these individuals.

Every class, property, individual gets an automatic IRI. These are perfectly fine for almost all cases.
In this case it's `https://spdx.org/rdf/3.0.1/terms/Core/NoneElement` and it's fine.

We've added custom IRIs to individual license elements to have a something simpler than mentioning ExpandedLicensing. These two are the exception, not the rule.
So, besides the automatically generated `https://spdx.org/rdf/3.0.1/terms/ExpandedLicensing/NoneLicense` there is also `https://spdx.org/rdf/3.0.1/terms/Licensing/None` and they are defined to be the same.

bact:
Thanks.

Then, do we need to update the diagram.

goneall:
Based on the tech call on 24 Sept 2024: We will move this to 3.1 - it has no breaking change and the inconsistency between the model and the diagram is resolved with removing individuals from the diagram

zvr:
I still maintain that this is not needed and should be closed.

PR #897, opened 2024-10-18
First draft of classes and properties as base for discussion WIP!

mkurzman:
as preparation for the peer review and discussion in the Operation Profile round.
Open points: 
- use of relations instead of properties?
- abstraction of properties to classes? (e.g. Delivery-class merely added as empty envelope)
- redundancy of properties that are already defined in other profiles?
- correct definition of types etc.


bact:
There's an error message from the spec-parser:

> ERROR:root:Missing top-level namespace file Operations.md

mkurzman:
Is that spec-parser available to use locally in my repo, so that I could do a pre-commit or pre-PR check?

bact:
> Is that spec-parser available to use locally in my repo, so that I could do a pre-commit or pre-PR check?

It is available at https://github.com/spdx/spec-parser

For that particular error, it means that you need a file named Operations.md in model/Operations/ folder.

See for example model/Core/Core.md

bact:
Btw, we're drafting a document about the formatting and convention for model Markdown.

If you find anything that you think if you know about them before it will save your time writing the model file, please share them in 
https://github.com/spdx/spdx-3-model/pull/867
Thank you

mkurzman:
Thank you for your hints. We will care about it. 
Hint: My PR was not meant to be merged in master yet, but merged in the profile-operations branch first (as a kind of "dev-branch", where we collaboratively prepare our PR for main. Is this the right way to go or will this approach mess things up? If yes, please advise for an alternative approach.

zvr:
This is the correct way of doing it, Marcel -- thanks!

bact:
@mkurzman Note that if you like to see if the PR actually passed the validation, at this moment you have to click to look inside the workflow details (spec-parser currently returns 0 even if there is an error, so GitHub will think that there is no error).

PR #898, opened 2024-10-21
Add Japanese Translation for spdx-3-model repo.

NorioKobota:
The Entries section of Vocabularies is next to the Metadata section(not translated), so could you check if it is okay to add the Japanese translation like this?
https://github.com/NorioKobota/spdx-3-model/blob/translation-ja-alpha/model/Core/Vocabularies/AnnotationType.md#entries-jp

And this PR is related to https://github.com/spdx/spdx-spec/pull/1138 .


bact:
Should the language code be `ja` instead of `jp`?

model/Core/Classes/Agent.md
```diff
@@ -19,3 +19,15 @@ This is not to be confused with tools that are used to perform tasks.
 - name: Agent
 - SubclassOf: Element
 - Instantiability: Concrete
+
+## Summary @ja
```
goneall, line 23: I noticed this has `@ja` - the next file has `@jp`

goneall (changes requested):
Noticed one inconsistency on the language code

model/Core/Classes/Agent.md
bact: Fixed by the most recent commit

zvr:
Converting this to draft, since it should be merged after 3.0.1 tagging and new branch has been created.

bact:
@NorioKobota does it possible to rebase this against `support/3.0` branch instead? thank you.

NorioKobota:
I resent [the PR](https://github.com/spdx/spdx-3-model/pull/953), so I will close this.

PR #905, opened 2024-11-12
Fixes typos found by OMG

zvr:
The OMG Editor went through all the specification and marked these two minor changes.


bact (approved):

bact:
Since the changes are reviewed by OMG and @zvr also made a pass on that, I think we can just merge so we can continue with PDF generation & etc.

Informing @bobmartin3000 @goneall @kestewart

PR #889, opened 2024-09-26
Final change log update for v3.0.1 model

bact:
This PR will be merge when we are happy to freeze the 3.0.1 and make tag release of the 3.0.1 model.

- [x] Remove "Unrelease" note from 3.0.1 heading
- [x] Grouped same type of changes together
- [ ] Update the date in the change log to the date of latest changes before freezing 3.0.1.
- To be logged
  - [x] #908
  - [x] #913

See change log update for the spec at https://github.com/spdx/spdx-spec/pull/1123

PR #908, opened 2024-11-12
VEX affected cardinalities

ilans:
- /Security/actionStatement:
The description in actionStatement states that:
"When an element is referenced with a VexAffectedVulnAssessmentRelationship, the relationship MUST include one actionStatement that SHOULD describe actions to remediate or mitigate the vulnerability".

This conforms to VEX_Use_Cases_Aprill2022:
"If a status is AFFECTED, the VEX document must have an action statement that tells the product user what to do".

The cardinality of actionStatement in VexAffectedVulnAssessmentRelationship is 0..1 but should to be 1..1.

- /Security/actionStatementTime:
It is optional and records the time when the actionStatement was first communicated (this conforms to the VEX docs).

The cardinality of actionStatementTime in VexAffectedVulnAssessmentRelationship is 0..* but should to be 0..1.

### DIFF
--- a/model/Security/Classes/VexAffectedVulnAssessmentRelationship.md
+++ b/model/Security/Classes/VexAffectedVulnAssessmentRelationship.md
@@ -47,8 +47,9 @@ following requirements must be observed:
 
 - actionStatement
   - type: xsd:string
-  - minCount: 0
+  - minCount: 1
   - maxCount: 1
 - actionStatementTime
   - type: /Core/DateTime
   - minCount: 0
+  - maxCount: 1


goneall:
@rnjudge - If you could just give this a quick review, I can merge

bact:
Note: This is a change that worth to be logged in the change log. 

Tag #889 so we won't forget.

rnjudge:
@goneall @ilans I agree with the min count changing to 1 in accordance with VEX standard. Is there any reason there can't be more than one, though? I suppose the VEX language says "an action statement" implying one but does not use strict language like "should" or "must". All of this to say... +1 to merge but want to make sure we are future proofing the cardinality as well.

bact:
Agree that from the language "must have an action statement", `actionStatement` cardinality could be 1..*

If we have two statements, we certainly have a statement (and fulfill the requirement).


Unfortunately, I think with the current design, if we have more than one `actionStatement`, it may not be possible to specify which `actionStatement` pairs with which `actionStatementTime`.

A new `hasActionStatement` relationship type may be able to capture that, together with the existing `Relationship`'s `startTime` (Could be 3.1?)

goneall:
Based on the above comments - I'm thinking we just go with a cardinality `1..1` and merge this PR.

I'll go ahead and approve and await a second approver.

bact (approved):
Approved based on VEX and based on current design constraints

Issue #910, opened 2024-11-13
packageVerificationCodeExcludedFile filename prefix


ilans:
From /Core/packageVerificationCodeExcludedFile:
"**In general**, every filename is preceded with a ./, see [RFC 3986](https://datatracker.ietf.org/doc/rfc3986/) ..."

Why use "**In general**" if the summary clearly states that this is "The **relative** file name...".
Shouldn't it always preceded with `./`?

If that's correct, I'll include a SHACL rule.

bact:
Because RFC 3986 also has another kind of relative reference that begins with `//`.

> A relative reference that begins with two slash characters is termed a network-path reference; such references are rarely used.

See "4.2.  Relative Reference" in https://datatracker.ietf.org/doc/rfc3986/

zvr:
We're talking about filenames, not URIs. So, there should only be "relative-path" references (to follow the nomenclature of RFC3986).

I don't even understand why there is a reference to this RFC. I'd rewrite the whole sentence as simply:

> Every filename starts with `./`.

bact:
Will "absolute-path" with a single slash (`/`) prefix (also from 4.2.  Relative Reference) possible in this context?

zvr:
I don't think so, since we're talking about which files _inside a package_ should be excluded when computing the verification hash.  It makes no sense to use absolute paths for these .

But probably @goneall should be the one to decide, since he's more proficient in this SPDXv2 verification stuff.

bact:
Thanks. Would be great if it can be simplified.

For reference, the RFC reference is from https://spdx.github.io/spdx-spec/v2.3/package-information/#742-intent and also the description there does not have "in general".

ilans:
Created a PR: https://github.com/spdx/spdx-3-model/pull/913

PR #913, opened 2024-11-14
use relative path syntax

ilans:
Require filenames to use the relative path syntax.

To fix https://github.com/spdx/spdx-3-model/issues/910

### DIFF
--- a/model/Core/Properties/packageVerificationCodeExcludedFile.md
+++ b/model/Core/Properties/packageVerificationCodeExcludedFile.md
@@ -12,9 +12,7 @@ The relative file name of a file to be excluded from the
 A relative filename with the root of the package archive or directory
 referencing a file to be excluded from the `PackageVerificationCode`.
 
-In general, every filename is preceded with a `./`, see
-[RFC 3986 Uniform Resource Identifier (URI): Generic Syntax](https://datatracker.ietf.org/doc/rfc3986/)
-for syntax.
+Every filename is preceded with a `./`.
 
 ## Metadata
 


bact:
To remove the entire RFC is a big change, considering this one already passed the OMG review. At this stage, I understand that the change should be minimal and as necessary. 

And the syntax for filename in RFC is more than just the prefix.

model/Core/Properties/packageVerificationCodeExcludedFile.md
```diff
@@ -12,9 +12,7 @@ The relative file name of a file to be excluded from the
 A relative filename with the root of the package archive or directory
 referencing a file to be excluded from the `PackageVerificationCode`.
 
-In general, every filename is preceded with a `./`, see
-[RFC 3986 Uniform Resource Identifier (URI): Generic Syntax](https://datatracker.ietf.org/doc/rfc3986/)
```
bact, line 16: Keep RFC please. 

Or defer this change to the next MINOR version update, which is 3.1.

bact (changes requested):

zvr:
Sorry, @bact, I must be misunderstanding something.

How is any of the RFC relevant?  The RFC talks about URIs, this field accepts filenames.

goneall:
Agree with @zvr and @ilans - we should remove this reference.

@zvr @ramartin - What will this do to the ISO / OMG review if we merge this in?

kestewart (approved):
Not sure why URI syntax is relevant to file name syntax, not sure why it was included initially.

goneall (approved):
Agree with the change - pending feedback from @zvr and @bobmartin3000 on impact to OMG review process

bact:
I'm fine to have the change. My concern this is clearly not a change in a category of typo.

The URI can be a filename, so there is no harm keeping it.

But there can be a potential harm of prolonging the OMG review process.
If there's not the case, I don't see problem to have this change go to 3.0.1.

We defer a lot of issues and PRs in the past few months because of that OMG review reason.
I'm just try to keep it consistent that way. This is my understanding.

bact:
This is Bob username @bobmartin3000

bact (approved):
If the process allows this to be in 3.0.1, please proceed.

bobmartin3000:
Functionally we aren't changing anything with the removal of the reference to the RFC. I'm okay with the change but it is stretching the editorial change label.

bobmartin3000 (approved):
LGTM.

PR #921, opened 2024-11-14
Use one-level unordered list to avoid formatting issue with MkDocs

### LINKED ISSUE #916: Indentation of bullets in Licensing profile is off
I checked on both Chrome and Safari and the bullet points are aligned to left:
> image

We use 2 spaces for indentation:
> image
Which should be fine. But some markdown formatters require 4 spaces.

I don't know if 4 spaces will solve it, but it's worth a try...

bact:
Option 2/3 to fix #916

Options:

- #920
- #921
- #922

**Only one of the three should be used**


goneall:
I'll go ahead and close this PR as the previous PR was accepted

PR #922, opened 2024-11-14
Collapse a nested list to a single-level list to avoid rendering issue in MkDocs

### LINKED ISSUE #916: Indentation of bullets in Licensing profile is off

bact:
Option 3/3 to fix #916

Options:

- #920
- #921
- #922

**Only one of the three should be used**


kestewart (approved):
Formatting change to improve readability.

goneall (approved):
LGTM

PR #928, opened 2024-11-21
Fixing typos about None and NoAssertion

### LINKED ISSUE #927: Relationship 'NONE' and 'NOASSERTION' individuals
From [Relationship](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Relationship/):

1. > To explicitly assert that no such relationships exist, the to property should contain the 'NONE' individual and no other elements.
A relationship that contains 'NONE' and additional elements in the to property is not valid.

2. > To explicitly assert that no assertions are being made regarding the existence of such relationships, the to property should contain the 'NOASSERTION' individual.

Questions:

- Why use 'NONE' and 'NOASSERTION' instead of the actual IDs of those individuals.
  - 'NONE' should be `NoneElement`
  - 'NOASSERTION' should be `NoAssertionElement`
- While constraint 1 requires 'NONE' to be the only `to` property, constraint 2 doesn't require it for 'NOASSERTION'. Was this intentional? Is it ok to have multiple `to` properties when using `NoAssertionElement` (in the same Relationship)?

bact:
- In [Core/Relationship](https://github.com/spdx/spdx-3-model/blob/main/model/Core/Vocabularies/RelationshipType.md), replace 'NONE' and 'NOASSERTION' with NoneElement and NoAssertionElement - to fix #927
- In [AI/autonomyType](https://spdx.github.io/spdx-spec/v3.0.1/model/AI/Properties/autonomyType/#autonomytype), <s>sort the list of "no", "noAssertion", and "yes". To make to order the same with the similar list at [Core/PresenceType](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/PresenceType/).</s>
  - <s>Unlike the list in Core/PresenceType, the list in AI/autonomyType cannot be automatically sorted by the spec-parser as it is not under "Entries" heading.</s>
  - <s>Also</s> add missing ending fullstops.

### DIFF
--- a/model/AI/Properties/autonomyType.md
+++ b/model/AI/Properties/autonomyType.md
@@ -12,10 +12,10 @@ involvement or guidance.
 Indicates if the system is fully automated or a human is involved in any of the
 decisions of the AI system.
 
+- yes: Indicates that the system is fully automated.
 - no: Indicates that a human is involved in any of the decisions of the AI
   system.
 - noAssertion: Makes no assertion about the autonomy.
-- yes: Indicates that the system is fully automated.
 
 ## Metadata
 
--- a/model/AI/Properties/autonomyType.md
+++ b/model/AI/Properties/autonomyType.md
@@ -12,10 +12,10 @@ involvement or guidance.
 Indicates if the system is fully automated or a human is involved in any of the
 decisions of the AI system.
 
-- yes: Indicates that the system is fully automated
 - no: Indicates that a human is involved in any of the decisions of the AI
-  system
-- noAssertion: Makes no assertion about the autonomy
+  system.
+- noAssertion: Makes no assertion about the autonomy.
+- yes: Indicates that the system is fully automated.
 
 ## Metadata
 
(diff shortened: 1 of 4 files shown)


model/AI/Properties/autonomyType.md
```diff
@@ -12,10 +12,10 @@ involvement or guidance.
 Indicates if the system is fully automated or a human is involved in any of the
 decisions of the AI system.
 
-- yes: Indicates that the system is fully automated
 - no: Indicates that a human is involved in any of the decisions of the AI
-  system
-- noAssertion: Makes no assertion about the autonomy
+  system.
+- noAssertion: Makes no assertion about the autonomy.
+- yes: Indicates that the system is fully automated.
```
zvr, line 19: I disagree with this change.
I think the order "yes", "no", "noAssertion" is much easier to understand.

zvr (changes requested):

model/AI/Properties/autonomyType.md
bact: I like to keep that order (yes, no, noAssertion) as well. I will revert that.

bact:
@zvr fixed. thank you.

goneall (approved):
LGTM - Thanks @bact

zvr (approved):

Issue #911, opened 2024-11-13
assessedElement to target SoftwareArtifact instead of Element


ilans:
From /Security/assessedElement:
"Specifies subpackages, files or snippets ..."

If it's limited to only packages, files and snippets then why its target is Element and not SoftwareArtifact?

Package, File and Snippet are the only subclasses of SoftwareArtifact.

zvr:
I think the target should be `/Core/Artifact`.

Since in the future we might have security info for other, non-software artifacts.

kestewart:
Having it as Element is going to provide more flexibility for future profiles.   Ok to discuss further but at this point it's going to 4.0 related change.

kestewart:
Removing the milestone until this is discussed.

goneall:
Discussed on tech call 2024-12-03:

- We want to allow the future relationship to hardware - propose Artifact rather than SoftwareArtifact
- Going from more restrictive to less restrictive would not be a breaking change, so we could restrict to SoftwareArtifact today and add HardwareArtifact in 3.1 without it being a breaking change
- Requirement don't block the hardware profile
- Requirement that we don't block functional safety
- In general, don't block any other future profiles
- Propose: Go with more restrictive now, relax the restrictions with the new profiles - this won't be breaking changes, so we meet the requirements
- Specific change is to change range from "Element" to "SoftwareArtifact"

PR #933, opened 2024-12-04
SoftwareArtifact as range for assessedElement

ilans:
See: https://github.com/spdx/spdx-3-model/issues/911

### DIFF
--- a/model/Security/Classes/VulnAssessmentRelationship.md
+++ b/model/Security/Classes/VulnAssessmentRelationship.md
@@ -20,7 +20,7 @@ assessment relationships. It factors out the common properties shared by them.
 ## Properties
 
 - assessedElement
-  - type: /Core/Element
+  - type: /Software/SoftwareArtifact
   - minCount: 0
   - maxCount: 1
 - publishedTime
--- a/model/Security/Properties/assessedElement.md
+++ b/model/Security/Properties/assessedElement.md
@@ -16,4 +16,4 @@ to specify the precise location where a vulnerability was found.
 
 - name: assessedElement
 - Nature: ObjectProperty
-- Range: /Core/Element
+- Range: /Software/SoftwareArtifact


bact:
- [  ] To update CHANGE LOG
- [  ] To update diagrams

goneall (approved):
LGTM - Thanks @ilans

zvr:
I've just realized that this introduces a dependency between the Security profile and the Software one. I don't know whether we had it before. Are we OK with this ?
@goneall @ilans @kestewart

zvr:
So, I've just checked. The dependency wasn't there before.
Can we simply make it `/Core/Artifact` to spare us long(er) discussions and more possible changes?

model/Security/Classes/VulnAssessmentRelationship.md
```diff
@@ -20,7 +20,7 @@ assessment relationships. It factors out the common properties shared by them.
 ## Properties
 
 - assessedElement
-  - type: /Core/Element
+  - type: /Software/SoftwareArtifact
```
goneall, line 23:
```suggestion
  - type: /Core/Artifact
```

model/Security/Properties/assessedElement.md
```diff
@@ -16,4 +16,4 @@ to specify the precise location where a vulnerability was found.
 
 - name: assessedElement
 - Nature: ObjectProperty
-- Range: /Core/Element
+- Range: /Software/SoftwareArtifact
```
goneall, line 19:
```suggestion
- Range: /Core/Artifact
```

goneall (changes requested):
Per Alexios' suggestion, added suggestion changing to `/Core/Artifact`

model/Security/Classes/VulnAssessmentRelationship.md
goneall: From call on 5 Dec - agreed to leave as SoftwareArtifact

model/Security/Properties/assessedElement.md
goneall: From call on 5 Dec - agreed to leave as SoftwareArtifact

goneall (approved):

bact (approved):

PR #935, opened 2024-12-06
Updated diagrams

zvr:
I've updated and simplified all diagrams.

This PR adds:

- the editable (draw.io) diagram (each diagram in a tab)
- the exported diagrams (in PNG format)

Once merged, they should also be copied in the spec repo.


bact (commented):
Please use all lowercase for filenames, as they were and also referred to in spdx-spec Markdown.

This will also allow GitHub to show diffs between old and new images.

bact:
As we have xsd types for properties, I think we should also keep xsd types for Simple Data Types as well, also the constraints.

We should simplify the diagram by removing the enums which can be rather long.

But those xsd types and constraints of Simple Data Types are just short, and from what I see in the Core PNG, we have plenty of spaces around the Simple Data Types box.

We can cut them shorter to:

> **SemVer**
xsd:string constrained to SemVer 2.0.0

> **MediaType**
xsd:string constrained to RFC 2046

> **DateTime**
xsd:string constrained to ISO-8601

I think that should fit the box.

zvr:
> Please use all lowercase for filenames, as they were and also referred to in spdx-spec Markdown.

The profile names are always with capital first letter, the tabs in the .drawio file are the same, I simply used the automatically generated name -- otherwise why even have "model-" ? Why introduce a new convention with lowercase here?

bact:
Because with all uppercase or all lowercase it is less error prone when we type in the code, compared to mixed cases.

bact:
I can see that the prefix "model-" is for file organizational purposes.

- When looking at the filename, one can make a fair guess that it should be something about "license model" and maybe not "license logo".
- When stored with other files in the same directory, prefix helps with grouping.
  - Not a great use case here, because in our images/ we only have one group of images.
  - But I guess in the case of OMG submission, we can see that in (resulting) images/ it could have some historical diagrams (spdx23) mixed with the diagrams here.
  - So if we like to change the prefix from "model-" to "spdx30-", I'm happy too.

--

On casing,
in the past five or six months, I have encountered a number of MkDocs issues that are related to wrong casing and make the build fail or result in deadlink. So if there's anything that will make that less likely repeated, I will do.

bact:
In earlier versions of SPDX 3 spec website, these model images are not even displayed. Because we just use the generated filename which contains a space. And that breaks when it's inside the URL. Somehow the space doesn't get automatically converted to %20.

That's why we now use "-" instead of a space (thus "model-" and not "model ").

zvr:
As I said, I use the standard filenames that drawio uses (which is filename-tabname).

I don't have time for this; feel free to change them afterwards.
I'm producing the PDFs now.

bact:
Please go ahead with the PDF.

bact:
Currently there are two styles of referencing an element from a profile:

- With prefix `/` : `/Profile/element` (see "File" box in Software diagram, for example)

- Without prefix `/` : `Profile/element` (see "VulnAssessmentRelationship" box in Security diagram, for example)

Not sure which one we are more preferred.

This is minor and does not affect the meaning of the diagram. Can be fixed for consistency later.

bact:
Yes. They were before you made edits. No worries.

goneall:
the `VexAffectedVulnAssessmentRelationship` `actionStatement` was updated to a min cardinality of 1 in the model - this is not reflected in the diagrams.

goneall:
the `VexAffectedVulnAssessmentRelationship` `actionStatementTime` was updated to a max cardinality of 1 in the model - this is not reflected in the diagrams.

goneall (changes requested):
Found two updates to the model that were missed in the diagram for the security profile detailed in the above 2 comments.

goneall (approved):
LGTM

bact (approved):
Will handle the entries ordering and etc later. This is good for now.

Issue #936, opened 2024-12-06
Shape validation fails related to 5 named individuals


bact:
These violations are all new:
- https://github.com/spdx/spdx-spec/actions/runs/12186737928/job/33996641030?pr=1154 (pyshacl)
- https://github.com/spdx/spdx-spec/actions/runs/12193560287/job/34015982565?pr=1134 (pyshacl)
- https://github.com/spdx/spdx-spec/actions/runs/12194034263/job/34017227934?pr=1134 (spdx3-validate)

**pyshacl** produces 32 violations. For example:

BlankNode:

> Constraint Violation in NodeKindConstraintComponent (http://www.w3.org/ns/shacl#NodeKindConstraintComponent):
	Severity: sh:Violation
	Source Shape: ns1:CreationInfo
	Focus Node: <https://spdx.org/rdf/3.0.1/creationInfo_NoneLicense>
	Value Node: <https://spdx.org/rdf/3.0.1/creationInfo_NoneLicense>
	Message: Value is not of Node Kind sh:BlankNode

Agent:

> Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class ns1:Agent ; sh:minCount Literal("1", datatype=xsd:integer) ; sh:nodeKind sh:IRI ; sh:path ns1:createdBy ]
	Focus Node: <https://spdx.org/rdf/3.0.1/creationInfo_NoneLicense>
	Value Node: <https://spdx.org/>
	Result Path: ns1:createdBy
	Message: Value does not have class ns1:Agent

string:

> Constraint Violation in DatatypeConstraintComponent (http://www.w3.org/ns/shacl#DatatypeConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:datatype xsd:string ; sh:maxCount Literal("1", datatype=xsd:integer) ; sh:nodeKind sh:Literal ; sh:path ns1:comment ]
	Focus Node: <https://spdx.org/rdf/3.0.1/creationInfo_NoneLicense>
	Value Node: Literal("This individual element was defined by the spec.", lang=en)
	Result Path: ns1:comment
	Message: Value is not Literal with datatype xsd:string

Abstract class:

> Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
	Severity: sh:Violation
	Source Shape: ns1:Element
[...]

PR #937, opened 2024-12-09
Create an IndividualElement class

goneall:
This will allow individuals to be of a type that can be instantiated and could resolve the issue with SHACL validation failing https://github.com/spdx/spec-parser/issues/165

### DIFF
--- a/model/Core/Classes/IndividualElement.md
+++ b/model/Core/Classes/IndividualElement.md
@@ -1,0 +1,21 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# IndividualElement
+
+## Summary
+
+A concrete subclass of Element used by Individuals in the
+Core profile.
+
+## Description
+
+Individuals, such as NoneElement and NoAssertionElement, need to reference a
+concrete subclass of ELement.
+
+This class provides the type used by the individuals.
+
+## Metadata
+
+- name: IndividualElement
+- SubclassOf: Element
+- Instantiability: Concrete
--- a/model/Core/Classes/IndividualElement.md
+++ b/model/Core/Classes/IndividualElement.md
@@ -12,7 +12,7 @@ Core profile.
 Individuals, such as NoneElement and NoAssertionElement, need to reference a
 concrete subclass of Element.
 
-This class provides the type used by the individuals.
+This class provides the type used by the individuals defined in the Core profile.
 
 ## Metadata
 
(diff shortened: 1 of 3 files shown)


goneall:
See this conversation in https://github.com/spdx/spdx-3-model/issues/936 for context

JPEWdev:
This might make it kinda hard to know how named individuals are supposed to be used; I think maybe we can deal with it if necessary but I need to do some testing before we go ahead and make this change

JPEWdev:
Ah, you know what, I was think of enums. This should be fine

model/Core/Classes/IndividualElement.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Summary
+
+A concrete subclass of Element used by Individuals in the
+Core profile.
+
+## Description
+
+Individuals, such as NoneElement and NoAssertionElement, need to reference a
+concrete subclass of ELement.
```
bact, line 13:
```suggestion
concrete subclass of Element.
```

bact (changes requested):

model/Core/Classes/IndividualElement.md
```diff
@@ -0,0 +1,21 @@
...
+
+This class provides the type used by the individuals.
```
bact, line 15:
```suggestion
This class provides the type used by the individuals.

For individuals that are subclass of AnyLicenseInfo, please use IndividualLicensingInfo.
```

As AnyLicenseInfo is also a subclass of Element, do we have to point that out here as well?

bact (changes requested):

model/Core/Classes/IndividualElement.md
```diff
@@ -0,0 +1,21 @@
...
+
+A concrete subclass of Element used by Individuals in the
+Core profile.
+
+## Description
+
+Individuals, such as NoneElement and NoAssertionElement, need to reference a
+concrete subclass of Element.
+
+This class provides the type used by the individuals.
```
zvr, line 15:
```suggestion
This class provides the type used by the individuals defined in the Core profile.
```

zvr (changes requested):

model/Core/Classes/IndividualElement.md
zvr: We shouldn't be talking about individuals defined in other profiles here.

model/Core/Classes/IndividualElement.md
bact: +1 to this for scoping.

bact (approved):

bact:
Approved, providing the suggested "for Core profile" by @zvr is added to the Description.

goneall:
@zvr - Accepted suggestion - if you could approve and merge everything should be working.

zvr (approved):

PR #963, opened 2025-01-23
Print verbose info during PR validation (target: 'develop' branch)

bact:
- Show the version of the spec-parser currently working with
- Add <s>`--debug`</s> `--verbose` to spec-parser options to print more information for the PR author to understand why the purposed changes failed the validation (if it failed)


zvr:
What is the purpose of the "validation"?
I assume to make sure that the input is "correct" / "valid".

What does this mean? 
More specifically:
- errors should most definitely be flagged
- what about warnings?

And how is it checked?
- is only the exit code of the process?
- or is the generated output / error messages also checked?

bact:
If it's valid it's good.

If it's not, having some information to understand why it is the case is good.

goneall:
@bact - Can you rebase to trigger the CI?  The old CI failed, but it is too stale to see the logs

@zvr - any concerns about merging?

bact:
Thanks. I will rebase / check the parameters again to keep up with latest spec-parser.

goneall:
Looks like this is still failing the CI

@bact - can you check?

bact:
The CI is good now. @goneall

goneall (approved):

bact:
All green. Will merge with one approval and 10+ day pass (7 months).

PR #966, opened 2025-01-30
Software: Add artifactSize in field

JPEWdev:
Adds a property to software Artifact that indicates the size, in bytes

### DIFF
--- a/model/Software/Properties/artifactSize.md
+++ b/model/Software/Properties/artifactSize.md
@@ -1,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# artifactSize
+
+## Summary
+
+Identifies the size of a software Artifact, in bytes
+
+## Description
+
+Identifies the size of a software Artifact, in bytes
+
+## Metadata
+
+- name: artifactSize
+- Nature: DataProperty
+- Range: xsd:nonNegativeInteger
--- a/model/Software/Properties/artifactSize.md
+++ b/model/Software/Properties/artifactSize.md
@@ -8,7 +8,7 @@ Identifies the size of a software Artifact, in bytes.
 
 ## Description
 
-Identifies the size of a software Artifact, in bytes
+Identifies the size of a software Artifact, in bytes.
 
 ## Metadata
 
--- a/model/Software/Properties/artifactSize.md
+++ b/model/Software/Properties/artifactSize.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Identifies the size of a software Artifact, in bytes
+Identifies the size of a software Artifact, in bytes.
 
 ## Description
 
--- a/model/Software/Classes/SoftwareArtifact.md
+++ b/model/Software/Classes/SoftwareArtifact.md
@@ -36,3 +36,7 @@ such as a package, a file, or a snippet.
 - contentIdentifier
   - type: ContentIdentifier
   - minCount: 0
+- artifactSize
+  - type: xsd:nonNegativeInteger
+  - minCount: 0
+  - maxCount: 1


bact:
Two things:

1) Does it possible for a `SoftwareArtifact` to have a size of 0 ?
    - If yes, then we may need to have this as a `xsd:nonNegativeInteger` type to allows zero

2) If we eventually have this, we may like to deprecate [`Dataset/datasetSize`](https://spdx.github.io/spdx-spec/v3.0.1/model/Dataset/Properties/datasetSize/) and let [`Dataset/DatasetPackage`](https://spdx.github.io/spdx-spec/v3.0.1/model/Dataset/Classes/DatasetPackage/) use this `Core/sizeInBytes` instead.
    - Good news: `DatasetPackage` is a `SoftwareArtifact`, and the `datasetSize` is in bytes
    - @rgopikrishnan91 @bennetkl what do you think?

JPEWdev:
Ya, 0 bytes makes sense, so we should change that.

zvr:
Why not simply call it `size`, or at most `artifactSize` ? We can have the description say that the number is in bytes.

zvr (approved):

kestewart (approved):
LGTM

bennetkl (approved):
I'm in agreement to add this field in software profile and remove from datasets.  I like the new name of field, better than with bytes in the field name.   You might want to put in examples of how to fill in this field.  I approve the change and for 3.1 we add a PR to remove similar field from dataset profile.

bact (approved):

bact:
> I'm in agreement to add this field in software profile and remove from datasets.  I like the new name of field, better than with bytes in the field name.   You might want to put in examples of how to fill in this field.  I approve the change and for 3.1 we add a PR to remove similar field from dataset profile.
> 

@bennetkl for backward compatibility with 3.0.x, I don't think we can remove datasetSize from the spec.

We can however deprecate it, put some description encouraging people to use the new artifactSize, also maybe considering putting a `owl:sameAs` to tell that the two are similar.

A 3.0 to 3.1 migration tool can also help users convert this.

model/Software/Properties/artifactSize.md
```diff
@@ -0,0 +1,17 @@
...
+
+# artifactSize
+
+## Summary
+
+Identifies the size of a software Artifact, in bytes
+
+## Description
+
+Identifies the size of a software Artifact, in bytes
```
bact:
```suggestion
Identifies the size of a software Artifact, in bytes.
```

model/Software/Properties/artifactSize.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# artifactSize
+
+## Summary
+
+Identifies the size of a software Artifact, in bytes
```
bact:
```suggestion
Identifies the size of a software Artifact, in bytes.
```

bact (changes requested):

bact (approved):

zvr (approved):

bact:
Additional notes

1) DCAT uses `byteSize` https://www.w3.org/TR/vocab-dcat-3/#Property:distribution_size

2) Based on the EU "Explanatory Notice and Template for the Public Summary of Training Content for general-purpose AI models" we may need a new field to record a dataset size with different units (for example, in "tokens of words").
    - The reporting template https://digital-strategy.ec.europa.eu/en/library/explanatory-notice-and-template-public-summary-training-content-general-purpose-ai-models
    - Initial mapping with SPDX 3 and notes of current model limitation https://docs.google.com/spreadsheets/d/15M-y6ibBeg1NtYCyNCtrVdj9fi4aINhA-aUr-ql5Y-A/edit?gid=414617654#gid=414617654

PR #967, opened 2025-02-10
Add geographic information to services

goneall:
Addresses use case 10


goneall:
Reviewed on the Software as a Service call on 10 Feb 2025 - approved

PR #971, opened 2025-02-13
Add spdx3-validate and online validator to JSON-LD validation guide

### LINKED ISSUE #929: Add spdx3-validate to JSON-LD validation doc
Add the use of [spdx3-validate](https://github.com/JPEWdev/spdx3-validate), which can handle `spdxId` defined in an `ExternalMap`, to https://github.com/spdx/spdx-3-model/blob/main/serialization/jsonld/validation.md

bact:
- Add instructions for spdx3-validate and https://tools.spdx.org/app/validate/
  - To resolves #929 
- Restructure and add table of contents


goneall (approved):
LGTM - Thanks @bact

serialization/jsonld/validation.md
```diff
@@ -1,28 +1,127 @@
...
+For example,
+
+- `dataset_datasetType` for a `datasetType` Property in the `Dataset` Profile
+- `expandedlicensing_CustomLicense` for a `CustomLicense` Class in the
+  `ExpandedLicensing` Profile
+- `Person` for a Class `Person` in the `Core` Profile (no prefix)
+
+### Cardinality
+
+A property with a cardinality greater than 1 must be represented as an array in
```
ilans, line 82: This is not required by json-ld and is not supported by json-ld libraries. Meaning that (most, if not all) json-ld serializers would omit the brackets in case of a single edge. So this may break for anyone using a non-custom-made json-ld serializer. Do we really require this?

serialization/jsonld/validation.md
```diff
@@ -1,28 +1,127 @@
...
+JSON, regardless of the actual number of values it holds.
+
+### Casing
+
+Note that SPDX 3 may use different casing than other vocabularies.
```
ilans:
```suggestion
Note that SPDX 3 may use different casing than vocabularies in previous versions.
```

ilans (changes requested):

serialization/jsonld/validation.md
bact: This has to be enforced, or it will not pass the Schema validation in the CI.

You can try take this [Software Profile example1.json](https://github.com/spdx/spdx-examples/blob/master/software/example1/spdx3.0/example1.json), remove the brackets from one of the `to`s, and run:

```shell
check-jsonschema \
        --verbose \
        --schemafile https://spdx.org/schema/3.0.1/spdx-json-schema.json \
        "example1.json"
```

(The command is taken from [bin/check-examples.sh](https://github.com/spdx/spdx-spec/blob/develop/bin/check-examples.sh) which will be run by [a validation workflow](https://github.com/spdx/spdx-spec/blob/develop/.github/workflows/validate_examples.yml) in spdx-spec repo)

ilans: Yes, but that constraint is imposed by the json schema, it's not required by the json-ld spec. To quote:
> Values of terms associated with an @set or @list container will always be represented in the form of an [array](https://infra.spec.whatwg.org/#list) when a document is processed—even if there is just a single value that would otherwise be optimized to a non-array form.

The current rule in the schema is:
```json
"allOf": [
    {
            "to": {
                "oneOf": [
                    {
                        "type": "array",
                        "minItems": 1,
                        "items": {
                            "$ref": "#/$defs/prop_Relationship_to"
                        }
                    }
                ]
            }
        },
    }
]
```

If we change it to:
```json
"allOf": [
    {
            "to": {
                "oneOf": [
                    {
                        "type": "array",
                        "minItems": 1,
                        "items": {
                            "$ref": "#/$defs/prop_Relationship_to"
                        }
                    },
                    {
                        "$ref": "#/$defs/prop_Relationship_to"
                    }
                ]
            }
        },
    }
]
```
It'll work!

The rule is already nested inside a "oneOf". I don't know why it wasn't utilized.

@JPEWdev ?

ilans: @bact 
After a long discussion with @JPEWdev and @goneall on the implementers call, I understand that this constraint is required. That's because the canonical serialization section in the original "Model and serializations" document isn't meant to describe a JSON-LD format but a proprietary JSON that was defined by the SPDX 3 community as a compromise between pure JSON and JSON-LD. On one hand, it allows to deserialize a canonical SPDX 3 JSON into RDF using standard JSON-LD libraries. On the other, it may only be serialized using proprietary tools which follow the specifications listed in this document.

The confusion arises from using the term JSON-LD in the context of serialization. JSON-LD should only be mentioned in relation to deserialization of "SPDX 3 JSON" and we should avoid using the term "SPDX 3 JSON-LD".

I opened an issue to discuss how to make this more concise and clearer: https://github.com/spdx/spdx-3-model/issues/1008

bact: Thank you. Until the schema has changed to what you have proposed, the documentation has to be kept as it is, so it will not break. (Please be noted that the text itself is already there, this PR just moves it up from the bottom)

After the schema gets updated, we can update this section of the doc in another PR.

bact:
Since this one has been around for a month and suggested changes were already incorporated. I will merge this documentation change with one approval - to prepare the file/avoid merge conflicts with incoming of name changes resulting from #1008. Please follow that issue for more discussion about the JSON vs JSON-LD name.

bact:
Note that in case we agreed to no longer call this a JSON-LD, apart from editing this file, the file itself might also need to be moved from "serialization/jsonld/" directory to somewhere else.

PR #972, opened 2025-02-13
Add spec-parser usage to Markdown formatting document

bact:
Let the model editors aware of the tool


goneall (approved):
LGTM

docs/format.md
```diff
@@ -527,3 +528,22 @@ recommendations when writing paragraph text and incorporating links.
...
+
+## Checking if everything is ok
+
+Use [spec-parser][] to check if your additions or edits are correctly
+formatted.
+
+Install:
+
+```shell
+git clone https://github.com/spdx/spec-parser.git
```
zvr: If the intent is to only run it, cloning the main branch and without any history would be faster.

zvr (changes requested):

docs/format.md
bact: Thank you. Added the branch and depth options.

bact:
Requested changes (git cloning the main branch without any history) made.
Will merge with one approval.

PR #973, opened 2025-02-13
Add Software as a Service Profile

goneall:
The Service profile is not available for review.

We will discuss this PR on the 4 March tech call.

One specific area to review is the geographic information.  Is this property something which should be in core and shared with other profiles?


model/Service/Service.md
```diff
@@ -0,0 +1,16 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Service
+
+## Summary
+
+The  Profile captures software as a service related information.
```
bact:
```suggestion
The Profile captures software as a service related information.
```

model/Service/Vocabularies/AuthenticationProtocolType.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Metadata
+
+- name: AuthenticationProtocolType
+
+## Entries
+
+- tls: Transport Layer Security, or TLS, is a widely adopted security protocol designed to facilitate privacy and data security for communications over the Internet.
+- ocsp: OCSP (Online Certificate Status Protocol) is a common schemes used to maintain the security of a server and other network resources.
+- crl: Certificate Revocation List (CRL) - A CRL is a list of revoked certificates that is downloaded from the Certificate Authority (CA).
```
bact, line 21:
```suggestion
- ocsp: Online Certificate Status Protocol, or OCSP, is a common scheme used to maintain the security of a server and other network resources.
- crl: Certificate Revocation List, or CRL, is a list of revoked certificates that is downloaded from the Certificate Authority (CA).
```

Make the writing style at the beginning of the sentence similar to TLS.

bact (changes requested):
Minor / stylistic / typo change suggestions.

bact:
Per decision documented in #226, we may need to rename `serviceHostingCountries` to `serviceHostingCountry`.

model/Service/Properties/serviceHostingCountries.md
```diff
@@ -0,0 +1,20 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# serviceHostingCountries
```
bact:
```suggestion
# serviceHostingCountry
```

model/Service/Properties/serviceHostingCountries.md
```diff
@@ -0,0 +1,20 @@
...
+## Description
+
+This property should specify all countries and territories where a service is located.
+This includes indirect locations (e.g. where the data may be hosted).
+The string must be in the [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) three-letter format.  
+See the [ISO 3166-1 alpha-3 Wikipedia page](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) for more information.
+
+## Metadata
+
+- name: serviceHostingCountries
```
bact:
```suggestion
- name: serviceHostingCountry
```

model/Service/Properties/serviceHostingCountries.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Summary
+
+Specifies a country code where a software service is hosted.
+
+## Description
+
+This property should specify all countries and territories where a service is located.
+This includes indirect locations (e.g. where the data may be hosted).
+The string must be in the [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) three-letter format.
```
bact:
```suggestion

The string must be in the [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) three-letter format.
```

model/Service/Classes/SoftwareService.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Properties
+
+- provider
+  - type: ServiceProvider
+  - minCount: 1
+- serverAuthenticationProtocol
+  - type: AuthenticationProtocolType
+  - minCount: 0
+- serviceHostingCountries
```
bact:
```suggestion
- serviceHostingCountry
```

bact (changes requested):
- Per decision documented in https://github.com/spdx/spdx-3-model/issues/226, we may need to rename :
  `serviceHostingCountries` -> `serviceHostingCountry`

model/Service/Vocabularies/AuthenticationProtocolType.md
bact:
```suggestion
- tls: Transport Layer Security, or TLS, is a widely adopted security protocol designed to facilitate privacy and data security for communications over the Internet.
- ocsp: Online Certificate Status Protocol, or OCSP, is a common scheme used to maintain the security of a server and other network resources.
- crl: Certificate Revocation List, or CRL, is a list of revoked certificates that is downloaded from the Certificate Authority (CA).
```

Make the writing style at the beginning of the descriptions consistent (based on `tls` entry description).

bact (changes requested):

bact (approved):
LGTM for the initial development commit. We can update the property name (serviceHostingCountries -> serviceHostingCountry) in the diagram later together with other feedbacks from the 4 March tech meeting.

bact:
Possible common properties among Hardware, Operations, and Software as a Service Profiles:

- Geographic location
  - Geographic coordinates
- Addresses
  - Street address
  - Mailing address
- Contact person
  - Contact details

Related standards:

- Schema.org GeoCoordinates
  https://schema.org/GeoCoordinates
- Schema.org location (Place, PostalAddres, VirtualLocation):
  https://schema.org/location
- Schema.org areaServed:
  https://schema.org/areaServed
- Schema.org ContactPoint:
  https://schema.org/ContactPoint
- Core Location Vocabulary
  https://semiceu.github.io/Core-Location-Vocabulary/releases/2.1.0/
- ISO 6709, Standard representation of geographic point location by coordinates
  https://www.iso.org/standard/75147.html
- Global Location Number https://www.gs1.org/standards/id-keys/gln
- Delivery Context Ontology
  - Location https://www.w3.org/TR/2009/WD-dcontology-20090616/location.html
  - Hardware https://www.w3.org/TR/2009/WD-dcontology-20090616/hardware.html

model/Service/Classes/ServiceProvider.md
stevenc-stb, line 1: The definition ServiceProvider does not appear to be an agent but instead is a role that an agent plays in a system.

This a role because it tells a function(software service provider) that is assumed or part played by an entity in a particular situation.
Agents often will have many roles…

Solution:
Add a role type to agents.
A) add a List of common roles as a property in agent.  This is in alignment with gs1 standard. gs1:organization:organizationRole 
https://www.gs1.org/sites/gs1/files/2022-06/image013.png
A ServiceProvider could be a roleType.
List of gs1roles: https://ref.gs1.org/voc/OrganizationRoleType

stevenc-stb (changes requested):

model/Service/Classes/ServiceProvider.md
goneall: In general, I agree with your observation - in a past life I've worked extensively in role based systems which can be very powerful (and very complex) - especially as it relates to authorization.

The use cases we've considered for the Software as a Service profile only requires that we identify the legal entity associated with the service - so in any case the Agent will be required in the model.

Having the service provide Agent as a property in the Service makes it very easy to answer the question "What legal entity is responsible for the service?"  Implementing a role based approach would make it much more complicated - especially since we can have the definition of the Agent in a separate serialization from the Service definition.

The other issue with having roles be properties of the Agent is that roles and the association of roles to Agents change.  Since Agents are immutable, we would need to somehow externalize the roles (e.g. role can be a subclass of relationship between an Element and Agent).

There may be other use cases where having the role definitions are necessary, but for now I would prefer to keep the Service profile relatively simple.

There is a good argument for making the service provider a relationship type rather than a property since any change in the service provide would necessarily obsolete the previous service definition and require the creation of a new service definition (since Elements are immutable).  We did discuss this long ago and decided to take the simpler property approach.

BTW - This same point about using roles could be made for many other uses of the Agent class.  If we want to go down the roles route, we may want to re-examine the overall Agent model.

stevenc-stb: The supply chain uses agents with the need to map multiple relationships to the same agent.
If we add an agent subclass that defines roles this will greatly complicate the process. A long list of agent subclasses will make the process more confusing as the number of subclasses grows.

Supply Chain Example:
A corporation manufactures a product and is a service provider. This requires the use of different agent types for the same company. The problem is assigning the correct agents to different use cases for the same company. Associations are complex because they need to show the same party.

A solution is to create an abstract (or Concrete) class of `Role` off of `Elements`. Make the `ServiceProvider` a subclass (or property) of that role class. Create a Relationship with the type “Role” between the `ServiceProvider` (or Role) and the `Agents` class.

model/Service/Classes/SoftwareService.md
```diff
@@ -0,0 +1,29 @@
...
+
+Software provided as a service over a network.
+
+## Description
+
+SoftwareService represents software being licensed, delivered and accessed online over a network.
+
+## Properties
+
+- provider
```
stevenc-stb, line 15: The provider in the SoftwareService.md is a type: ServiceProvider
but in the provider.md is  Range: /Core/Agent

model/Service/Classes/ServiceProvider.md
goneall: Thanks @stevenc-stb for the additional context and example.  I think I now understand the issue.  

What if we just get rid of the `ServiceProvider` class and have the `provider` property point straight to an `Agent`.

There are no other properties in the `ServiceProvider` class and the semantics of the service provider could be understood through the use of the property.

The service use case is still satisfied even without having the `ServiceProvider` class.

@nishakm - any thoughts?

stevenc-stb: yes... provider property point straight to an Agent.

model/Service/Classes/SoftwareService.md
bact: Will be resolved by #1002

model/Service/Classes/ServiceProvider.md
bact: Will be resolved by #1002

stevenc-stb (commented):
With PR the https://github.com/spdx/spdx-3-model/pull/1002/files  #1002  I see no issues. 

The Model.Drawio and model-service.png appear out-of-date

goneall:
> With PR the https://github.com/spdx/spdx-3-model/pull/1002/files #1002 I see no issues.
> 
> The Model.Drawio and model-service.png appear out-of-date

I'll update the drawio with the changes

goneall:
@bact @stevenc-stb - I just updated the diagrams - if you could do a quick review and let me know if I missed anything

images/model-Service.png
bact, line 1: - serviceHostingCountries -> serviceHostingCountry (singular)

images/model-Core.png
bact, line 1: The "CountryCodeAlpha3" heading in Simple Data Types may need to be in **bold face**, like other headings in the same box.

docs/model.drawio
```diff
@@ -1627,4 +1633,44 @@
...
+          <mxGeometry y="26" width="344" height="30" as="geometry" />
+        </mxCell>
+        <mxCell id="KPL05Jm-FdMU-5SZgOHn-6" value="+ serverAuthenticationProtocol:&amp;nbsp;AuthenticationProtocolType[0..*]" style="text;strokeColor=none;fillColor=none;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;fontFamily=Verdana;fontSize=10;" parent="KPL05Jm-FdMU-5SZgOHn-4" vertex="1">
+          <mxGeometry y="56" width="344" height="30" as="geometry" />
+        </mxCell>
+        <mxCell id="KPL05Jm-FdMU-5SZgOHn-7" value="+ serviceHostingCountries: CountryCodeAlpha3[0..*]" style="text;strokeColor=none;fillColor=none;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;fontFamily=Verdana;fontSize=10;" parent="KPL05Jm-FdMU-5SZgOHn-4" vertex="1">
```
bact:
```suggestion
        <mxCell id="KPL05Jm-FdMU-5SZgOHn-7" value="+ serviceHostingCountry: CountryCodeAlpha3[0..*]" style="text;strokeColor=none;fillColor=none;align=left;verticalAlign=middle;spacingLeft=4;spacingRight=4;overflow=hidden;points=[[0,0.5],[1,0.5]];portConstraint=eastwest;rotatable=0;whiteSpace=wrap;html=1;fontFamily=Verdana;fontSize=10;" parent="KPL05Jm-FdMU-5SZgOHn-4" vertex="1">
```

model/Service/Properties/serviceHostingCountry.md
```diff
@@ -0,0 +1,21 @@
...
+# serviceHostingCountry
+
+## Summary
+
+Specifies a country code where a software service is hosted.
+
+## Description
+
+This property should specify all countries and territories where a service is located.
+This includes indirect locations (e.g. where the data may be hosted).
```
bact:
```suggestion
Specifies a country or territory where a software service is located.
This includes indirect locations (e.g., where the data may be hosted).
```

Update description to reflect the singular nature of the property name.

bact (changes requested):
serviceHostingCountry is singular

stevenc-stb (approved):
This looks OK.

goneall:
@bact - I updated the diagram and accepted your change to wording above - let me know if you see anything else.

bact (approved):
Thank you. Looks good to me.

bact:
> The Service profile is not available for review.

This is probably "***now*** available for review".

goneall:
Since this has 2 review approvals, I'll go ahead and merge.  We can always make additional PR's if we find any issues.

PR #974, opened 2025-02-14
Capitalised the first word in SupportType and DatasetType entry descriptions, make consistent with other vocabs

bact:
- Capitalised the first word in entry description in DatatsetType. DatasetAvailabilityType, and SupportType (make it the same as other vocabs)
- <s>Make examples in DatasetType more generic (Wikipedia article -> encyclopedic article / removing a reference to "Tom Hanks" which can be cultural specific)</s>


bact:
@bennetkl @rgopikrishnan91 minor updates here for consistency, if you can do a quick pass and mark approval (as a comment is good as well). Thanks.

bact:
@bobmartin3000 Will the "Tom Hanks" thing be flagged by ISO? I just noticed this, sorry.

bennetkl (commented):
@bact   Art,  We reviewed this PR in 0305 WG Meeting, and it was decided to delay merge this fix until 4.0 since it would impact backward compatibility.   Bob also mentioned that no changes of this type until we get ISO certification.  The change seems simple enough that we could delay it I hope you agreee.

bobmartin3000:
As Gary noted we agreed to wait on these.

bact:
I'm fine to delay this but I disagree that this will introduce any compatibility issue.

Most of them are about changing a lowercase letter to an uppercase letter.
It doesn't introduce any new field, remove anything or alter any semantic.

(also, delaying this to 3.1 is understandable, but to 4.0 is questionable - as there are PRs that introduce more changes than this PR and they are targeting 3.1 / For example, #994 is removing paragraphs and potentially targeting "3.0.2")

The only "substantial" change in this PR is removing references to Wikipedia and Tom Hanks.
And if that sounds too big, we can remove that particular change from the change set.

bact:
I'm happy to remove any change that may cause a compatibility issue with 3.x from the change set if you can kindly point to them.

bobmartin3000:
The Tom Hanks one is the part that should be delayed.

bact:
Thank you Bob, I will remove that part then.

Update: Tom Hanks and Wikipedia restored. The concerned parts are removed from the changeset. This should be ok to merge now.

zvr (approved):

PR #995, opened 2025-03-06
Update `develop` branch name in Contributing.md

bact:
Update development branch name to be the current name: `develop`


bact:
Duplicated with a subset of #972

PR #999, opened 2025-03-07
Delete model/Operations/Vocabularies/applicationType.md

bact:
Already use model/Operations/Vocabularies/ApplicationType.md (uppercase A in Application) instead.


bact:
@umm0 @mkurzman in case you haven't aware of this yet. thank you.

umm0 (approved):

PR #1000, opened 2025-03-07
Formatting entries in /Operations/ApplicationType

bact:
- Normalise casing
- Add fullstops


goneall:
It looks like ApplicationType is no longer in the operations profile - removing

PR #1002, opened 2025-03-08
Remove the ServiceProvider class

goneall:
See this discussion for context: https://github.com/spdx/spdx-3-model/pull/973#discussion_r1981661406


goneall:
@stevenc-stb - this should resolve both the issues you raised in #973 - please take a look and let me know if I missed anything

bact (approved):

goneall:
From the tech call on 11 March 2025, no objections to merging this

PR #1001, opened 2025-03-08
Update CHANGELOG.md for post-3.0.1 changes

bact:
Log changes from PRs merged after [3.0.1 release](https://github.com/spdx/spdx-3-model/releases/tag/3.0.1).

PRs that change/update the model and will go to 3.1:

- #966
- #1019

Typos/formatting/error fixing PRs that don't change the model, which also backport to 3.0.x:

| PR description | To `develop` branch | To `support/3.0` branch |
|-|-|-|
|  Make formatting of SupportType and DatasetType entry descriptions consistent with other vocabs | #974 | #1037 |
| VulnAssessmentRelationship example fixes | #994 | #1018 |
|  Fix email RFC link in ExternalIdentifierType vocab | #1003 |  #1006 |
|  Fix kilowatt-hour value typo in EnergyConsumptionDescription | #1004 | #1007 |
| Fix typos: provenence, licences | #1040 | #1038 |

- Please also see a question in #996.

Issue #996, opened 2025-03-06
Do we need a patch release for possible ISO review changes?


goneall:
Reference conversation in this pull request: https://github.com/spdx/spdx-3-model/pull/994#event-16619418845

If we get feedback requiring any changes to the model, we'll need to have some mechanism for a release.

I would suggest we create a milestone 3.0.2 for a patch release.  We can base the changes off the support branch.

bact:
This will mean that in the ISO publication it will be called "SPDX 3.0.2"?

Implication is all the IRIs have to be updated as well, since we have the PATCH version part in the IRI pattern.

bact:
Or we can have a "3.0.1-iso" milestone?

"3.0.1-iso" name will indicate that this is a spec document that the textual content (for human consumption) is similar to the ISO publication -- but there is no difference in the semantic model from "3.0.1", so we can keep the same IRIs for terms.

This "3.0.1-iso" release will be published to the "3.0.1" spec website https://spdx.github.io/spdx-spec/v3.0.1/ - replacing the current "3.0.1".

Keeping the same IRIs and website will make it less work for updating tools/docs (some of them may hard coded the IRIs).

bact:
Track changes since 3.0.1, see change log updates in #1001
(items with an asterisk * are changes that may go to `support/3.0` branch as well)

bact:
For v2.2.1, these are the tags we have

- [v2.2.1](https://github.com/spdx/spdx-spec/releases/tag/v2.2.1)
- [v2.2.1-ISO-submittal](https://github.com/spdx/spdx-spec/releases/tag/v2.2.1-ISO-submittal)
- [v2.2.1-ISO-final](https://github.com/spdx/spdx-spec/releases/tag/v2.2.1-ISO-final)

bact:
Note: a "3.0.2" milestone has been created https://github.com/spdx/spdx-3-model/milestone/8

goneall:
Since we have a 3.0.2 milestone, I'll close this issue.  We can decide what we label the release if and when we get to that point.

PR #1004, opened 2025-03-09
fix kilowatt-hour typo

ilans:


bact (approved):

bact:
- @bobmartin3000 you may want to be aware this typo fix.
- We may want to fix this in `support/3.0` branch too. See also #996

bact:
I will merge this obvious correction and log it to #1001, so we can have a single place to see track changes (and decide which one we like to have it applied to the support branch as well).

bact:
See this fix for `support/3.0` branch at:
- #1007

PR #1009, opened 2025-03-13
New structure proposal for Operations profile

mkurzman:
The structure was adapted in a way to follow a supply chain idea with upstream/inbound on the left side, downstream/outbound on the right side and the delivery and different operation relationships in the middle. The original fields were still kept but would need to be aligned in the group as a next step.


bact:
Unrelated to the change introduced in this PR, but the validation workflow still failed because "name" metadata of the property "reviews" in model/Operations/properties/reviews.md is "relationType".

Please change that line (Line 15) to "name: reviews".

Or, better, "name: review" and rename the filename to "review.md".
Because, by convention, the SPDX 3 property name should be in the singular form.

goneall:
It looks like this PR is outdated - I'll go ahead and close it - @mkurzman - feel free to re-open a new PR based on the current profile-operations if needed

PR #1016, opened 2025-04-11
Fix paragraph in zh-Hans trans. in NoAssertionElement

bact:
> ***This is a PR against support/3.0 branch.***

What this PR does:

1) In Simplified Chinese translation of the Description section, add the missing blank line after the bullet list in Markdown source to enforce the correct formatting.

    - The last paragraph of the Description should be separated from the last item in the bullet list.

        zh-Hans text

    - English text for comparison:

        en text

2) Restored the original typographic emphasis of classes and entry values.

    Current Simplified Chinese (zh-Hans) text:

    > 例如，关系`relationshipType=ancestorOf`、`from=Element1`和`to=NoAssertionElement`明确表示不对`Element1`的任何潜在后代作出任何断言。

    Current English (en) text :

    > For example, a Relationship with `relationshipType`="ancestorOf", `from`=Element1, and `to`=NoAssertionElement is explicitly expressing that no assertion is being made about any potential descendants of Element1.

    it is possible that the rationale for current style of English text is:

    - relationshipType, from, to: in `fixed-width font`, indicates that they are class or property or datatype.
    - ancestorOf: in between "quotation marks", indicates that is is an entry value (of RelationshipType).
    - Element1, NoAssertionElement: in regular font, without any typographic device, indicates that they are instances.

--

Btw, on (2), although the style is different from the original English text, the Simplified Chinese translators are actually trying to enforce more consistent style across all of its text (see other files in the model for comparison). Classes and properties are all in `fixed-width font` across zh-Hans. Texts in English version, by contrast, is less consistent in the application of styles.

We may have to discuss and agreed upon consistent style choice, documented it, and apply it to every languages.
This has been raised before, during the 3.0.1 editorial process for ISO submission and postponed to 3.1 instead:
- https://github.com/spdx/spdx-spec/issues/1000

### DIFF
--- a/model/Core/Individuals/NoAssertionElement.md
+++ b/model/Core/Individuals/NoAssertionElement.md
@@ -45,4 +45,10 @@ no assertion is being made about any potential descendants of Element1.
 - SPDX创建者尝试但无法达成合理的客观判断；
 - SPDX创建者未尝试确定此字段；
 - SPDX创建者故意不提供信息（这样做不应被暗示为有任何含义）。
-例如，关系`relationshipType=ancestorOf`、`from=Element1`和`to=NoAssertionElement`明确表示不对`Element1`的任何潜在后代作出任何断言。
+
+例如，关系
+`relationshipType`="ancestorOf"、
+`from`=Element1
+和
+`to`=NoAssertionElement
+明确表示不对 Element1 的任何潜在后代作出任何断言。


bact:
@gomico @ZhengZhenyu while point (2) may need longer discussion, I think point (1) is obvious.

If you can approve point (1) fix, we can selectively merge only point (1) for now.

gomico:
Sorry for the late reply. Can confirm that modification in point 1 is correct and should be merged.

bact:
Thanks. I have created another PR (#1021) to fix only point 1 (newline) and will leave this one open for point 2 (style) discussion.

goneall:
@bact - should this be merged?  If so, please rebase otherwise close.

bact:
We can close for now

PR #1019, opened 2025-04-19
Update info about SPDX 3 JSON

bact:
- Use "SPDX 3 JSON" name, per 15 Apr 2025 Tech Team meeting
  - This PR update instances in spdx-3-model repo, while https://github.com/spdx/spdx-spec/pull/1197 (merged) does the same for spdx-spec repo
- Update links to use new branch structure
- Add validation doc link to README
- Add table of contents to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,6 @@ standard.
 Translations of the information model may be available.
 English remains the normative language in all cases.
 
-[![Core Profile diagram](./images/model-Core.png "Core Profile diagram")](./images/model-Core.png)
-
 ## Table of contents
 
 - [Branch structure](#branch-structure)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -18,7 +18,20 @@ standard.
 Translations of the information model may be available.
 English remains the normative language in all cases.
 
-## Branch Structure
+[![Core Profile diagram](./images/model-Core.png "Core Profile diagram")](./images/model-Core.png)
```
goneall: Do we want to have the diagram here?  Seems like we'd have to scroll more to get to the other content.

serialization/README.md
```diff
@@ -4,7 +4,7 @@
 
 For information on what can be included in an SPDX serialization and how they
 are structured, please refer to the
-[Serialization information](https://github.com/spdx/spdx-spec/blob/development/v3.0.1/docs/serializations.md)
+[Model and serializations](https://github.com/spdx/spdx-spec/blob/develop/docs/serializations.md)
```
goneall: Why the change in title to include Model?

goneall (commented):
Thanks @bact  - just a couple comments for consideration.

serialization/README.md
bact: I took the title from the document at the URL.  It is Chapter 6 of the spec.

bact: I am now update the link to the specific section, retain the previous section reference, and update the text that it is a "Serialization information" section within a "Model and serializations" chapter.

README.md
bact: Agree that it's too much of the scroll. Too larger than I was expected.
I have removed the full size diagram from here and instead add a 120-pixel-height version under the Model heading. What do you think?

goneall (approved):
Changes LGTM - Thanks @bact

PR #1023, opened 2025-05-01
boundaryParameters -> boundaryParameter

bact:
Property name should be in a singular form.


bact:
@stevenc-stb per https://github.com/spdx/spdx-3-model/issues/226, the property name should be in singular

stevenc-stb:
Agreed. look good

bact:
Thank you. I will merge this to the profile-hardware branch.

PR #1030, opened 2025-06-07
Update SoftwareService description

### LINKED ISSUE #1013: Suggested changes for SoftwareService description
The Description for SoftwareService at https://github.com/spdx/spdx-3-model/blob/service-profile/model/Service/Classes/SoftwareService.md currently says:

> Description
> SoftwareService represents software being licensed, delivered and accessed online over a network.

I'd suggest removing "licensed" and "delivered" from this description:

* "licensed": Typically, SaaS services are provided pursuant to services agreements that grant permissions to e.g. "access and use" the SaaS service. This is quite different from "licensing" as typically used in the SPDX context, since SaaS services generally don't grant e.g. copyright licenses to the underlying software that's running on the server. 
* "delivered": This may lead to confusion, as any downloaded software would probably be considered "software being... delivered... over a network".

I'd also probably suggest tweaking the reference to "software" for similar reasons, since what is really being provided by SaaS services is the "service" of access to the hosted system, rather than the underlying software itself.

So I'd suggest using something like this as the description instead:

> Description
> SoftwareService represents a software-powered service offered for access online over a network.

bact:
To avoid confusion as discussed in #1013

### DIFF
--- a/model/Service/Classes/SoftwareService.md
+++ b/model/Service/Classes/SoftwareService.md
@@ -8,7 +8,7 @@ Software provided as a service over a network.
 
 ## Description
 
-SoftwareService represents software being licensed, delivered and accessed online over a network.
+SoftwareService represents a service based on software components offered for access online over a network.
 
 ## Properties
 


goneall (approved):
LGTM

swinslow (approved):

PR #1037, opened 2025-07-02
Fix inconsistent vocab entry format

bact:
> It is the equivalent of #974 which was already approved and merged to `develop`.

Update the formatting of entry descriptions in DatatsetType, DatasetAvailabilityType, SupportType, etc, to make their format consistent with other vocabs.

- Uppercase the first letter of the description
- End the description with a fullstop


goneall (approved):
LGTM

PR #1041, opened 2025-07-05
Fix Hardware relationshipType, formatting and typos

bact:
> This PR targets `profile-hardware` branch.

- Fix relationship type:
  - `Generated` -> `generates`
  - `effected` -> `affects`
- Formatting:
  - Make "Relationship:" heading consistent across Hardware profile
  - MkDocs needs a newline to separate paragraphs for readability
- Typos:
  - For Each -> For each
  - class or subclass class -> class or subclass
  - an `/Core/Element` -> a `/Core/Element`
  - an `planProcess` -> a `PlanProcess` (it is a class, starts with uppercase)
  - an `Requirements` -> a `Requirements` (the class may need to be renamed to its singular form separately though)

Still pending issues:

- These relationshipTypes are not exist:
  - `Creator` (used in BoundaryDefinitionAction)
  - `resolution` (used in ResolutionAction)

### DIFF
--- a/model/Hardware/Classes/BoundaryDefinitionAction.md
+++ b/model/Hardware/Classes/BoundaryDefinitionAction.md
@@ -8,10 +8,11 @@ The boundary definition is used to define boundaries.
 
 ## Description
 
-Boundaries can be physical or abstract. This is the act of defining the boundaries. 
+Boundaries can be physical or abstract. This is the act of defining the boundaries.
 
-Relationship: 
-For Each `BoundaryDefinitionAction` there is one and only one `/Core/Relationship` class or subclass with the relationshipType of 'Creator’ on the to and an `/Core/Agent` class or subclass on the from. 
+Relationship:
+
+For each `BoundaryDefinitionAction` there is one and only one `/Core/Relationship` class or subclass with the relationshipType of 'Creator’ on the to and a `/Core/Agent` class or subclass on the from.
 
 ## Metadata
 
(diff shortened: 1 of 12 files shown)


bact:
@stevenc-stb There are two non-existing relationship types: "Creator" and "resolution", not sure if they are typos or just new relationship types waiting to be defined. (in any case, the relationship type name should starts with lowercase).

stevenc-stb:
There is some relationship type that will need some updates; 
affects - right for Vulnerability only. 
creator -- we need to add 
hasOutput - right now is build only but should change 
hasInput- right now is build only but should change
resolution- we need to add.

bact:
Thank you. We can add that later but please note that relationship types tend to be in the verb forms. So we may need something like "creates" and "resolves".

@stevenc-stb apart from the additional relationships above, do you agree with the typo fixes and suggestions in the PR? Thank you

bact:
@stevenc-stb for now, do you agree with the  typos and suggestions in the PR? Thank you

stevenc-stb:
> Thank you. We can add that later but please note that relationship types tend to be in the verb forms. So we may need something like "creates" and "resolves".
> 
> @stevenc-stb apart from the additional relationships above, do you agree with the typo fixes and suggestions in the PR? Thank you
I agree with the typo fixes

bact:
Thank you @stevenc-stb - I will merge this first (to `profile-hardware` branch) and the relationship type additions/updates can be sorted in another PR.

PR #1047, opened 2025-07-08
[3.0] "SPDX-3.0" -> "SPDX 3.0"

bact:
> This PR targets `support/3.0` branch.
> For a similar PR targets `develop`, see #1048

- To fix https://github.com/spdx/spdx-spec/issues/1237 (ISO issue)

PR #1048, opened 2025-07-08
[develop] "SPDX-3.0" -> "SDPX 3.1"

### LINKED ISSUE #1237: Consistency when referring to version
There are uses of “SPDX-3.0” and “SPDX 3.0” (and possibly others).

The spec should only refer to second-level numbering (never 3.0.1, for example), and in a consistent manner "SPDX 3.0".

The RDF metadata should also only refer to the same numbering (to be fixed in the model repo).

bact:
> For a similar PR targets `support/3.0`, see #1047

- To fix https://github.com/spdx/spdx-spec/issues/1237 (ISO issue)
- Note that some of these may able to just use "SPDX 3" if the context allowed it to be applied to 3.0, 3.1, and any future 3 releases.


bact:
Per Tech team meeting on 19 Aug 2025, if the context allows the use of "SPDX 3" (without minor version number), consider to use it to avoid update work in the future releases.

Updated.

goneall (approved):
LGTM

zvr (approved):

PR #1049, opened 2025-07-08
[develop] "this specification" -> "this document"

### LINKED ISSUE #1234: Spec should be referred to as "this document"
Only "this document" is allowed.

There are case of "this standard", "this specification", ...

bact:
> For a similar PR targets `support/3.0`, see #1050 

- To fix https://github.com/spdx/spdx-spec/issues/1234 (ISO issue)

See also the same fix in spdx-spec repo:
- https://github.com/spdx/spdx-spec/pull/1239


zvr (approved):

goneall (approved):
LGTM

PR #1050, opened 2025-07-08
[3.0] "this specification" -> "this document"

bact:
> For a similar PR targets `develop`, see #1049

- To fix https://github.com/spdx/spdx-spec/issues/1234 (ISO issue)

See also the same fix in spdx-spec repo:
- https://github.com/spdx/spdx-spec/pull/1238


zvr (approved):

goneall (approved):
LGTM

PR #1052, opened 2025-07-09
[develop] RDF URL "3.0.1" -> "3.1"

### LINKED ISSUE #1046: The RDFs should only go to second-level version
For publishing the specification, RDF should remain constant, even though textual changes happen in the specification.

Therefore, instead of `https://spdx.org/rdf/3.0.1/terms/Core/Element` they should simply be `https://spdx.org/rdf/3.0/terms/Core/Element`

These are defined in the profile definition files (like `Core.md`).

See also https://github.com/spdx/spdx-spec/issues/1237

bact:
> For a similar PR that targets `support/3.0`, see #1051

- To fix #1046 (ISO issue)

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,15 +68,15 @@ The editable files are automatically processed by
 and the following are generated:
 
 - Input for [MkDocs](https://www.mkdocs.org/), which then generates the
-  [specification](https://spdx.github.io/spdx-spec/v3.0.1/)
+  [specification](https://spdx.github.io/spdx-spec/v3.1-dev/)
 - [JSON-LD context](http://niem.github.io/json/reference/json-ld/context/)
-  file: [spdx-context.jsonld](https://spdx.org/rdf/3.0.1/spdx-context.jsonld)
+  file: [spdx-context.jsonld](https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-context.jsonld)
 - Model [SHACL](https://en.wikipedia.org/wiki/SHACL) and
   [OWL](https://www.w3.org/OWL/) files:
   - [Turtle format](https://en.wikipedia.org/wiki/Turtle_(syntax)):
-    [spdx-model.ttl](https://spdx.org/rdf/3.0.1/spdx-model.ttl)
+    [spdx-model.ttl](https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-model.ttl)
   - [JSON-LD format](https://json-ld.org/):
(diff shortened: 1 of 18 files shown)


zvr (changes requested):
All these should become 3.1

bact:
Done. All term IRIs are now 3.1.

Notes:

- Since 3.1 resource URLs are not available yet, I'm now putting them as generic "X.Y"  (instead of either 3.0 or 3.1) in serialization documentation.
  - We can update this later when the resources are available under 3.1 URLs or we can keep them as generic as that (so no need to update it again in 3.2)
- The command line examples in validation documentation will also need to be updated later as well when the 3.1 URLs are ready.
  - Or we can change them now if we don't mind that all of them will not be runnable.

README.md
```diff
@@ -68,15 +68,15 @@ The editable files are automatically processed by
 and the following are generated:
 
 - Input for [MkDocs](https://www.mkdocs.org/), which then generates the
-  [specification](https://spdx.github.io/spdx-spec/v3.0.1/)
+  [specification](https://spdx.github.io/spdx-spec/v3.1-dev/)
 - [JSON-LD context](http://niem.github.io/json/reference/json-ld/context/)
-  file: [spdx-context.jsonld](https://spdx.org/rdf/3.0.1/spdx-context.jsonld)
+  file: [spdx-context.jsonld](https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-context.jsonld)
```
bact, line 73: Use this github.io URL for now, as the https://spdx.org/rdf/3.1/spdx-context.jsonld is not yet available.

README.md
```diff
@@ -68,15 +68,15 @@ The editable files are automatically processed by
...
 - Model [SHACL](https://en.wikipedia.org/wiki/SHACL) and
   [OWL](https://www.w3.org/OWL/) files:
   - [Turtle format](https://en.wikipedia.org/wiki/Turtle_(syntax)):
-    [spdx-model.ttl](https://spdx.org/rdf/3.0.1/spdx-model.ttl)
+    [spdx-model.ttl](https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-model.ttl)
```
bact, line 77: Use this github.io URL for now, as the https://spdx.org/rdf/3.1/spdx-model.ttl is not yet available.

goneall (approved):
LGTM

zvr (approved):

PR #1053, opened 2025-07-09
Fix Markdown list issue in Simplified Chinese

bact:
Missing space after the number for numbered list, makes it not rendered as a list.


goneall (approved):
LGTM

kestewart (approved):
Thanks for cleaning this up.

PR #1057, opened 2025-07-15
Clarify 'exploited' property in Security Profile

### LINKED ISSUE #652: Remove 'exploited' field in model/Security/Classes/ExploitCatalogVulnAssessmentRelationship.md
Per my understanding, a vulnerability is entered into an exploit catalog such as KEV is only when it has been already exploited.
If we know that it has been already exploited, why do we need a Boolean field known as 'exploited' to indicate if it has been
exploited or not? I suggest removing this field (unless there's a case when a vulnerability will be return back to
'unexploited' from 'exploited', which I don't think it will ever will).

### LINKED ISSUE #912: [SHACL] exploited always true
From /Security/exploited:
"This field is set when a CVE is listed in an exploit catalog."

But it is mandatory in `ExploitCatalogVulnAssessmentRelationship` together with `catalogType` and `locator`.

So isn’t it always true?...

In that case, I don't understand why this field even exists.
Should I add a SHACL rule to enforce always `true`?

rnjudge:
Update the description of the property so it is more descriptive and usable. Discussed this change in the July 15th Security call.

Resolves #912
Resolves #652

### DIFF
--- a/model/Security/Properties/exploited.md
+++ b/model/Security/Properties/exploited.md
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Describe that a CVE is known to have an exploit because it's been listed in an exploit catalog.
+Denote whether a CVE is present in an exploit catalog.
 
 ## Description
 
-This field is set when a CVE is listed in an exploit catalog.
+This field is set to True when a CVE is present in an exploit catalog and False when the CVE is not present.
 
 ## Metadata
 


model/Security/Properties/exploited.md
```diff
@@ -4,11 +4,11 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Describe that a CVE is known to have an exploit because it's been listed in an exploit catalog.
+Denote whether a CVE is present in an exploit catalog.
 
 ## Description
 
-This field is set when a CVE is listed in an exploit catalog.
+This field is set to True when a CVE is present in an exploit catalog and False when the CVE is not present.
```
bact, line 11: True/False -- FYI, I don't think we have a clear style guide on this but other xsd:boolean instances (all from ExpandedLicensing) use "true" and "false" (lowercase, sometimes within a quote).
- https://spdx.github.io/spdx-spec/v3.0.1/model/ExpandedLicensing/Properties/isFsfLibre/
- https://spdx.github.io/spdx-spec/v3.0.1/model/ExpandedLicensing/Properties/isOsiApproved/
- https://spdx.github.io/spdx-spec/v3.0.1/model/ExpandedLicensing/Properties/isDeprecatedAdditionId/
- https://spdx.github.io/spdx-spec/v3.0.1/model/ExpandedLicensing/Properties/isDeprecatedLicenseId/

The canonical serialization also prefers lowercase, but this is specific to one form of serialization and not necessary applies to the model description in general: https://spdx.github.io/spdx-spec/v3.0.1/serializations/#canonical-serialization

bact (approved):

goneall (approved):
LGTM

PR #1061, opened 2025-07-22
Adding safety profile png and model for Requirement Class 

nicpappler:
# Added an overview png of the currently stable Classes etc. of the Safety Profile
Note: currently only the details for requirement are in this PR, but all the content of this png is stable for a release candidate, the details will follow in the next PRs

# Added the Requirement class, RequirementContext class and its properties:

## Why we used Element as parent, and not Artifact:
- The CreationInfo class already has everything we need to identify Agent and the creation date, tools used to create the requirement and a comment.
- The verifiedUsing can be re-used to indicate the verification Method to verify the requirement (review, inspection, simulation etc.) - as it is currently just defined as a string, we are open to fill it with any text we want 

## Why we use and enhance LifeCycleScopeType:
Instead of creating something very similar, we re-use LifecycleScopeType to explain to which development Stage this requirement belongs to and add the lifecycle phases Update and Decommission

## Why we use and enhance ContentIdentifierType
Instead of creating something very similar, re-use ContentIdentifier, as a Requirement UID (unique-ID, requested by all quality and safety standards for requirements management)
To work for a requirement UID, the list of possible ContentIdentifierTypes needs to be enhanced, the actual identifier value is a string as already defined.


model/Core/Vocabularies/LifecycleScopeType.md
```diff
@@ -21,4 +21,6 @@ This enumeration summarizes common phases when dependency and other relationship
 - build: A relationship has specific context implications during an element's build phase, during development.
 - test: A relationship has specific context implications during an element's testing phase, during development.
 - runtime: A relationship has specific context implications during the execution phase of an element.
+- update: A relationship has specific context implications for a product update
+- decommission: A relationship has specific implications for a product's retirement and/or decommissioning
```
bact:
```suggestion
- update: A relationship has specific context implications for a product update.
- decommission: A relationship has specific implications for a product's retirement and/or decommissioning.
```

model/FunctionalSafety/Classes/RequirementContext.md
```diff
@@ -0,0 +1,31 @@
...
+
+# RequirementContext
+
+## Summary
+
+Context of the requirement relating it to a project and its structure
+
+## Description
+
+The requirement context describes the the hierarchical level of the requirement within a project. It does not give any idea where this requirement ends up in the final full system. It defines the project as a context for the requirement, e.g. Zephyr, Xen, Linux, any Application, any Library etc.; don't make it optional, as requirements levels without the context to the project will cause chaos in final systems. if not sure, don't use RequirementLevel in your Requirement Class at all.
```
bact: Is "RequirementLevel" exists elsewhere? Or should it be "requirementHierarchyLevel" ?

model/Software/Vocabularies/ContentIdentifierType.md
```diff
@@ -18,3 +18,4 @@ ContentIdentifierType specifies the type of a content identifier.
...
 - swhid: SoftWare Hash IDentifier, a persistent intrinsic identifier for digital artifacts, such as files, trees (also known as directories or folders), commits, and other objects typically found in version control systems. The format of the identifiers is defined in the [SWHID specification](https://www.swhid.org/specification/v1.1/4.Syntax) (ISO/IEC DIS 18670). They typically look like `swh:1:cnt:94a9ed024d3859793618152ea559a168bbcbb5e2`.
+- ReqUID: the UID used by a requirements management or any other lifecycle management tool to uinquly identify a requirement item. UID, or unique ID, is a standard term in requirements engineering.
```
bact: Entry names should be in camelCase (first word is in lowercase).

model/FunctionalSafety/Properties/requirementUID.md
```diff
@@ -0,0 +1,17 @@
...
+
+# requirementUID
+
+## Summary
+
+Provides an unique Requirement ID
+
+## Description
+
+An unique ID, as used by requirements management or other lifecycle management tools, usually a string that follows a project specific naming convention.
```
bact:
```suggestion
A unique ID, as used by requirements management or other lifecycle management tools, usually a string that follows a project specific naming convention.
```

model/FunctionalSafety/Properties/requirementUID.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementUID
+
+## Summary
+
+Provides an unique Requirement ID
```
bact:
```suggestion
Provides a unique Requirement ID.
```

model/FunctionalSafety/Classes/RequirementContext.md
```diff
@@ -0,0 +1,31 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# RequirementContext
+
+## Summary
+
+Context of the requirement relating it to a project and its structure
```
bact:
```suggestion
Context of the requirement relating it to a project and its structure.
```

model/FunctionalSafety/Properties/contextOrigin.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# contextOrigin
+
+## Summary
+
+Origin of the requirement context
```
bact:
```suggestion
Origin of the requirement context.
```

model/FunctionalSafety/Properties/contextOrigin.md
```diff
@@ -0,0 +1,18 @@
...
+# contextOrigin
+
+## Summary
+
+Origin of the requirement context 
+
+## Description
+
+Specification, why the context is like it is. Describes what it means, where the details of the Context are defined, can be a Specification, a Requirements Management Plan, a conference discussion etc.
+
```
bact, line 12:
```suggestion
```

model/FunctionalSafety/Properties/devLifeCycleStage.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# devLifeCycleStage
+
+## Summary
+
+The product lifecycle phase, the requirement is applicable for
```
bact:
```suggestion
The product lifecycle phase, the requirement is applicable for.
```

model/FunctionalSafety/Properties/requirementContext.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementContext
+
+## Summary
+
+Context of the requirement relating it to a project and its structure
```
bact:
```suggestion
Context of the requirement relating it to a project and its structure.
```

model/FunctionalSafety/Properties/requirementContext.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Description
+
+The requirement context describes the the hierarchical level of the requirement within a project. It does not give any idea where this requirement ends up in the final full system. It defines the project as a context for the requirement, e.g. Zephyr, Xen, Linux, any Application, any Library etc.; don't make it optional, as requirements levels without the context to the project will cause chaos in final systems. if not sure, don't use RequirementLevel in your Requirement Class at all.
+
```
bact, line 12:
```suggestion
```

model/FunctionalSafety/Properties/requirementHierachyLevel.md
```diff
@@ -0,0 +1,18 @@
...
+# requirementHierachyLevel
+
+## Summary
+
+Number indicating the hierachical level of the requirement whithin its namespace
+
+## Description
+
+The level of the requirement within the project that gave the requirement its namespace, using a positive number, e.g. in Zephyr: 1 for the Zephyr Safety Claims, 2 for Zephyr system requirements and 3 for Zephyr Software requirements 
+
```
bact:
```suggestion
```

model/FunctionalSafety/Properties/requirementNote.md
```diff
@@ -0,0 +1,18 @@
...
+# requirementNote
+
+## Summary
+
+Text used to define the rational or additional information
+
+## Description
+
+additional detail used to define the reason/rational, why this requirement is there or give some additional information, needed to understand the requirement. Usually less formal than the requirement wording of the RequirementText.
+
```
bact:
```suggestion
```

model/FunctionalSafety/Properties/requirementNote.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementNote
+
+## Summary
+
+Text used to define the rational or additional information
```
bact:
```suggestion
Text used to define the rational or additional information.
```

model/FunctionalSafety/Properties/requirementNamespace.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementNamespace
+
+## Summary
+
+Namespace of the requirement
```
bact:
```suggestion
Namespace of the requirement.
```

model/FunctionalSafety/Properties/requirementHierachyLevel.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementHierachyLevel
+
+## Summary
+
+Number indicating the hierachical level of the requirement whithin its namespace
```
bact:
```suggestion
Number indicating the hierarchical level of the requirement within its namespace.
```

model/FunctionalSafety/Classes/RequirementContext.md
```diff
@@ -0,0 +1,31 @@
...
+- name: RequirementContext
+- SubclassOf: none
+
+## Properties
+
+- requirementNamespace
+  - type:xsd:string
+  - minCount: 0
+  - maxCount: 1
+- requirementHierachyLevel
```
bact:
```suggestion
- requirementHierarchyLevel
```

model/FunctionalSafety/Properties/contextOrigin.md
```diff
@@ -0,0 +1,18 @@
...
+# contextOrigin
+
+## Summary
+
+Origin of the requirement context 
+
+## Description
+
+Specification, why the context is like it is. Describes what it means, where the details of the Context are defined, can be a Specification, a Requirements Management Plan, a conference discussion etc.
```
bact, line 11: Does "Requirements Management Plan" intentionally put with Caps as a proper noun or a special term?

model/FunctionalSafety/Properties/requirementNote.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Description
+
+additional detail used to define the reason/rational, why this requirement is there or give some additional information, needed to understand the requirement. Usually less formal than the requirement wording of the RequirementText.
```
bact:
```suggestion
Additional detail used to define the reason/rational, why this requirement is there or give some additional information, needed to understand the requirement. Usually less formal than the requirement wording of the RequirementText.
```

model/FunctionalSafety/Properties/requirementText.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementText
+
+## Summary
+
+the statement describing the actual need defined by the requirement
```
bact:
```suggestion
The statement describing the actual need defined by the requirement.
```

model/FunctionalSafety/Properties/requirementText.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Description
+
+ the text portion of the requirement, usually following specific rules and best practices of requirements engineering. e.g. EARS
```
bact:
```suggestion
The text portion of the requirement, usually following specific rules and best practices of requirements engineering. e.g. EARS.
```

model/FunctionalSafety/Properties/requirementHierachyLevel.md
```diff
@@ -0,0 +1,18 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementHierachyLevel
```
bact:
```suggestion
# requirementHierarchyLevel
```

Need to rename the file as well

bact (changes requested):
Few typos

model/FunctionalSafety/Classes/RequirementContext.md
nicpappler: Right, it's a leftover, for when RequirementContext was still named RequirementLevel.

model/FunctionalSafety/Properties/contextOrigin.md
nicpappler: It is a special term in most engineering process standards (for QM or safety or similar), that's why it is in caps.

model/Software/Vocabularies/ContentIdentifierType.md
nicpappler: ok

model/FunctionalSafety/Properties/contextOrigin.md
bact: Thank you. Resolved

model/FunctionalSafety/Classes/Requirement.md
```diff
@@ -0,0 +1,40 @@
...
+- name: Requirement
+- SubclassOf: /Core/Element
+
+## Properties
+
+- requirementUID
+  - type: ContentIdentifier
+  - minCount: 0
+  - maxCount: 1
+- DevLifeCycleStage
```
bact, line 24:
```suggestion
- devLifeCycleStage
```

model/Software/Vocabularies/ContentIdentifierType.md
```diff
@@ -18,3 +18,4 @@ ContentIdentifierType specifies the type of a content identifier.
...
 - swhid: SoftWare Hash IDentifier, a persistent intrinsic identifier for digital artifacts, such as files, trees (also known as directories or folders), commits, and other objects typically found in version control systems. The format of the identifiers is defined in the [SWHID specification](https://www.swhid.org/specification/v1.1/4.Syntax) (ISO/IEC DIS 18670). They typically look like `swh:1:cnt:94a9ed024d3859793618152ea559a168bbcbb5e2`.
+- reqUID: the UID used by a requirements management or any other lifecycle management tool to uinquly identify a requirement item. UID, or unique ID, is a standard term in requirements engineering.
```
bact:
```suggestion
- reqUID: the UID used by a requirements management or any other lifecycle management tool to uniquely identify a requirement item. UID, or unique ID, is a standard term in requirements engineering.
```

bact (changes requested):

model/FunctionalSafety/Classes/RequirementContext.md
```diff
@@ -0,0 +1,35 @@
...
+
+- requirementNamespace
+  - type:xsd:string
+  - minCount: 0
+  - maxCount: 1
+<<<<<<< HEAD
+- requirementHieratchyLevel
+=======
+- requirementHierarchyLevel
+>>>>>>> 20a5242082bf8888c11507c5a1f27d4826129717
```
bact, line 28:
```suggestion
- requirementHierarchyLevel
```

bact (changes requested):

stevenc-stb:
The Requirement class should be in core because it is a core requirement for other profiles. 
It look like right now it going in Functional safety profile?

model/FunctionalSafety/Classes/Requirement.md
nicpappler: Just fixed this in the model - Requirements Class is now moved to Core, as the HW team suggested.

model/FunctionalSafety/Classes/RequirementContext.md
nicpappler: fixed now in the RequirementContext class that has been moved to Core

model/Software/Vocabularies/ContentIdentifierType.md
nicpappler: fixed

nicpappler:
> The Requirement class should be in core because it is a core requirement for other profiles. It look like right now it going in Functional safety profile?

-> moved to Core, including the Context class and the Properties

model/Core/Vocabularies/LifecycleScopeType.md
```diff
@@ -21,4 +21,6 @@ This enumeration summarizes common phases when dependency and other relationship
 - build: A relationship has specific context implications during an element's build phase, during development.
 - test: A relationship has specific context implications during an element's testing phase, during development.
 - runtime: A relationship has specific context implications during the execution phase of an element.
+- update: A relationship has specific context implications for a product update.
+- decommission: A relationship has specific implications for a product's retirement and/or decommissioning.
```
bact:
```suggestion
- decommission: A relationship has specific context implications for a product's retirement and/or decommissioning.
```

Should we have "context" qualifier here the same as other entries?

bact (changes requested):

model/FunctionalSafety/Classes/RequirementContext.md
bact: Thank you. I can see it now.
So should we need to remove this file and Requirement.md from FunctionalSafety/Classes directory?

model/FunctionalSafety/Classes/Requirement.md
```diff
@@ -0,0 +1,40 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Requirement
```
bact, line 3: As there is already Requirement.md in Core/Classess, should we delete this file? thank you.

bact (changes requested):

model/FunctionalSafety/Properties/requirementHierarchyLevel.md
```diff
@@ -0,0 +1,17 @@
...
+
+Number indicating the hierarchical level of the requirement within its namespace.
+
+## Description
+
+The level of the requirement within the project that gave the requirement its namespace, using a positive number, e.g. in Zephyr: 1 for the Zephyr Safety Claims, 2 for Zephyr system requirements and 3 for Zephyr Software requirements 
+
+## Metadata
+
+- name: requirementNamespace
```
bact, line 15:
```suggestion
- name: requirementHierarchyLevel
```

The name here does not match the filename and the name at the title.

model/FunctionalSafety/Properties/requirementHierarchyLevel.md
```diff
@@ -0,0 +1,17 @@
...
+
+# requirementHierarchyLevel
+
+## Summary
+
+Number indicating the hierarchical level of the requirement within its namespace.
+
+## Description
+
+The level of the requirement within the project that gave the requirement its namespace, using a positive number, e.g. in Zephyr: 1 for the Zephyr Safety Claims, 2 for Zephyr system requirements and 3 for Zephyr Software requirements
```
bact, line 11:
```suggestion
The level of the requirement within the project that gave the requirement its namespace, using a positive number, e.g. in Zephyr: 1 for the Zephyr Safety Claims, 2 for Zephyr system requirements and 3 for Zephyr Software requirements.
```

bact (changes requested):

model/Core/Properties/requirementNote.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementNote
```
stanislaw, line 3: Is it possible to create a separate rationale field while keeping `requirementNote` for general comments?

Typically, a note/comment/remark field is used to capture explanations that help others understand the requirement itself, rather than the reason it exists. Very often, these comments appear after a requirement has been created, during discussions between teams or suppliers. In that context, comments record the answers to questions raised when someone is trying to interpret the meaning of the requirement.

The rationale serves a different function: it explains why the requirement exists. This purpose is already partly covered by the current Description field.

I would propose splitting this into two fields:

`requirementRationale` and `requirementNote` and making the `requirementRationale` a mandatory field, while keeping `requirementNote` optional.

Note that INCOSE defines rationale as `A1 - Rationale` attribute.

A quick summary of all fields can be found on the page 136 of the INCOSE "Guide to Writing Requirements":

https://www.incose.org/docs/default-source/working-groups/requirements-wg/gtwr/incose_rwg_gtwr_v4_040423_final_drafts.pdf?sfvrsn=5c877fc7_2

model/Core/Properties/requirementText.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementText
+
+## Summary
+
+The statement describing the actual need defined by the requirement.
```
stanislaw, line 7: I would vote for calling this field `requirementStatement`.

Depending on a project, the word "text" can mean different things, such as a general comment (NOTE, REMARK, COMMENT), a rationale, e.g., RATIONALE, or many other things, likely meaning a comment, not a statement.

Making this requirement field a `statement` would pinpoint it.

As a reference, the Guide to Writing Requirements always calls the requirement statements `statements` (the word appears 601 times in this guide).

https://www.incose.org/docs/default-source/working-groups/requirements-wg/gtwr/incose_rwg_gtwr_v4_040423_final_drafts.pdf?sfvrsn=5c877fc7_2

model/Core/Properties/contextOrigin.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# contextOrigin
```
stanislaw, line 3: The existence of this field in the proposal may suggest that the current model lacks a concept for Chapters/Sections that group requirements together.

Once a requirement document grows large, it becomes very difficult to read and write requirements without any structural organization. If the SPDX model does not include explicit support for document structure, reviewing a large set of requirements becomes harder. In such cases, the origin field alone is insufficient, because both the document origin and the chapter/section context are important. Many requirement documents have deeply nested structures, so ideally, SPDX should support this hierarchy explicitly.

By comparison, the SDoc format already provides explicit modeling of document structures. Having a similar feature in SPDX would greatly improve usability.

❗ Plain, unstructured lists of requirements are very difficult to work with, especially at scale. ❗

model/Core/Properties/devLifeCycleStage.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# devLifeCycleStage
```
stanislaw, line 3: The earlier comment on `contextOrigin` partially applies here as well.

It is good that this field is optional, since I have personally never seen it used in either projects or textbooks. But is this information truly necessary if SPDX were to support sections/chapters and explicit document structure? In that case, the devLifeCycle’s `product lifecycle phase the requirement is applicable for` could instead be inferred from its context - namely, the enclosing section, chapter, or document.

In any case, it would be great to have a chapter in this file called `## Example` where two or three requirement example would be given with an explanation who and how would use this requirement.

model/Core/Properties/requirementHierarchyLevel.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementHierarchyLevel
```
stanislaw, line 3: My earlier point about modeling document structures (see [this comment](https://github.com/spdx/spdx-3-model/pull/1061/files?utm_source=chatgpt.com#r2324881244)) also applies here.

Fixing a requirement’s hierarchy/type level as an attribute could reduce portability, since its position within an enclosing section, chapter, document, or specification might change.

If SPDX does not support document structures, users may end up with a flat list of requirements where requirementHierarchyLevel values vary (1, 2, 3, …) but provide no real context about how those requirements are related.

A concrete example: in the case of Zephyr requirements, will all of them be exported into a single SPDX file, or will each Zephyr document get its own SPDX document? If the latter, then the hierarchy level field becomes redundant, because one document would contain only level 1 requirements, another only level 2, and so on.

model/Core/Properties/requirementNamespace.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementNamespace
```
stanislaw, line 3: The purpose of this field is clear, but I am missing a concrete example of how it would actually be used.

For instance, does this imply that SPDX-conformant tools must support queries like `find all requirements for ZEP::* or XEN::*`? A realistic use case example would help clarify the intent.

Another unclear aspect concerns re-used requirements: what happens if a requirement is reused multiple times? For example, suppose a requirement originates in namespace FOO, then is reused in project BAR, and later again in BAZ. Which namespace value should be recorded as the requirementNamespace? Either way, in practice this will look like this:

```
BAR::BAR-001
BAR::BAR-002
ZEP::ZEP-003 (is this how reuse supposed to work?)
```

Does re-use mean reusing without changing any other information meta information?

 A practical example covering this case would be very useful.

model/Core/Properties/requirementContext.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementContext
+
+## Summary
+
+Context of the requirement relating it to a project and its structure.
```
stanislaw, line 7: My comment about `contextOrigin` strongly applies here as well: https://github.com/spdx/spdx-3-model/pull/1061/files#r2324881244.

Would it not make sense to identify the context for an enclosing document and then have all requirements inherit this property by lookups. This makes requirements more portable and reduces repetition.

What defines the context is there this requirement appears, e.g., in which document, section, subsection, etc.

model/Core/Classes/Requirement.md
```diff
@@ -0,0 +1,40 @@
+SPDX-License-Identifier: Community-Spec-1.0
```
stanislaw, line 1: I am giving this feedback as a developer of [StrictDoc](https://github.com/strictdoc-project/strictdoc), which also models Requirement as one of the core types.  

StrictDoc deliberately reduces a number of requirement fields to just the following:

- MID (machine-generated unique identifier, optional but recommended for robust change-tracking)
- UID (human-generated, e.g., REQ-001)
- TITLE
- STATEMENT
- RATIONALE
- COMMENT (multiple comments possible)

All other fields can be created/customized by a user if they need them for a project.

Does the requirement model of SPDX envision a similar kind of extensibility? If so, I would propose that three layers of requirements fields are maintained:

1) The core fields that every project and every requirement must have (I would keep it the same as what StrictDoc has)

2) The optional fields that some projects would want to have, e.g., `STATUS`, `TAGS`, etc. It could be discussed which fields SPDX could recommend as such default but optional fields.

3) The user-provided fields that users are free to choose given their context.

It would be great to align on these sets and flexibility because this would greatly simplify the mental mapping and programming of the converters between SDPX, SDoc and any other interfaces.

I would also like to mention the INCOSE guidelines for writing requirements which could be cross-checked for creating such a list of mandatory/optional/user-provided fields:

- [Guide to Writing Requirements](https://www.incose.org/docs/default-source/working-groups/requirements-wg/gtwr/incose_rwg_gtwr_v4_040423_final_drafts.pdf?sfvrsn=5c877fc7_2)
- [Needs and Requirements Manual (NRM) version 2 (soft copy)](https://portal.incose.org/commerce/store?productId=INCOSE-NRVVLM)

model/Core/Properties/requirementNote.md
bact: II think requirementNote is redundant of `/Core/comment`.

We also have `/Core/description`.

There should be text to describe when to use which property (including one for rationale).

gregshue: According to [description.md](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/description.md):

> The intent is to provide recipients of the SPDX file with a detailed technical explanation of the functionality, anticipated use, and anticipated implementation of the Element.
>
> This field may also include a description of improvements over prior versions of the Element.

The `description` field clearly goes beyond the limits of a requirement statement.

According to [comment.md](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/comment.md):

> A comment is an optional field for creators of the Element to provide comments to the readers/reviewers of the document.

The `comment` field may work for holding Notes about the requirement statement. Notes on requirements must be non-normative content. OTOH, I believe Rationale is required in many cases and is evaluated as part of a conformance audit. In order to generate requirements/design tracing information the `requirement` and `rationale` statements must be distinct from `comment` and `description`.

model/Core/Properties/requirementText.md
gregshue: I agree with Stanislav. "requirementStatement" is more appropriate. 

BTW, I ran across presentations on EARS by Alistair Maven (it's creator) where he said that Textual presentation only effectively covers about 80% of the needs. The remaining portion of requirements are more effectively communicated with tables, diagrams, charts, etc.

model/Core/Properties/contextOrigin.md
gregshue: `context` and `contextOrigin` seem to contain information I would expect to find in Assumptions of Use, Architecture Decision, and Architecture Design documents - which all must(?) be included in the requirements tracing.

Stanislav's concern for element organization within a document is very necessary for human consumption. Some of the organization can be taken care of through the document tree. Other parts of the organization must be done within an individual document. System Architecture documentation has evolved to describing the system from multiple viewpoints. An early example of this is [the "4+1 Architectural View Model](https://en.wikipedia.org/wiki/4%2B1_architectural_view_model). In my experience each viewpoint requires a different organizational sub-structure.

Every time I go through a system decomposition cycle the organization of the output layer documents vary from each other and none are identical to the organization of the input layer.

Whenever I bring in content from a 3rd party component I may wrap the original content with an internal document assigning new statement identifiers and presenting it in a different organization. 

Support for organizing content is very practical.

model/Core/Properties/devLifeCycleStage.md
gregshue: @stanislaw wrote:

> the devLifeCycle’s product lifecycle phase the requirement is applicable for could instead be inferred from its context - namely, the enclosing section, chapter, or ...

INCOSE recommends requirement statements be self-contained and stand-alone. Nothing should be left for inference from the document context of the statement.

model/Core/Properties/requirementHierarchyLevel.md
gregshue: @stanislaw wrote:

> Fixing a requirement’s hierarchy/type level as an attribute ...

I first thought the requirementHeirarchyLevel represented the document outline structure, but that doesn't seem to be true. Safety (and security) requirements cut across functional and non-functional requirements. AFAICT High-level vs low-level requirements really represent the inputs to and outputs from an Architectural Decomposition/Design cycle. NASA documents indicate their mission designs go through 5-8 decomposition cycles, so multiple cycles is nothing new. When I tried mapping Zephyr ecosystem into supporting a distributed IoT solution I found the System Architecture needed to go through at least two cycles; Solution -> Products/Tools -> Elements (e.g., PCBA, case, immutable bootloader, mutable device FW). Then I could map all the MCU FW images to be supported by the same product line architecture implemented as an extended version the Zephyr ecosystem. (The FW for resource-rich devices would come out of a different product line codebase. The mobile application images would come out of a 3rd product line codebase).

I think `requirementHeirarcyLevel` needs some more clarification and modeling before it is practical.

model/Core/Properties/requirementNamespace.md
gregshue: @stanislaw I think the mapping of one requirement namespace/ID to another must be done without modifying the SBOMs of the constituent 3rd party components.

Perhaps one good example would be how NordicSemi should describe their NCS fork of Zephyr. IIRC, Nordic has changed some Zephyr files (e.g., symbols, algorithms, tools) in a non-backwards-compatible way so this is a leverage-and-potential-reuse paradigm instead of a reuse-and-extend paradigm. It would be natural for NordicSemi to want to insert a prefix before each of the namespaces they inherited.

bact:
Note that if our requirement is outside the SBOM, we can also use [ExternalRef](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Classes/ExternalRef.md).
With that, we may like to add a new entry (perhaps "requirement") in [ExternalRefType](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Vocabularies/ExternalRefType.md).

nicpappler:
To implement all the changes requested in the discussions of the last two weeks, this PR is now replaced by https://github.com/spdx/spdx-3-model/pull/1141

bact:
> To implement all the changes requested in the discussions of the last two weeks, this PR is now replaced by #1141

Should we close this PR?

goneall:
Superseded by PR #1141

PR #1063, opened 2025-07-25
Operations: profile-operations reset

umm0:
Reset of the operations branch to enable tracking accepted content vs. draft.

The branch is reset to 'develop' with adding the Drawio state from 'profile-hardware'.

Content for the Operations profile is to be addes again with dedicated pull requests.

@mkurzman please acknowledge the reset.


umm0:
@goneall can help me with this?
I am not to sure merging my branch will have the desired result or if this should rather be done by pulling the state from my branch an force pushing it.

goneall:
@umm0 - I'm not sure I have enough context to really help too much.  Is this PR just a rebase of the develop branch + the updated draw.io file?

Looking at the files changed, there are quite a few differences outside of the operations profile directory - so it doesn't look like it includes all of the changes.

Perhaps rebased spdx:profile-operations to the develop branch first to bring it up to current levels then rebasing the profile-operations-reset branch to sync it with the spdx:profile-operations would get things in sync.

bact:
Maybe we should single out the draw.io diagram and manage it separately.

The file is 1) huge, 2) more likely to conflict with other new profiles, 3) not stable (open and save without edit can still produce diffs) = not very ideal to work with in parallel.

umm0:
Hi @goneall,
this is not just a rebase but a reset. We decided to go back to the current state of 'develop' in the branch.

This is necessary because with all the ideas and discussions the past months, we do not have a clear state in the branch as is.
Rather than going through the current state discussing and removing bit by bit, what is outdated (we tried and failed with that), we think it will get us there faster if we start again with a clean canvas and add to that again.

We retain the old state on umm0/spdx-3-model and boschglobal/spdx-3-model.
If you think it makes sense to have it here also, it would be great if you could add a profile-operations-legacy branch.

umm0:
If we go with retaining the old state here, I would also be fine with leaving the profile-operations branch as is and adding a new one (e.g. profile-operations-rc or profile-operations-2025).
We mainly need a way to set focus on sorting out, what is ready for the RC vs. what is outdated or too unrefined for now.

bact:
I the past edits are retained in `umm0/spdx-3-model` and `boschglobal/spdx-3-model`, I think it's ok to reset `spdx:profile-operations` to `spdx:develop`.

Then we can have a set of PRs (selected from `umm0/spdx-3-model` and `boschglobal/spdx-3-model`) gradually added to `spdx:profile-operations`.

goneall:
If we don't need the history in this branch, then just a reset should be fine. 

It looks like there is one merge conflict with the model.drawio file that's preventing a merge.  @umm0 - if you can resolve the conflict, I'll go ahead and merge.

goneall:
Actually - I can just do a force push of a reset to the profile-operations branch since I have permissions.

If you want to post the git commands to reset to the develop branch, I can execute them.

umm0:
I am not exactly a git expert - so I hope this makes sense. I would do

```
git checkout develop
git pull
git push -u origin develop:profile-operations --force
```

then this pull request should be just one commit ahead (the one for the drawio).

goneall:
> I am not exactly a git expert - so I hope this makes sense. I would do
> 
> ```
> git checkout develop
> git pull
> git push -u origin develop:profile-operations --force
> ```
> 
> then this pull request should be just one commit ahead (the one for the drawio).

Always nervous when I rewrite history.

@zvr @bact @JPEWdev - does the above git commands look good for resetting the operations profile to the top of the develop branch?

JPEWdev:
Maybe; we can look at it in the call tomorrow? I just want to make sure I understand the operation that is being attempted before I confirm.

Maybe something like this better would be:
```
git checkout -b profile-operations origin/profile-operations
git reset --hard origin/develop
git push -f
```

I usually try to avoid checking out and touching branches that aren't relevant to the operations and using tracking branches as much as possible has the best protections IMHO.

umm0:
Thank you @JPEWdev ,
your approach should accomplish the goal as well. And you are right, using tracking is the safer way.

goneall:
@umm0 - The branch has been reset - you may need to rebase your branch

umm0:
My branch was based on develop. It looks like your push did not include the most recent commits from develop (4 commits behind).
But his should not be a problem. Merging my branch should remedy this, since it includes the 4 commits difference - plus the two from me that are ahead of 'develop'.

PR #1066, opened 2025-07-30
Add Dataset/inLanguage

### LINKED ISSUE #744: 3.1: Dataset: Add a "language" field
AI team meeting 2024-04-24 made a decision to explore about the addition of a field/fields about language/linguistic properties of a dataset and a model.

While some repository may made this language filed a mandatory metadata,
we should keep in mind that not every datasets are having language aspect (or even human-associated data)
and not every models are language models.

No Assertion or No Relevant could be a possible value.

More discussion to follow.

bact:
`inLanguage` to indicates the language of a dataset. 

- The name of the property is from Schema.org: https://schema.org/inLanguage
  - Which is also used by MLCommons Croissant dataset https://docs.mlcommons.org/croissant/docs/croissant-spec.html#recommended
- Using [IETF BCP 47 language tag](https://datatracker.ietf.org/doc/rfc5646/) which is recommended by Schema.org and considered a web standard (also used in RDF)
  - It is also used by Hugging Face data card https://github.com/huggingface/hub-docs/blob/main/datasetcard.md
- See more at #744

--

If there is a use case of use this `inLanguage` to describe language in `File`, `Regulation`, `Annotation` or other artifacts, we may move this property to `Core` profile instead.

To resolve #744

### DIFF
--- a/model/Core/Datatypes/LanguageTag.md
+++ b/model/Core/Datatypes/LanguageTag.md
@@ -1,0 +1,36 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# LanguageTag
+
+## Summary
+
+A string constrained to the IETF BCP 47 language tag format.
+
+## Description
+
+The string must be in the format as describes in the Section 2 of
+[IETF RFC 5646](https://datatracker.ietf.org/doc/rfc5646/) (part of BCP 47).
+
+It provides a standardized way of indicating the language of the content or
+performance or used in an action.
+
+*Example*
+
+- `es` (Spanish)
+- `es-UY` (Spanish as used in Uruguay)
+- `hy-Latn-IT-arevela` (Eastern Armenian written in Latin script, as used in Italy)
+- `sl-rozaj-biske` (San Giorgio dialect of Resian dialect of Slovenian)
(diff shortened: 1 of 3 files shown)


stevenc-stb (approved):
LGTM

bact:
Discussed in 30 July 2025 AI team call - agreed to have this field and keep it in Dataset profile, not Core as there is no strong enough use case outside of the Dataset profile.

bact:
Note - a LanguageTag datatype is added to have some constraints over the format of the string.

goneall (changes requested):
The descriptions should be updated to describe whether the Language referred to is "Human Language" or the more restrictive "Natural Language" to disambiguate it from a computer language.  IETF BCP 47 refers to human language, but I'm not sure if that is the intent of this usage.

Reference: [Wikipedia Language definition](https://en.wikipedia.org/wiki/IETF_language_tag).

bact:
Thank you @goneall. Yes, the property and the datatype are intended for human language.
I have updated the description to clarify that.

goneall (approved):
LGTM - Thanks @bact

PR #1067, opened 2025-07-30
Revise standardName and standardCompliance descriptions

### LINKED ISSUE #893: Specify relationship between standardName and standardCompliance
## Background

In SPDX 3.0, we have two properties related to standards:
- [Core/standardName](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Properties/standardName/)
  - Summary:
    > The name of a relevant standard that may apply to an artifact.
  - Description:
    > Various standards may be relevant to useful to capture for specific artifacts.
- [AI/standardCompliance](https://spdx.github.io/spdx-spec/v3.0.1/model/AI/Properties/standardCompliance/)
  - Summary:
    > Captures a standard that is being complied with.
  - Description:
    > A free-form text that captures a standard that the AI software complies with.
    > This includes both published and unpublished standards, such as those developed by ISO, IEEE, and ETSI.
    > The standard may, but is not necessarily required to, satisfy a legal or regulatory requirement.

## What can be improved

There are at least two things we can improved here:

1) Amend the summary/description to explicitly specify the difference between the two properties and how they can working together
2) Considering revise the description and making the `standardCompliance` to be more generic and allow it to be use in non-AI context (in a non-breaking way for SPDX 3.x)
  - Moving the property to Core Profile will change the IRI and it's a break. But amending the description will not.

## Proposals

There can be at least two options:

1) Keep both and use them in different way (that can compliment each other):

    Currently, in the [AI BOM whitepaper](https://www.linuxfoundation.org/research/ai-bom) (p. 56), @bennetkl distinguished the two in this way:
    - `standardName` - standards adhered to but that compliance was not obtained
    - `standardCompliance` - standards that compliance are obtained (for example, from a third-party attestation or certification)

    standardName vs standardCompliance

2) Drop/deprecate `standardCompliance`

    Modify model description of `standardCompliance` to say that this property is deprecate and `standardName` is preferred.

    From this commit comment, https://github.com/spdx/spdx-3-model/commit/027a1fa20efca7bdd3160fc42f81fbd36c03f151 , it looks like `standardName` was created with AI and Dataset Profiles in mind in the beginning.

### See also
- #387 proposes additional properties (or Class) related to standards

bact:
To address the ambiguous relationship between `Core/standardName` and `AI/standardCompliance`, as discussed in #893.

- We might need to thinking should we like to move `AI/standardCompliance` to Core profile or not, as it can be use with other artifacts as well.
  - If we chose this, it means we will deprecate `AI/standardCompliance`; and create new `Core/standardCompliance` and let it be a property of an `Artifact`.

Other related relationship and class:
- [`hasSpecification`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Vocabularies/RelationshipType.md) relationship type: "Every `to` Element is a specification for the `from` Element (`from` hasSpecification `to`), during a LifecycleScopeType period."
  - This may not imply compliance.
- `conformsTo` relationship type, to be added in 3.1: "The `from` Element conforms to each `to` Regulation." 
- Hardware profile, to be added in 3.1, adds [`Specification`](https://github.com/spdx/spdx-3-model/blob/profile-hardware/model/Core/Classes/Specification.md) class which can also include standard ("specifications are referenced in this class. Requirements, standards, specification and processes are referenced in this class.")
  - As `Specification` is also an Element, we can have `from` hasSpecification `Specification`.

To resolve  #893

### DIFF
--- a/model/AI/Properties/standardCompliance.md
+++ b/model/AI/Properties/standardCompliance.md
@@ -4,17 +4,20 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Captures a standard that is being complied with.
+Captures a standard that an artifact is being complied with.
 
 ## Description
 
-A free-form text that captures a standard that the AI software complies with.
+A free-form text that captures a standard that an artifact complies with.
 
 This includes both published and unpublished standards, such as those developed
-by ISO, IEEE, and ETSI.
+by ETSI, IEEE, IETF, ISO, and W3C, as well as national standards bodies.
 
 The standard may, but is not necessarily required to, satisfy a legal or
 regulatory requirement.
+
+If the artifact is using a standard as a reference or guideline, but not
+necessarily compliant with it, use the `standardName` property instead.
 
 ## Metadata
 
--- a/model/AI/Properties/standardCompliance.md
+++ b/model/AI/Properties/standardCompliance.md
@@ -10,14 +10,14 @@ Captures a standard that an artifact is being complied with.
 
 A free-form text that captures a standard that an artifact complies with.
 
-This includes both published and unpublished standards, such as those developed
-by ETSI, IEEE, IETF, ISO, and W3C, as well as national standards bodies.
-
 The standard may, but is not necessarily required to, satisfy a legal or
 regulatory requirement.
 
 If the artifact is using a standard as a reference or guideline, but not
 necessarily compliant with it, use the `standardName` property instead.
+
+For a detailed compliance information, please consider defining
+a `Relationship` with "conformsTo" relationship type to a `Regulation`.
 
 ## Metadata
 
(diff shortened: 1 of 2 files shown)


bact:
Suggestions from AI Team call 30 July 2025 about standard compliance:

- Good to add descriptions to differentiate between the two properties - no objection.
- Not a good idea to move standardCompliance to Core profile
- Way forward, "conformsTo" relationship (with Regulation/Specification) is better way to describe compliance. 
As a Relationship, we can also provide evidence, etc. of the compliance - instead of just a declared free-form text field.
  - We may like to add this suggestion at the end of `standardCompliance` description.

kestewart (approved):
Makes sense to clarify this aspect.

goneall (approved):
LGTM

bact:
Note that

1) Relationship type "conformsTo" is in Hardware profile, which is under review at PR https://github.com/spdx/spdx-3-model/pull/1076 
2) `standardCompliance` property is still in AI profile

PR #1069, opened 2025-07-30
Let DatasetPackage uses artifactSize instead

bact:
Based on discussion between @bennetkl and me in https://github.com/spdx/spdx-3-model/pull/966#pullrequestreview-2587185097

- We propose to deprecate `Dataset/datasetSize` and suggest to use `Software/artifactSize` property instead for `DatasetPackage`
- As `DatasetPackage` is also `SoftwareArtifact`, it will automatically have `artifactSize` as one of its properties
- Using the same property, it will allow easier size comparison between `DatasetPackage` and other `SoftwareArtifact`
- `artifactSize` is new in SPDX 3.1 (from #966) and has the same range as `datasetSize` (`xsd:nonNegativeInteger`), so should be straightforward to migrate 

This PR does not actually change the model, it just add a deprecation notes to `datasetSize`


bobmartin3000 (approved):
LGTM

bennetkl (approved):
No issues in using artifactSize for dataset size.  Approved.

goneall (approved):
Agree with the changes

kestewart (approved):

PR #1073, opened 2025-08-07
Requirements synchronization with Fusa/Hardware

stevenc-stb:
Requirements synchronization with Fusa/Hardware


model/Core/Classes/Requirement.md
```diff
@@ -0,0 +1,39 @@
...
+
+- requirementUID
+  - type: xsd:anyURI
+  - minCount: 0
+  - maxCount: 1
+- devLifeCycleStage
+  - type: LifecycleScopeType
+  - minCount: 0
+  - maxCount: *
+- requirementText
```
bact, line 28: Is it possible to reuse `/Core/description` here?
It is available to all Elements.
https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Properties/description/

Then we can add something similar to this to the `Requirement`'s description above:

> Statement describing the actual need can be put inside `description`.
>
> Rational or additional information can be put inside `comment`.

model/Core/Classes/Requirement.md
```diff
@@ -0,0 +1,39 @@
...
+  - type: xsd:string
+  - minCount: 1
+  - maxCount: 1
+- requirementContext
+  - type: RequirementContext
+  - minCount: 0
+  - maxCount: 1
+- requirementNote
```
bact, line 36: Is it possible to reuse `/Core/comment` here?
It is available to all Elements.

https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Properties/comment/

bact (commented):
I wonder if we can reuse few properties here.

model/Core/Classes/Requirement.md
stevenc-stb: We have a this talk about it in the Fusa group.  most current view is that the fielids the requirementText and requirementNote are need. @nicpappler comments?

bact: We can just merge this and amend it later.
I just ask because of curiosity.

mkurzman:
Decisions of todays profile operations meeting:
1. Project class will be based on Core/Bundle => With this the "context" of bundle may replace some of the properties that can deleted then
2. The "VulnAssessmentRelationship" has a different nature with the "Vulnerabilities" as tiggers. Ummo will provide a new individual proposal.
3. as the nature or OperationsAssessment and VulnAssessment is different, this would also make the Core/AssessmentRelationship obsolete and can thus be deleted.

stevenc-stb:
> Decisions of todays profile operations meeting:
> 
>     1. Project class will be based on Core/Bundle => With this the "context" of bundle may replace some of the properties that can deleted then
> 
>     2. The "VulnAssessmentRelationship" has a different nature with the "Vulnerabilities" as tiggers. Ummo will provide a new individual proposal.
> 
>     3. as the nature or OperationsAssessment and VulnAssessment is different, this would also make the Core/AssessmentRelationship obsolete and can thus be deleted.
i think that this text is for PR  https://github.com/spdx/spdx-3-model/pull/1074/files

bact (approved):
I don't see any issue with syntax/formatting and general spellings.
Should be good to get merge now (to `profile-hardware`) and we can amend things in other PRs.

goneall (approved):
Discussed on the tech call 12 August 2023 - Approved

PR #1094, opened 2025-08-21
[develop] "Profile" -> "profile"

bact:

In model description and elsewhere, change "Profile" (starts with upper case) to "profile" (starts with lower case):

- To fix editorial issue as explained in https://github.com/spdx/spdx-spec/issues/1250
- Also update headings in "Glossary of terms" to be a sentence case, see https://github.com/spdx/spdx-spec/issues/1232


goneall (approved):
I'm not sure all of the changes are required for the ISO update, but the changes do make the case consistent

PR #1095, opened 2025-08-21
[3.0] "Profile" -> "profile"

bact:

Change "Profile" (starts with upper case) to "profile" (starts with lower case):

- To fix editorial issue as explained in https://github.com/spdx/spdx-spec/issues/1250
- Also update headings in "Glossary of terms" to be a sentence case, see https://github.com/spdx/spdx-spec/issues/1232


bact:
@RexJaeschke fyi. Apart from the spdx-spec repo, we need some changes here as well.

goneall (approved):

bact:
Merged. We still need to fix the "Profile" (uppercase at first letter) in https://github.com/spdx/spdx-spec/blob/support/3.0/docs/conformance.md . I will open another PR for that.

PR #1108, opened 2025-09-26
Fix Vulnerability example (locator)

bact:
> For similar PR that targets `support/3.0` branch, see #1107.

Since the locator and identifierLocator cardinality are 0..*, we need list [...] here.


goneall (approved):
LGTM

kestewart (approved):

PR #1109, opened 2025-10-02
Profile hardware improve  ProductSpecification

stevenc-stb:
ProductSpecification added 
- partNumber
- itemVerson
- intendedUse


model/Hardware/Classes/ProductSpecification.md
```diff
@@ -20,3 +20,11 @@ A product specification (product spec) is a detailed document that outlines the
...
 - hazard
   - type: /Core/DefinedType
+- partNumber
+  - type: xsd:string
+  - minCount: 1
+- itemVerson
+  - type: xsd:string
+- intendedUse
+  - type: xsd:string
+  - maxCount: 1
```
goneall: Would it be possible and/or useful to have more than one intendedUse?  Since intendedUse is a string, you can always create a string describing multiple uses, so allowing multiple may not be necessary.

goneall: After reading the description for the field, I agree with the cardinality 0..1 since it is describing the boundaries of use.

goneall (approved):
LGTM - I left one review comment, but I don't think a change is warranted.

model/Hardware/Properties/itemVerson.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# itemVerson
```
bact:
```suggestion
# itemVersion
```

Also at the filename.

model/Hardware/Properties/itemVerson.md
```diff
@@ -0,0 +1,17 @@
...
+
+Version identifier for the item.
+
+## Description
+
+Describes the specific version of item.
+
+## Metadata
+
+- name: itemVerson
```
bact:
```suggestion
- name: itemVersion
```

bact (changes requested):

model/Hardware/Classes/ProductSpecification.md
```diff
@@ -20,3 +20,11 @@ A product specification (product spec) is a detailed document that outlines the
 
 - hazard
   - type: /Core/DefinedType
+- partNumber
+  - type: xsd:string
+  - minCount: 1
+- itemVerson
```
bact:
```suggestion
- itemVersion
```

bact (changes requested):

bact:
Dataset profile also has [`intendedUse`](model/Dataset/Properties/intendedUse.md) with similar semantic:

> Describes what the given dataset should be used for.

Compares it to what proposed here for the Hardware profile:

> The intendedUse property is designed to capture a summary of how or for what purpose a product is meant to be used.

Does it possible to have this in Core profile instead - and having both Hardware and Dataset classes use this? The summary of `/Core/intendedUse` can be more generic like:

> The intendedUse property holds a summary describing how the element is meant to be used.

The `/Dataset/DatasetPackage` can deprecate the use of `/Dataset/intendedUse` in favour of the new `/Core/intendedUse`.

(The `/Dataset/intendedUse` will be kept, so not a breaking change. See example of the deprecation of `/Dataset/datasetSize` in favour of the new `/Software/artifactSize` [here](https://github.com/spdx/spdx-3-model/blob/develop/model/Dataset/Properties/datasetSize.md)).

goneall (approved):
Updates look good - thanks @stevenc-stb

kestewart (approved):
Thanks for the changes, LGTM.

kestewart:
We discussed deprecating intended use in Dataset on tech call,  so going ahead and merging this for now into hardware branch.

bact:
@kestewart Thank you. I will make another PR to deprecate the intendedUse in DatasetPackage. Ping @rgopikrishnan91 @bennetkl

PR #1090, opened 2025-08-19
Create promptModality.md

ElyasRashno:
The promptModality field has been added. Please review @kestewart , @bact , @bennetkl , @zvr , @goneall , and @rgopikrishnan91 .

PR #1085, opened 2025-08-19
Create hasRAG.md

ElyasRashno:
The hasRAG field has been added. Please review @kestewart , @bact , @bennetkl , @zvr , @goneall , and @rgopikrishnan91 .

PR #1082, opened 2025-08-19
Create promptPattern.md

ElyasRashno:
promptPattern field has been added here. Please review @kestewart , @bact , @bennetkl , @zvr , @goneall , and @rgopikrishnan91 .

PR #1078, opened 2025-08-19
Create ModalityType.md

ElyasRashno:
Adding modality type as part of changes to the AI profile to represent the prompt and agent. Please review @kestewart  , @bact , @bennetkl

PR #1110, opened 2025-10-02
Add AI Prompt

bact:
Based on 1 Oct 2025 AI WG meeting, this PR includes changes needed to supporting AI prompt.

It incorporates some information from the model diagram at https://lists.spdx.org/g/spdx-ai/message/338 and @ElyasRashno PRs:

- #1090 -> `/Core/modality`
- #1078 -> `/Core/Modality`
- #1085 -> `/AI/isContextAugmented`
- #1082 -> `/AI/promptPattern`

Prompt chaining can be described by using the new "follows" relationship type, see:

- #1116

**Note that more properties can be added later, this is just a skeleton.**

ai-prompt

> When `isContextAugmented` is set to `true`, a Relationship of type `usesTool`
  can be optionally used to describe the context augmentation mechanism or tool
  employed (e.g., referencing a retrieval-augmented generation tool).

```mermaid
graph LR;
  B(Prompt) -->|usesTool| A(RAG)
```

## Design notes

There are few redesigns (see notes below) and revisions in class/property definitions, please look at the changes in each individual file.

1) Subclassing **`/AI/Prompt`** from `/Software/SoftwareArtifact`:

    - It represents a reasonable unit of content that can be attributed to a specific source.
    - It is lighter than a `/Dataset/DatasetPackage` and requires less modification to the external properties restrictions.
    - Reuse `/Software/contentIdentifier` for prompt content URI. The type of content identifier URI can be either "Gitoid" or "SWHID" (if needed, a new content identifier type can be added to `/Software/ContentIdentifierType`).

2) The **`/Core/contentModality`** (property) and **`/Core/Modality`** (vocab) are trying to be generic enough for use cases in foundation model and in human-computer interaction.
    - Note that modality is more about mode of communication/interaction, and less about data format, but it's not 100% dichotomy here (see also [Hugging Face modalities](https://huggingface.co/datasets), which is also mixed).
    - Alternatively, we can adopt a much simpler vocab from Schema.org. Schema.org [`accessMode`](https://schema.org/accessMode) has 4 approved values: "auditory", "tactile", "textual", "visual". See: [Schema.org Accessibility Properties for Discoverability Vocabulary](https://www.w3.org/community/reports/a11y-discov-vocab/CG-FINAL-vocabulary-20241209/#accessMode-vocabulary).
    - In any case, we can make a decision to also include [`/Core/contentType`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/contentType.md) which indicate media type of a content. Since it is based on MIME type, it is limited, but should cover general use cases (like differentiate still image from moving image, or tell if the "textual" input is actually a CSV data).

3) Use **`isContextAugmented`** name, instead of `hasRAG`, as we may like to focus on the inclusion of automatically augmented context/supporting content to the prompt content, rather than the technique (RAG, etc).

4) Some existing definitions are slightly revised to make it accommodate more use cases or make it more precise. For example:

    - In `/Dataset/dataPreprocessing`, use "input data" instead of "raw data", since the data to be preprocessed could be data from other previous stages in the pipeline and not necessarily raw data.
    - In `/Dataset/dataCollectionProcess`, use "data" instead of "dataset". This will be more precise to the property name and make it more more naturally applied to a prompt.
        - We may also need to explain what "data collection" means here for prompt. Does it mean how a prompt is constructed? If not relevant, we can remove this field.


bact:
From 8 Oct 2025 AI WG meeting, need a definition of

- [x] prompt
- [ ] context

bact:
Follow-up of the discussion from 8 Oct 2025 AI WG meeting on defining ordinality between two prompts (Prompt1 preceeds Prompt2), a possibility of having a "follows" relationship type is proposed to model a sequence of prompts.

See the "follows" PR: 
- https://github.com/spdx/spdx-3-model/pull/1116

goneall:
@bact @rgopikrishnan91 - It looks like this is still under discussion.  Should this be a draft PR?  Perhaps the AI team could review the open PR's and update with comments and/or close and/or merge.

bact:
Thanks. We will put all AI-related PRs that still under discussion to draft first.

model/AI/Classes/Prompt.md
```diff
@@ -0,0 +1,54 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# Prompt
+
+## Summary
+
+Specifies the content of a prompt and its associated information.
```
ElyasRashno: Specifies the prompt content and its associated information.

model/AI/Classes/Prompt.md
```diff
@@ -0,0 +1,54 @@
...
+For example, referencing a Retrieval-Augmented Generation tool.
+
+## Metadata
+
+- name: Prompt
+- SubclassOf: /Software/SoftwareArtifact
+- Instantiability: Concrete
+
+## Properties
+
```
ElyasRashno, line 43: - /Core/autonomyType
   - type: /Core/PresenceType
   - minCount: 0

model/AI/Classes/Prompt.md
```diff
@@ -0,0 +1,54 @@
...
+  - minCount: 0
+  - maxCount: 1
+- /Dataset/hasSensitivePersonalInformation
+  - type: /Core/PresenceType
+  - minCount: 0
+  - maxCount: 1
+- isContextAugmented
+  - type: xsd:boolean
+  - minCount: 0
+  - maxCount: 1
```
ElyasRashno, line 68: - promptPattern
   - type: xsd:string
   - minCount: 0

ElyasRashno (changes requested):

ElyasRashno:
@bact, I’ve added some comments with change suggestions in the prompt class. I also need access to this PR to create the corresponding vocabulary.

model/AI/Classes/Prompt.md
bact: Added based on summary and description from https://github.com/spdx/spdx-3-model/pull/1082

bact:
> I’ve added some comments with change suggestions in the prompt class. I also need access to this PR to create the corresponding vocabulary.

@ElyasRashno Thank you - incorporated. <s>Will update the diagram.</s> (updated)

GitHub has no facility to share access to an individual PR. Please instead create a PR that targets the branch of this PR (https://github.com/bact/spdx-3-model/tree/add-ai-prompt). Like the example that I have emailed you. Or you can also use the "Add a suggestion" button when commenting on code.

github-add-a-suggestion

model/AI/Classes/Prompt.md
bact: An agent can have autonomy/automation level, as it can take action on its own.
Not sure if a prompt can have the autonomy/automation level, as it is more of a piece of communication between agents (human and machine) ?

bact:
For current definitions of "prompt" and "context", see in the files:
https://github.com/spdx/spdx-3-model/pull/1110/files

> A prompt is the input text, instruction, or data given to an AI model to
initiate and guide its generated output.

> Context is the information provided to the model that shapes or influences
  its generated output.

bact:
From today's Prompt/Context meeting (2025-10-17):

**ONLY FOR MEETING PARTICIPANTS UNDERSTANDING DURING DISCUSSION**

56443822-4230-47A3-8443-45D4CEDE57FA

99D0854D-11BB-4A41-9B10-963D1977D742

713F7297-C704-4955-BA91-BDE98556A545

kestewart:
Please let us know if this won't make 3.1 and should be moved to 3.2

bact:
With the current state of direction within the AI WG, the fundamental design has changed.

We can close this and a new PR with the updated design can be opened later.

PR #1119, opened 2025-10-14
Change to ISO "shall" and "may"

bact:
> For its counterpart PR targeting `support/3.0` branch, see #1134

Using verb forms from https://www.iso.org/sites/directives/current/part2/index.xhtml#_idParaDest-29 from requirements, recommendations, and permissions.

- [x] must/MUST -> shall
- [x] should/SHOULD -> shall (if intended to be requirement) - only done for clear cases (see a comment below)
- [x] may/MAY/might -> may / can (depends on intention)
- To resolve partially https://github.com/spdx/spdx-spec/issues/1236

Note that an example in the description of model/AI/Properties/limitation.md is slightly edited to avoid having both "can" and "cannot" in the same sentence. It does not change the semantic of the property and that part of the text is clearly an example, not a definition.


bact:
The "SHOULD" instances are all changed to "should" in this PR, since they are clear to be RFC "SHOULD" which is a recommendation.

I have done few of "should" but they need more time to review since they involve the decision between requirement (change it to "shall") vs recommendation (keep it as "should").

I would suggest we review this PR first if we can; merge it and have another PR for "should" - that another PR may need more reviews from each profile contributor who know more about that part of the spec.

model/Core/Properties/suppliedBy.md
```diff
@@ -12,7 +12,7 @@ referenced by the Element.
 Identify the actual distribution source for the artifact (e.g., snippet, file,
 package, vulnerability) or VulnAssessmentRelationship being referenced.
 
-This might or might not be different from the originating distribution source
+This can or can not be different from the originating distribution source
```
goneall: This one reads a bit off - should have further review. Is may or may not allowed?  If so, it may read better.

goneall (commented):
Overall - this looks good.  I reviewed all the "shall"s and it looked fine to me.

There was one change to "can" that didn't read quite right to me.  If the change is required or if it reads fine to others, I'm OK with approving and merging the changes.

model/Core/Properties/suppliedBy.md
bact: Thank you.

Yes, it can be either a permission (may) or a possibility (can).

goneall:
```suggestion
This may or may not be different from the originating distribution source
```

goneall (changes requested):
Added a suggestion to change "can" to "may" in one of the files.

model/Core/Properties/suppliedBy.md
bact: Updated as suggested.

zvr (approved):

goneall:
@bact - can you resolve the conflicts - I'll merge once completed

bact:
@goneall conflicts resolved.

goneall (approved):
LGTM

PR #1113, opened 2025-10-04
Add AIAgent as subclass of SoftwareAgent

bact:
Draft of `/AI/AIAgent` which is subclass of `/Core/SoftwareAgent`.

ai-aiagent

Apart from base properties from `/Core/SoftwareAgent`, it will have additional properties as proposed by @ElyasRashno, @stevenc-stb, @cngo20 and others:

- `/Core/suppliedBy` (from Core profile)
- https://github.com/spdx/spdx-3-model/issues/1093#issue-3342194071 -> `/Software/enabledByDefault`
- #1088 -> `/Software/inputModality`
- #1087 -> `/Software/outputModality`
  - Note that the `Modality` vocab is the same as in PR #1110 (copied here to make it passed the parser check)
- #1089 -> `/AI/agentCapability`
- #1084 -> `/AI/agentMemoryMode`
  - Few of the suggested values look a bit overlap with `/AI/promptPattern` here? https://github.com/spdx/spdx-3-model/pull/1110/files
- #1064 -> `/AI/automationLevel`

## Relationship types

Some properties are converted from the original proposal to existing/new relationship types to maximise benefits of existing ecosystem:

- "runs" (from https://lists.spdx.org/g/spdx-ai/message/338) -> `invokedBy`
  - Using `invokedBy` relationship, we can utilised the existing `delegatedTo` ("on behalf of") relationship too
  - `runsOn`, proposed from Hardware profile, can be repurposed / considered
- #1086 -> `usesTool`
- #1079 -> `hasPersistentMemory`
  - The new relationship type could potentially be used in other non-AIAgent context as well.
  - If we feel this one is better to be property, can do that as well - to be discussed
  - Alternatively, we can pick the name `usesLongTermMemoryStore` - to be discussed

```mermaid
graph LR;
  D(AIPackage) -->|invokedBy| B(AIAgent)
  B(AIAgent) -->|invokedBy| C(Person)
  B(AIAgent) -->|hasPersistentMemory| A(Package)
  B(AIAgent) -->|usesTool| E(Package)
```

(Please see a table of existing relationship types at
https://docs.google.com/spreadsheets/d/1g9TophkZ2tWhTPMGGHqGOUurzZvOdkBtBJCvyBsa8kc/edit?usp=sharing )

Note that more properties can be added later, this is just a skeleton.

PR #1125, opened 2025-10-17
Ai agent elyas extension

ElyasRashno:
Added three agent fields on Art PR.


bact:
We use PR https://github.com/bact/spdx-3-model/pull/2 instead.
It was merged to `bact:add-ai-agent` branch, which is used by PR https://github.com/spdx/spdx-3-model/pull/1113.

PR #1129, opened 2025-10-21
ISO edit: Use sentence casing for headings

bact:
To address issue
- https://github.com/spdx/spdx-spec/issues/239

PR #1130, opened 2025-10-21
[3.0] ISO edit: Use sentence casing in headings

bact:
To address issue
- https://github.com/spdx/spdx-spec/issues/239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Change Log
+# Change log
 
 ## 3.0.1 (2024-12-10)
 
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,4 +1,4 @@
-# Contributing to the SPDX 3.0 Model
+# Contributing to the SPDX 3.0 model
 
 ## General
 
@@ -60,7 +60,7 @@ This method of development was agreed upon by the SPDX Tech team on 2023-01-17.
 Translations of model descriptions are welcome.
 Please see [translation.md][translation] for details.
 
-## Profile Maintainers
+## Profile maintainers
 
 In accordance with the development model described above,
 each profile has at least one maintainer in charge of merging profile-specific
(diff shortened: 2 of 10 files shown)


bact:
CI failed because the new heading value does not matched what spec-parser expected
("Property values" vs "Property Values"):

https://github.com/spdx/spec-parser/blob/3550116f10972d2398561b0d178ee7aca0e76e1f/spec_parser/loaders.py#L97-L98

zvr (commented):
I am not in favor of any of these.

There is no need to change anything in the model files ("Profile Conformance", "Property Values", etc.). The generated section names come from templates, not the input.

I also don't see the point in following the ISO restrictions in our own documents (like README, Changelog, Contributing, etc.)

bact:
> There is no need to change anything in the model files ("Profile Conformance", "Property Values", etc.). The generated section names come from templates, not the input.

I am not in favor of applying different rules.
Having one thing that applied consistently makes things easier.

I see not harm of being consistent.

bact:
It is more difficult to tell people that,

"when write for spdx-spec uses sentence casing, but when write for spdx-3-model uses title casing - unless spec-parser will failed"

than

"just use sentence casing everywhere".

zvr:
No, we tell them:

- in text, in the spec repo, use sentence case
- in special formatted files, in the model repo, use one of the predefined headers

goneall:
Moved to 3.1 milestone to close before next release

kestewart:
This is a bit too late for the ISO submission.    Let's discuss getting this consistence going forward for 3.1, unless ISO feedback points it out.  More important right now to get it submitted.

goneall:
@bact @zvr - can we close this and the related #1129 issues since we already submitted to ISO?  If not, should we discuss on one of the tech calls?

zvr:
I am fine with closing this.

goneall:
Closing per @zvr comment

bact:
Thanks. I will reopen this for 3.1 instead.

bact:
Thanks. I will reopen this for 3.1 instead.

PR #1133, opened 2025-10-21
Add definitions for comp hw, comp program, fw, hw, info sys, sw, system

gregshue:
Add definitions for the following terms:

- Computer Hardware
- Computer Program
- Firmware
- Hardware
- Information System
- Software
- System

These definitions were copied or leveraged from

- https://pascal.computer.org/sev_display/index.action,
- https://csrc.nist.gov/glossary, and
- https://datatracker.ietf.org/doc/html/rfc4949:

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -15,7 +15,7 @@ A representation of a scope/set of individual instances of a particular “conce
 
 Each individual instance of a class has an Internationalized Resource Identifier (IRI) and is asserted as a member of a particular class via a type statement.
 
-## Computer Hardware
+## Computer hardware
 
 All or part of the physical components of an information system
 (Leveraged from 'hardware' definition in ISO/IEC 2382:2015 Information technology -- Vocabulary)
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -15,6 +15,16 @@ A representation of a scope/set of individual instances of a particular “conce
 
 Each individual instance of a class has an Internationalized Resource Identifier (IRI) and is asserted as a member of a particular class via a type statement.
 
+## Computer Hardware
+
+All or part of the physical components of an information system
+(Leveraged from 'hardware' definition in ISO/IEC 2382:2015 Information technology -- Vocabulary)
+
+## Computer Program
+
+Combination of computer instructions and data definitions that enable computer hardware to perform computational or control functions
+(From ISO/IEC/IEEE 24765:2017 Systems and software engineering-Vocabulary)
+
 ## Constraint
 
 An explicit conformance requirement specifying details of an SPDX-valid relationship between a property and a concept class including such details as the valid type of the property on that class, the cardinality (e.g., `0..1`, `1`, `1..*`, etc.) of the property on that class, the value of the property on that class, etc.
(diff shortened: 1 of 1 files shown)


docs/glossary.md
```diff
@@ -15,6 +15,16 @@ A representation of a scope/set of individual instances of a particular “conce
 
 Each individual instance of a class has an Internationalized Resource Identifier (IRI) and is asserted as a member of a particular class via a type statement.
 
+## Computer Hardware
```
bact:
```suggestion
## Computer hardware
```

If not a proper noun, we will use a "sentence casing" for heading in this repo and in spdx-spec repo. See https://github.com/spdx/spdx-spec/issues/239

docs/glossary.md
```diff
@@ -15,6 +15,16 @@ A representation of a scope/set of individual instances of a particular “conce
...
+
+All or part of the physical components of an information system
+(Leveraged from 'hardware' definition in ISO/IEC 2382:2015 Information technology -- Vocabulary)
+
+## Computer Program
```
bact:
```suggestion
## Computer program
```

docs/glossary.md
```diff
@@ -41,6 +51,21 @@ A class representing a concept of primary focus within a given domain.
...
+
+Computer programs and data stored in hardware - typically in read-only memory (ROM) or programmable read-only memory (PROM) - such that the programs and data cannot be dynamically written or modified during execution of the programs.
+(From https://csrc.nist.gov/glossary/term/firmware, from CNSSI 4009-2015, **leveraged** from IETF RFC 4949 Ver 2 at https://datatracker.ietf.org/doc/html/rfc4949)
+
+## Hardware
+
+The material physical components of a system.
+(From https://csrc.nist.gov/glossary/term/hardware, from CNSSI 4009-2015)
+
+## Information System
```
bact:
```suggestion
## Information system
```

zvr:
I am not sure what purpose these serve here.

Remember, the "glossary" is about explaining some basic stuff about the model and the structure of the files, to be useful for people who want to contribute to this repo.
It is not published or shown anywhere else.

gregshue:
Oh!

I assumed that Glossary fed into the spdx-spec/docs/terms-and-definitions.md, which is provided in the ISO/IEC to the public an unambiguous meaning for the terms used in the specification. (See ISO/IEC 5962:2021(en), Information technology — SPDX® Specification V2.2.1<https://www.iso.org/obp/ui/en/#iso:std:iso-iec:5962:ed-1:v1:en> section 3 Terms and definitions).

The readers of the spec need clarity on what the authors of this spec meant by each term used in this spec. This is especially true for those making utilities that translate between SPDX and a different format. Where do these definitions need to be captured such that they end up in the ISO/IEC publication of this standard?
________________________________

bact:
For "Terms and definitions" section in the spec, see https://github.com/spdx/spdx-spec/blob/develop/docs/terms-and-definitions.md 

But, yes, it is confusing.

gregshue:
The terms-and-definitions.md file does not contain the definitions presented in Section 3 of the ISO/IEC publishing of SPDX Specification V2.2.1.

I see that the text at the head of Section 3 is contained in this file. Where are the presented definitions captured in the SPDX repositories?

The ISO lookup gives a variety of definitions for "hardware". I did not notice any that covered the scope of "hardware" being discussed in the spdx-hardware special interest group. The closest seemed to be "Hardware Component" from ISO 18202:2015(en), 1.4.2<https://www.iso.org/obp/ui#iso:std:iso:18202:ed-2:v1:en:term:1.4.2>. Neither definition available in the IEC lookup seemed sufficient either. In order for our audience to clearly understand what we mean by "hardware" we need to identify the meaning we are using.

Also, the ISO and IEC lookups give a variety of definitions for "software". ;-)
________________________________

bact:
I think we can propose terms we like to see in the "Terms and definitions" section (one at https://github.com/spdx/spdx-spec/blob/develop/docs/terms-and-definitions.md ), if it does not existed (or used to be existed in previous version of the spec).

Agreed on the point that we can't just say that let's rely on ISO/IEC standand, as they can be more than one definitions of the same term. If necessary, should pick one and document that in the "Terms and definitions" section within the SPDX spec.

I can see this more necessary in SPDX 3, compared to SPDX 2, and look forward to 3.1 where the usages (through profiles) expand beyond software and there can be more chance of "competing" definitions from different domains.

gregshue:
I want to put the definitions in the place where they will automatically be picked up and sorted in the ISO/IEC submission.  Where is the source of the existing definitions published in ISO/IEC 5962:2021(en)?

gregshue:
I found the SPDX v2.2.1 definitions in spdx-spec/chapters/terms-and-definitions.md. That location/version of the file does not reference the ISO or IEC lookups. spdx-spec branch for v3.0 has restructured the content, and replaced all definitions but `package` with links to the ISO and IEC lookups.

I'll create a new PR adding the same proposed definitions to terms-and-definitions.md. It seems like we should now reject or abandon this PR.

bact:
The new PR on spdx-spec repo:

https://github.com/spdx/spdx-spec/pull/1294

PR #1135, opened 2025-10-23
Move RelationshipType naming convention to Markdown guide instead

bact:
- Put info about directionality and temporality to `Relationship` description.
- Move the sentences: *"Relationship names should be descriptive enough to easily deduce the correct direction from their name. The best way to do this is to make sure that the relationship name completes the sentence.."* out of the `RelationshipType` description and put them to the "Naming convention" section in the Markdown guide (docs/format.md) instead.
  - They are not meant for the users of the spec but for the authors of the spec.
  - As the `RELATIONSHIP` is in capital letters, looks like it is from SPDX 2 - should remove it.
- Update an example for property naming convention
  - Replace "import" (singular noun) / "imports" (singular verb) example with less ambiguous "element"/"elements"
  - As "import" / "imports" can be plural verb and plural noun as well, and that can be confusing with the relationship type naming convention.


model/Core/Classes/Relationship.md
```diff
@@ -21,6 +21,14 @@ To explicitly assert that no assertions are being made regarding the
 existence of such relationships, the `to` property shall contain the
 NoAssertionElement individual.
 
+A relationship can be either directional or non-directional (directionless).
+For a directional relationship,
+the direction of a relationship is always from the `from` Element to the
+`to` Element, as defined by the `relationshipType` property.
```
zvr, line 27: Sorry, I do no understand this.

What are these two types of relationships (directional and directionless)? 

And what does it mean that the `relationshipType` property "defines" `from` and `to`?

zvr (changes requested):

model/Core/Classes/Relationship.md
bact: > What are these two types of relationships (directional and directionless)?

This is a directional relationship.

```mermaid
  graph LR;
    A[from] -->|directional relationship| B[to]
```

This is a directionless relationship.

```mermaid
  graph LR;
    A[from] ---|non-directional relationship| B[to]
```

Since v3.0.1 https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/RelationshipType/, we have one directionless relationship type, which is "other":

> other: Every `to` Element is related to the `from` Element where the relationship type is not described by any of the SPDX relationship types (this relationship is directionless).

The rest are directional.

--

> And what does it mean that the relationshipType property "defines" from and to?

The text didn't say that.
It said the `relationshipType` defines the direction of a relationship.
This aligns with the last paragraph of `RelationshipType` description: 

> Relationship names should be descriptive enough to easily deduce the correct direction from their name.

bact: Revised. Please review

goneall (approved):

model/Core/Classes/Relationship.md
```diff
@@ -21,6 +22,14 @@ To explicitly assert that no assertions are being made regarding the
 existence of such relationships, the `to` property shall contain the
 NoAssertionElement individual.
 
+A relationship can be either directional or non-directional (directionless).
+For a directional relationship,
+the direction of a relationship is always from the `from` Element to the
```
stevenc-stb, line 27: i think that the direction of a relationship statement is unnecessary and introduce confusion.   (For a directional relationship) could be misinterpreted as implying a semantic direction that is not correct.  ex types like hasMetadata, hasOptionalComponent, hasAddedFile, hasSpecification, hasTest, and hasRole (used in RoleRelationship)... , where the semantic flow can be the opposite.

stevenc-stb (changes requested):
Clarify that the statement "direction of a relationship" is about relationship direction (structural) vs. semantic direction (meaning). Without that distinction, the comment might be misunderstood.

PR #1136, opened 2025-10-24
Rename `islocatedAt` --> `locatedAt` in RelationshipType

bact:
Follows the pattern of `amendedBy`, `availableFrom`, `invokedBy`, `delegatedTo` ..., omitting "is" prefix for brevity.

This is not a breaking change as the relationship type is new and not released.


stevenc-stb (approved):

goneall (approved):
LGTM

PR #1137, opened 2025-10-24
Fix model.drawio diagram error

umm0:
Fix file error by removing duplicate entries for VEI3-CmWlY-LwqXBLSO8-2 and VEI3-CmWlY-LwqXBLSO8-1

PR #1138, opened 2025-10-24
fix Corrupted Drawio file.

stevenc-stb:
A Fix For a Corrupted and not useable Drawio file.


stevenc-stb:
see #1137  same fix this is not need

PR #1139, opened 2025-10-30
Updated model.Drawio

stevenc-stb:


bact:
If we can have a short summary of changes, it will make the review easier.

stevenc-stb:
> If we can have a short summary of changes, it will make the review easier.
- added the SupportRelationship to model.drawio 
- fixed fonts to Verdana size 10 
- fixed spaces in class rows example " + supportLevel  : SupportType [0..1]" -> "+ supportLevel: SupportType[0..1]"  10+ places
- fixed External properties restrictions  -name: xsd:string[1]> "+ /Core/Element/name: xsd:string[1]"
-  fixed  to show External properties restrictions + /Core/Element/name: xsd:string[1]
- fixed an Core// Element to Core/ Element in build
- change External properties restrictions to blue from red for background colour in SC + hardware .. core updates

stevenc-stb:
-  add + put inLanguage in the right spot in the DatasetPackage.

bact (approved):

bact:
> fixed External properties restrictions -name: xsd:string[1]> "+ /Core/Element/name: xsd:string[1]"

I guessed these are for Package, File, etc. - checked all instances of "name", looks good.

Three minor points on spacing:

In Hardware profile:

Screenshot 2568-10-31 at 20 59 35

the last two properties of Hardware class:

- `additionalInformationSpecification: /Core/Specification [0..*]`
- `additionalInformation: /Core/DictionaryEntry [0..*]`

there should be no space before `[0..*]`

In SupplyChain:

Screenshot 2568-10-31 at 20 53 24

the first property of State class:

- `/Core/Artifact/validUntilTime :DateTime [0]`


- there should be no space before `:`,
- there should be one space after `:`,
- there should be no space before `[0]`

These are more cosmetic and don't affect any semantic (we don't actually have these officially defined anyway) but would be nice to keep them consistent with others.

bact:
Apart from those minor spacing consistency, all LGTM.

Btw, in the future you may also like to export the diagrams to PNGs and put them in https://github.com/spdx/spdx-3-model/tree/develop/images -- so the GitHub can show us the ["image diffs"](https://github.blog/news-insights/behold-image-view-modes/) (GitHub's onion skin image review can be very helpful).

stevenc-stb:
Was replaced by #1145

PR #1143, opened 2025-11-06
Add Decimal section to common validation error

bact:
Add a section on `xsd:decimal` to common errors in SPDX 3 JSON validation document

Follow up from https://github.com/spdx/spdx-spec/pull/1299#issuecomment-3492933941

### DIFF
--- a/serialization/jsonld/validation.md
+++ b/serialization/jsonld/validation.md
@@ -101,6 +101,47 @@ For example, in the SPDX 3.0 Software profile, the `homePage` property uses an
 uppercase "P," while SPDX 2.3 uses the DOAP `homepage` property, which has a
 lowercase "p."
 
+### Decimal
+
+In JSON-LD, always enclose decimal values in quotes.
+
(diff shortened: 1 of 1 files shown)


bact:
@goneall @JPEWdev please see if this make senses.

Probably need to configure Jackson for SPDX Java Library to put quotes around decimal as well?

bact:
I see it now, the JSON-LD store now see it as a double, so no quotes

https://github.com/spdx/spdx-java-v3jsonld-store/blob/b4638a631e9556c73f904e97a35fed2eeb92a78a/src/main/java/org/spdx/v3jsonldstore/JsonLDSchema.java#L76

We may need to update this Canonical serialization doc https://github.com/spdx/spdx-3-model/blob/develop/serialization/canonical.md to cover decimal case.

goneall (approved):
Nice addition - thanks @bact

JPEWdev (approved):

zvr (approved):

PR #1158, opened 2025-11-14
AI/Dataset: Relax property and relationship requirements

bact:
From 2025-10-29 AI Team Meeting, we discussed the possibility to relax property and relationship requirements.

Update: On 2025-11-19 AI Team Meeting, we are agreed to drop these requirements:

1) properties in the "External properties restrictions" of `AIPackage` and `DatasetPackage` classes:
    - `releaseTime`
    - `suppliedBy`
    - `downloadLocation`
    - `packageVersion`; and

2) license relationships `hasConcludedLicense` and `hasDeclaredLicense`, in the "Profile conformance" sections of AI and Dataset profiles. (Sames as Software profile)

We will keep `primaryPurpose` as a mandatory property for both AIPackage and DatasetPackage.
The `primaryPurpose` has a range of `SoftwarePurpose` vocab, which allows "other" as a value.

----

(Old description)

- The general tone from the 2025-10-29 meeting we understand that these information are very useful for different purposes, but they may not always be available.
  - For example, some dataset may not be obtainable thru download.
  - So the question is which one should we keep as mandatory and which one we should made it optional.
- <s>*We are not yet reach the agreement* specifically on each individual property/relationship. This PR is to work that out.</s>

Also note that if we want to require the `AIPackage` or the `DatasetPacakge` to have  `hasConcludedLicense` relationship, we can always ask for the conformance of Licensing profile -- without the need to repeat the same requirement in the AI or Dataset profile.

Since this is a relaxation, it is not a breaking change. An SBOM that conforms SPDX 3.0 will conform SPDX 3.1 too.

@bennetkl @rgopikrishnan91 @cngo20 @stevenc-stb @ElyasRashno

### DIFF
--- a/model/AI/Classes/AIPackage.md
+++ b/model/AI/Classes/AIPackage.md
@@ -8,7 +8,7 @@ A Package that contains AI software or an AI model.
 
 ## Description
 
-A software package that contains code or a model that provides
+A Package that contains code or a model that provides
 artificial intelligence capabilities.
 
 ## Metadata
--- a/model/AI/Classes/AIPackage.md
+++ b/model/AI/Classes/AIPackage.md
@@ -4,11 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Specifies an AI package and its associated information.
+Refers to any unit of content that can be associated with a distribution of
+AI software or AI model.
 
 ## Description
 
-Metadata information that can be added to a package to describe an AI application or trained AI model.
+An AI Package is a specialized type of software Package that includes
+code or model that implements artificial intelligence functionalities.
 
 ## Metadata
 
@@ -74,16 +76,3 @@ Metadata information that can be added to a package to describe an AI applicatio
   - type: /Core/PresenceType
   - minCount: 0
   - maxCount: 1
-
-## External properties restrictions
-
-- /Core/Artifact/releaseTime
-  - minCount: 1
-- /Core/Artifact/suppliedBy
-  - minCount: 1
-- /Software/Package/downloadLocation
-  - minCount: 1
-- /Software/Package/packageVersion
-  - minCount: 1
-- /Software/SoftwareArtifact/primaryPurpose
-  - minCount: 1
--- a/model/AI/Classes/AIPackage.md
+++ b/model/AI/Classes/AIPackage.md
@@ -4,13 +4,12 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Refers to any unit of content that can be associated with a distribution of
-AI software or AI model.
+A Package that contains AI software or an AI model.
 
 ## Description
 
-An AI Package is a specialized type of software Package that includes
-code or model that implements artificial intelligence functionalities.
+A software package that contains code or a model that provides
+artificial intelligence capabilities.
 
 ## Metadata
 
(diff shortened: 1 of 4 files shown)


bact:
From 2025-11-19 AI Team Meeting, we agreed that we will make these properties as optional:

- releaseTime
- suppliedBy
- downloadLocation
- packageVersion

and keep this property as mandatory:

- primaryPurpose (because the SoftwarePurpose enum allows "other")

and we also agreed that we will drop the license relationship requirements in the profile conformance of AI and Dataset profiles.

@bennetkl will bring this to @rgopikrishnan91 and @kestewart to confirm.

model/AI/Classes/AIPackage.md
```diff
@@ -4,11 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Specifies an AI package and its associated information.
+Refers to any unit of content that can be associated with a distribution of
```
stevenc-stb: This is too broad; is it about distribution?  for example, wget is associated with the distribution of an AI model
suggestion

Refers to any unit of content that contains one or more unit of AI software or an AI model.

the suggestion might need some more rework

model/Dataset/Classes/DatasetPackage.md
```diff
@@ -4,11 +4,14 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-Specifies a data package and its associated information.
+Refers to any unit of content that can be associated with a distribution of
```
stevenc-stb: This is too broad, is it about distribution? this more about the dataset not the distribution. 
suggestion
content that contains one or more datasets. 

the suggestion might need some more rework

stevenc-stb (changes requested):
the changes in the summary; in both the AI package and data package made the distribution of it by a tool (for example wget) then look like a AI package and data package.

model/AI/Classes/AIPackage.md
bact: Thank you. 

The sentence is inspired and made parallel to one in the software [`Package`](https://github.com/spdx/spdx-3-model/blob/develop/model/Software/Classes/Package.md):

> Refers to any unit of content that can be associated with a distribution of software.

Anyway, the wget observation is legitimate, so we probably need to update the `Package` summary too.

model/Dataset/Classes/DatasetPackage.md
bact: Same as the AIPackage, the DatasetPackage's summary is a parallel of the software Package. May need to revise all three.

model/AI/Classes/AIPackage.md
bact: Btw, I think the term "distribution" here is not an action but a package.
Same as in "Linux distribution".
But, yes, it is ambiguous and we should avoid it.

bact: Update the summary to: "A Package that contains AI software or an AI model."

We can update the definition of /Software/Package separately.

model/Dataset/Classes/DatasetPackage.md
bact: Update the DatasetPackage summary to : "A Package that contains a dataset."

stevenc-stb (approved):
LGTM to bring to the tech meeting

zvr (approved):

bobmartin3000 (approved):
LGTM

Issue #1149, opened 2025-11-12
use QUDT instead of defining your own units


VladimirAlexiev:
SPDX has made a little foray into units of measure:
```ttl
<https://spdx.org/rdf/3.1/terms/AI/EnergyUnitType/kilowattHour> a owl:NamedIndividual,
        ns7:EnergyUnitType ;
    rdfs:label "kilowattHour" ;
    rdfs:comment "Kilowatt-hour."@en .

<https://spdx.org/rdf/3.1/terms/AI/EnergyUnitType/megajoule> a owl:NamedIndividual,
        ns7:EnergyUnitType ;
    rdfs:label "megajoule" ;
    rdfs:comment "Megajoule."@en .

<https://spdx.org/rdf/3.1/terms/AI/EnergyUnitType/other> a owl:NamedIndividual,
        ns7:EnergyUnitType ;
    rdfs:label "other" ;
    rdfs:comment "Any other units of energy measurement."@en .
```
Please don't do this. Physical quantities and units only SEEMS a simple topic.

Reuse the excellent work done by QUDT starting 15 years ago with NASA funding.
Here you can find tons of information about these units, including conversion factors, physical quantities (quantity kinds), dimension vectors:
- https://qudt.org/vocab/unit/MegaJ
- http://qudt.org/vocab/unit/KiloW-HR

As a minimum, state `skos:exactMatch` between your units and the QUDT units.

Here you can find info about the physical quantities represented by these units (and lists of other commensurate units).
- https://qudt.org/vocab/quantitykind/Energy
- https://qudt.org/vocab/quantitykind/ActiveEnergy: use this one, since this is what your electrical bill pays (there's also Reactive, Apparent, etc etc)

Rather than `ns7:EnergyConsumptionDescription`, you can use https://qudt.org/schema/qudt/QuantityValue with fields `qudt:hasUnit` and `qudt:value`.

PS: Needless to say `EnergyUnitType/other` is utterly useless. It's better not to record any unit in such a case. 
In the semantic web, due to the Open World Assumption, you don't (need to) say "I don't know".

bact:
Thank you. SPDX 3.1 will have QUDT.

See:
- https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Classes/UnitOfMeasure.md
- https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/quantity.md 
[...]

PR #1159, opened 2025-11-14
Update quantity.md -clarification of quantity expectations.

stevenc-stb:
Added more clarification of quantity expectations.
See  #1149

### DIFF
--- a/model/Core/Properties/quantity.md
+++ b/model/Core/Properties/quantity.md
@@ -9,6 +9,7 @@ Quantity is the amount in the selected QUDT unit.
 ## Description
 
 Quantity is a property that defines the count, multitude or magnitude.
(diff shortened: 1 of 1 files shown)


model/Core/Properties/quantity.md
```diff
@@ -9,6 +9,7 @@ Quantity is the amount in the selected QUDT unit.
 ## Description
 
 Quantity is a property that defines the count, multitude or magnitude.
+Express it as an xsd:decimal when possible. If it is not an xsd:decimal, place the notation that is being used in the parent element comment field.
```
bact:
```suggestion

Express it as an xsd:decimal whenever possible.
If it is not an xsd:decimal, place the notation being used in the parent element's comment field.
```

put 's after element, and shorten the second sentence a bit.

bact (approved):

goneall (approved):
LGTM

PR #1189, opened 2026-01-13
Add Markdown lint

bact:
Lint Markdown and make it relatively strict to avoid MkDocs issues. To reduce work of reviewers.

Ignore few rules due to specific use in our model format. For example:

- MD034 - Bare URL used / We need bare URL in metadata section
- MD036 - Emphasis used instead of a header / We use `*Example*` in description
- MD041 - First line in file should be a top level header / Our first line has to be the license line

This PR also fixed few issues reported by Markdownlint, to make the check passes:

- Remove double blank lines in ExternalIdentifierType
- Rearrange position of "release notes" in the change log, to avoid duplicated heading


goneall:
I like adding this in, but lets merge this in after RC1 release.

goneall (approved):
LGTM - Since we've done the RC1 release, once we get one more review we can merge

Ping @zvr @kestewart

bact:
PR open for more than two months. Also non-content. Will merge with one approval.

PR #1201, opened 2026-01-20
ExternalRefType: Add "bom"

### LINKED ISSUE #1191: ExternalRefType entry for SBOMs / SPDX documents
In our context (automotive SBOMs) we are currently using explicit external references to point to SBOM documents that are coming with binary artifacts.
Currently the externalRefType is "other" as there is no match for SBOM documents.

We may consider to have it as a new type in ExternalRefType.

Regards

bact:
To bring to Tech call. Target 3.1-rc2.

To resolve #1191 brought up by @AlexanderDenkBMW

### DIFF
--- a/model/Core/Vocabularies/ExternalRefType.md
+++ b/model/Core/Vocabularies/ExternalRefType.md
@@ -19,6 +19,7 @@ ExternalRefType specifies the type of an external reference.
 - altDownloadLocation: A reference to an alternative download location.
 - altWebPage: A reference to an alternative web page.
 - binaryArtifact: A reference to binary artifacts related to a package.
+- bom: A reference to a bill of materials related to a package.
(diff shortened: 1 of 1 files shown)


bact:
Btw, if we can actually specify each of the BOMs by IDs, may it be more appropriate to use the `externalIdentifier` instead of the `externalRef` ?

AlexanderDenkBMW:
> Btw, if we can actually specify each of the BOMs by IDs, may it be more appropriate to use the `externalIdentifier` instead of the `externalRef` ?

At least for our use case not every BOM will have the characteristic of an externalIdentifier, however at least some could fulfil the requirements.

goneall:
Discussed on 10 Feb 2026:
- If it is an SPDX BOM, a direct  link using the external map is preferred
- If it is a generic BOM where you don't need the internal structure, the an external ref would make sense
- Examples, CDX formatted, non-standard spreadsheets, hardware

goneall (approved):
No objection on 10 Feb tech call - approved

PR #1202, opened 2026-01-20
Fix customIdToLicense: wrong range type in text

bact:
customIdToLicense range is /`Core/ElementMap` not `/Core/DictionaryEntry`.

The text should be 'ElementMap'.

(The paragraph was copied from customIdToUri without modification; customIdToUri uses DictionaryEntry)

@JPEWdev please review.


JPEWdev (approved):

goneall (approved):
Good catch

PR #1213, opened 2026-01-23
Make DateTime pattern more portable

bact:
Make the patterns more defensive/portable, keep the original semantics.

A SHACL processor may rely on platform's regular expression engine and some engine interpret `\d` as any Unicode decimal digit, beyond ASCII `0` - `9` characters, by default.

In those engines, `\d` can match, for example: १ २ ३ ๔ ๕ ๖ ၉

Tested with .NET System.Text.RegularExpressions and PyICU (ICU4C).


goneall (approved):

bact:
@JPEWdev will this relevant to regex in shacl2code?

zvr:
I believe this is a bug in the matching library.  A `\d` in a regex should not match any digit, only the regular `0`-`9`. For example, the superscripts ² and ³ should not be matched.

In Unicode terms, it should not be equivalent to `\p{Digit}`, but actually to `\p{Numeric_Type=Decimal}` property (or `\p{GC=Decimal_Number}` if one wants to use General_Category definitions).

I'm OK with this change, nevertheless.

JPEWdev:
Likely you would need to specify that the regex use whatever the equivalent of the "ASCII-only" flag is (`re.ASCII` in python), but I like this solution better because it doesn't rely on any flags.

JPEWdev (approved):

bact:
Removed SemVer from change set, as it is already covered by PR #1234.

bact:
For .NET, this is intended, not a bug. Noted the "the decimal digits of many other character sets" at the end of the first paragraph. If other behavior is needed, a flag is required.

> \d matches any decimal digit. It's equivalent to the \p{Nd} regular expression pattern, which includes the standard decimal digits 0-9 and the decimal digits of many other character sets.
>
> If ECMAScript-compliant behavior is specified, \d is equivalent to [0-9]. For information on ECMAScript regular expressions, see the "ECMAScript Matching Behavior" section in [Regular Expression Options](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options).

https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-classes-in-regular-expressions#decimal-digit-character-d

Unicode defines "decimal digits" here https://www.unicode.org/terminology/digits.html with example of ٠١٢٣٤٥٦٧٨٩ U+0660..U+0669 code points (see the link for more decimal digits in other scripts).

Screenshot 2569-03-14 at 18 31 54

PR #1215, opened 2026-01-28
Add guidelines for spdxId URI

### LINKED ISSUE #1326: Add recommendation on spdxId and SPDX document namespace
Related to issue #1323 and its PR #1325, also a question about spdxId raised from Go lang implementation team during 5 Jan 2026 Implermenters call, we may like to have an information about suggested spdxId schema/prefix (`https://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]`) and suggested SPDX document namespace (own domain or `https://spdx.org/spdxdocs/`), for SPDX 3.x specs.

See a similar information in SPDX 2.x:
https://spdx.github.io/spdx-spec/v2.3.1-dev/document-creation-information/#65-spdx-document-namespace-field

bact:
Heavily copied from https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#65-spdx-document-namespace-field

To fix https://github.com/spdx/spdx-spec/issues/1326

Previous relevant discussions:
- https://github.com/spdx/spdx-spec/issues/1323
- https://github.com/spdx/spdx-spec/pull/1325

### DIFF
--- a/model/Core/Properties/spdxId.md
+++ b/model/Core/Properties/spdxId.md
@@ -8,10 +8,67 @@ Identifies an Element to be referenced by other Elements.
 
 ## Description
 
-An spdxId uniquely identifies an Element which may thereby be referenced by other Elements.
+An spdxId uniquely identifies an Element which may thereby be referenced by
+other Elements.
 These references may be internal or external.
-While there may be several versions of the same Element, each one needs to be able to be referred to uniquely
+
+While there may be several versions of the same Element,
+each one needs to be able to be referred to uniquely
 so that relationships between Elements can be clearly articulated.
+
+Best practices for creating spdxIds for SPDX documents (and their Elements) on
+the public internet are as follows:
+
+SPDX document:
+
+```text
+https://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]
+```
+
+Element inside an SPDX document:
+
+```text
+https://[CreatorWebsite]/[pathToSpdx]/[DocumentName]-[UUID]#[ElementName]
+```
+
+where:
+
+- `CreatorWebsite` is a website hosted by the creator of the document.
+  (e.g. an SPDX document provided by the SPDX project would be `spdx.org`)
+- `PathToSpdx` is a path to where SPDX documents are stored on the website.
+- `DocumentName` is a name given to the SPDX document,
+  typically the primary package name followed by its version.
+- `UUID` is a universally unique identifier.
+  The UUID could be a random UUID or a name-based version-5 UUID generated from
+  an SHA-1 hash known to be unique for the specific SPDX document version.
+- `ElementName` is a name given to the Element,
+  typically the Element type followed by a sequence number.
+- If the creator does not own their own website, a default `CreatorWebsite` and
+  `PathToSpdx` can be used: `spdx.org/spdxdocs`. Note that the SPDX documents
+  are not currently stored or accessible on this website.
+  The URI is only used to create a unique ID following the above conventions.
+
+The SPDX document spdxId shall not contain a URI "part"
+(e.g. the `#` character), since it may be used in SPDX Element spdxIds
+to separate the document namespace from the element’s SPDX identifier.
+
+*Example*
+
+```text
+https://spdx.org/spdxdocs/spdx-tools-v2.0.4-a0b427e6-9427-414c-9acc-d6e7c3a0d205
+```
+
+The spdxId shall be unique for the SPDX document including the specific
+version of the SPDX document. If the SPDX document is updated, thereby
+creating a new version, a new spdxId for the updated document shall be used.
+There can only be one spdxId for an SPDX document and only one SPDX document
+for a given spdxId.
+
+NOTE:
+While an spdxId is a URI, it does not need to be accessible.
+It is only intended to provide a unique ID.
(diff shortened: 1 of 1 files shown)


bact:
@kzantow @augelu-tng is this SPDX document namespace and element URIs pattern align with what you use currently?

augelu-tng:
In our case we have very large documents. Our goal was therefore to keep the spdxId as short as possible. At the moment we use the following format:
`[Prefix]:[UUID]/[DocumentName]/[SequenceNumber]`
where prefix defaults to `urn:spdx.dev` and DocumentName is one of `source`, `build`, or `output`. 
All three documents describe artifacts within the same kernel build and thus share the same UUID. 
For example:
- `urn:spdx.dev:17fbbbad-b3a2-5a13-9558-00e25f843ee9/source/13`
- `urn:spdx.dev:17fbbbad-b3a2-5a13-9558-00e25f843ee9/build/42`
- `urn:spdx.dev:17fbbbad-b3a2-5a13-9558-00e25f843ee9/output/42`

With a custom JSON-LD context this allows to have very short spdxIds in the document:
```json
{
    "@context": [
        "https://spdx.org/rdf/3.0.1/spdx-context.jsonld",
        {
            "o": "urn:spdx.dev:17fbbbad-b3a2-5a13-9558-00e25f843ee9/output/"
        }
    ],
    "@graph": [
        {
            "spdxId": "o:0",
        }
    ]
}
```

kzantow:
This change seems to align with my current usage. 👍 

Specifically: if the user doesn't explicitly specify a document URI, I'm [following the the 2.3 guidance](https://github.com/spdx/tools-golang/blob/spdx3/spdx/v3/internal/id_generator.go#L18-L30) to generate one, e.g. `https://spdx.org/spdxdocs/<uuid>`. Additionally, I've currently implemented compaction of element URIs [using the namespace map](https://github.com/spdx/tools-golang/blob/spdx3/spdx/v3/v3_0/spdx.go#L155), which I've seen to significantly improve both readability and size of the documents.  In other words, elements other than the `SpdxDocument` have IDs generated with `<prefix>:` followed by `<type-name>-<number>` which would expand to the `<SpdxDocument ID>` + `#` + `<type-name>-<number>` . For example: `SPDXRef:AiPackage-1`, where `SPDXRef` has a namespace map entry for the URI `https://spdx.org/spdxdocs/<uuid>#`. These documents seem to validate properly but because the JSONLD context is not used  but rather the namespace map, the IDs are not automatically expanded.

kzantow:
One note: @augelu-tng using a custom LD context is [not allowed by the spec for SPDX JSON](https://spdx.github.io/spdx-spec/v3.0.1/serializations/#serialization-in-json-ld). I am strongly in favor of leaning in to JSON LD here, as without it extensions have to be in fully expanded form, which is very large and difficult to read, but the spec is specific about using `"@context": "https://spdx.org/rdf/3.0.1/spdx-context.jsonld"`.

bact:
> but the spec is specific about using `"@context": "https://spdx.org/rdf/3.0.1/spdx-context.jsonld"`.

@kzantow we currently have a very long discussion about that at https://github.com/spdx/spdx-spec/issues/1312

augelu-tng:
@kzantow It should be allowed already. The spec says: "Additional namespace mappings may be defined within a separate object within the context.".
However, tooling does not support this currently. 
See also https://github.com/spdx/tools-java/issues/263#issuecomment-3755946182

goneall:
Discussed on 10 Feb tech call - agreed we don't need to tie this to a specific release milestone

goneall (approved):
LGTM - I was thinking we may want to reference the namespaceMap as a way to create shorter ID's in the serialized document.  However, this may over complicate the suggestions.

maxhbr (approved):
Looks good. But I would suggest to have something like `spdx.docs` as a fallback domain suggested for projects that do not want to set a default.

bact:
From @maxhbr :

> Looks good. But I would suggest to have something like `spdx.docs` as a fallback domain suggested for projects that do not want to set a default.

You mean using `spdx.docs` instead of `spdx.org`?

So instead of

```text
https://spdx.org/spdxdocs/spdx-tools-v2.0.4-a0b427e6-9427-414c-9acc-d6e7c3a0d205
```

it will be

```text
https://spdx.docs/spdx-tools-v2.0.4-a0b427e6-9427-414c-9acc-d6e7c3a0d205
```

I like the idea. It will make the ID a bit shorter.
(None of the URLs can be resolved anyway).

From @goneall : 

> LGTM - I was thinking we may want to reference the namespaceMap as a way to create shorter ID's in the serialized document. However, this may over complicate the suggestions.

We can do that. I think in terms of tooling, we are near to support that soon (from what I understand from the Go side and the shacl2code side). Suggesting a shorter ID will greatly improve file size and benefit everyone.

For now, with two approvals, I will merge this PR as it is. And we can improve the spdxId description (suggested domain + shorter ID + @kzantow's suggested `<SpdxDocument ID>` + `#` + `<type-name>-<number>` pattern for Element's spdxId) in another PR.

bact:
`spdx.docs` domain name is taken. Whois information is not available, however.

bact:
I open another issue here https://github.com/spdx/spdx-3-model/issues/1223 to document suggestions being made here and during the 17 Feb 2026 Tech call.

PR #1217, opened 2026-02-10
Generalize startTime, endTime in Action, Build, Project

### LINKED ISSUE #1205: Generalize: Use /Core/startTime and /Core/endTime when possible; instead of creating xxxStartTime
From https://github.com/spdx/spdx-3-model/blob/develop/model/Operations/Classes/Project.md

Does it possible to reuse `/Core/startTime` and `/Core/endTime` for `/Operations/Project` ?

Additional semantic of these two properties in the context of Project can be added to the description of Project.

Note: Maybe this is already in mind of Operations WG as I can see the "startTime" and "endTime" in the Operations model diagram.

bact:
To fix #1205

- Replace projectStartTime, projectEndTime (added in 3.1-RC1) with startTime, endTime
- Replace actionStartTime, actionEndTime (added in 3.1-RC1) with startTime, endTime
- Replace buildStartTime, buildEndTime (since 3.0) with startTime, endTime
  - Deprecated buildStartTime, buildEndTime
- Update startTime, endTime descriptions
- Update change log


goneall (approved):
LGTM - Thanks @bact

stevenc-stb (commented):
Do we need to update the startTime/endTime definition? 

endTime
Summary
Specifies the time from which an element is no longer applicable / valid.
To 
endTime
Summary
Specifies the time from which an element is no longer applicable / valid or has ended. 

and 
startTime
Summary
Specifies the time from which an element is applicable / valid.
to 
startTime
Summary
Specifies the time from which an element is applicable / valid or has started 
That might help it accurately describes the boundary of a temporal object (Action)
Other then that LGTM.

bact:
Thank you @stevenc-stb I have updated the summary/description of startTime/endTime, please review.

(I was putting scheduled time aspects in this commit https://github.com/spdx/spdx-3-model/pull/1217/changes/fc65dba5d7181987f978a95842b5e87cd07dffd1, but then removed it since I think may be it's better to keep it more simple and leave that scheduled time aspects in the /Operations/Project description instead).

stevenc-stb (approved):
LGTM

PR #1224, opened 2026-02-26
Adding certificate, key, and signature to security profile 

stevenc-stb:
Adding certificate,key and signature to security profile. This is for review.


model/Security/Properties/notValidAfter.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# notValidAfter
```
bact, line 3: Does it possible to use [`/Core/validUntilTime`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/validUntilTime.md) (verbatim or with revision)?

`validUntilTime` naming is also aligned with the Schema.org [`validUntil`](https://schema.org/validUntil) and avoids negation in the name.

model/Security/Properties/notValidBefore.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# notValidBefore
```
bact, line 3: We don't have an equivalent of this yet. So it makes sense to create new one.
But still concerning about negation in the name.

Schema.org uses [`validFrom`](https://schema.org/validFrom), which pairs with [`validUntil`](https://schema.org/validUtil).

As we already have [`/Core/validUntilTime`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/validUntilTime.md), would it be possible to create `/Core/validFromTime` and use it instead?

model/Security/Properties/oid.md
```diff
@@ -0,0 +1,18 @@
...
+# oid 
+
+## Summary
+
+The Object Identifier (OID) that uniquely identifies the extension type.
+
+## Description
+
+This property holds the registered Object Identifier (OID) string that uniquely identifies the type or purpose of a certificate extension. OIDs are hierarchical numeric identifiers (e.g., '2.5.29.15' for Key Usage) assigned by standardization bodies or organizations. 
+For standard X.509 extensions, this OID corresponds to a well-known value; for custom extensions, it is a unique identifier registered within an organization's OID arc. This property enables precise, machine-readable identification of the extension beyond any human-readable name.
```
bact, line 13:
```suggestion

For standard X.509 extensions, this OID corresponds to a well-known value; for custom extensions, it is a unique identifier registered within an organization's OID arc. This property enables precise, machine-readable identification of the extension beyond any human-readable name.
```

The part before this is generic description of the OID.
This part is about specific context (of X.509). So it would help reader if we separate them as different paragraphs. In Markdown, we can do this with a blank line.

model/Security/Properties/parameter.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# parameter
```
bact, line 3: We can also consider move this to Core, make it a generic `/Core/parameter`.

We can then use the `/Core/parameter` here, for cryptographic algorithm documentation.

And we can also deprecate [`/Build/parameter`](https://github.com/spdx/spdx-3-model/blob/develop/model/Build/Properties/parameter.md), let the `Build` use this new `/Core/parameter` instead.

model/Security/Properties/serialNumber.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# serialNumber
```
bact, line 3: Would it be possible to share this with Hardware use case?

So instead of having `/Security/serialNumber` and [`/Hardware/serialNumber`](https://github.com/spdx/spdx-3-model/blob/develop/model/Hardware/Properties/serialNumber.md), we can have one `/Core/serialNumber` that can be used in different context.

model/Security/Properties/signatureFormat.md
```diff
@@ -0,0 +1,25 @@
...
+
+The signatureFormat property specifies the signature container format used to encode the signature data. This identifies the structure and encoding of how the signature is packaged, enabling appropriate parsing and verification tools to be selected.
+
+Common signature formats include:
+
+    PGP - Pretty Good Privacy format, commonly used for OpenPGP RFC 4880 signatures with support for radix64 encoding
+    CMS/PKCS#7 - Cryptographic Message Syntax format, used for S/MIME and code signing with ASN.1 DER encoding
+    JSON Web Signature (JWS) - A compact, URL-safe format for signing JSON data, used in modern web APIs
+    Detached signature - A format where the signature exists separately from signed content, common in package managers like RPM and apt
+    XMLDSig - Digital signatures for XML documents, used in web services and document signing
```
bact:
```suggestion
- PGP - Pretty Good Privacy format, commonly used for OpenPGP RFC 4880 signatures with support for radix64 encoding
- CMS/PKCS#7 - Cryptographic Message Syntax format, used for S/MIME and code signing with ASN.1 DER encoding
- JSON Web Signature (JWS) - A compact, URL-safe format for signing JSON data, used in modern web APIs
- Detached signature - A format where the signature exists separately from signed content, common in package managers like RPM and apt
- XMLDSig - Digital signatures for XML documents, used in web services and document signing
```

model/Security/Properties/signatureTime.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# signatureTime
```
bact, line 3: Can we use [`/Core/created`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/created.md) here?

model/Security/Properties/sourceElement.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# sourceElement
```
bact: The filename has a typo. (extra `I` before `Element`)

stevenc-stb: Rename sourceIElement.md to sourceElement.md  fixed

PR #1220, opened 2026-02-11
 Update diagram with the generalized properties

bact:
Based on changes in these approved and merged PRs:

- https://github.com/spdx/spdx-3-model/pull/1217
- https://github.com/spdx/spdx-3-model/pull/1218
- https://github.com/spdx/spdx-3-model/pull/1219

PR #1227, opened 2026-02-27
Profile operations mk202602

mkurzman:
This branch was created from develop as of today.
The draw.io file was extended by the OpenSourceProject Class and merely the md-file for OpenSourceProject was added in the model section as base for next review round.
This is merely a PR as base for common review in the operations profile group.
The missing md-files for properties etc. still need to be added in the next iteration after the high-level discussion on Class-level have been done.

### DIFF
--- a/docs/model.drawio
+++ b/docs/model.drawio
@@ -1,2615 +1,2688 @@
-<mxfile host="app.diagrams.net" compressed="false" agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:147.0) Gecko/20100101 Firefox/147.0" version="29.3.4" pages="13">
-  <diagram id="MqBy92FNWJOg9jOokCTN" name="Core">
-    <mxGraphModel dx="2191" dy="1967" grid="1" gridSize="10" guides="1" tooltips="1" connect="1" arrows="1" fold="0" page="1" pageScale="1" pageWidth="827" pageHeight="1169" math="0" shadow="0">
-      <root>
-        <mxCell id="0" />
-        <mxCell id="1" parent="0" />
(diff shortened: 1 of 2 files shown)


bact:
I would suggest to leave draw.io diagram unchanged in this PR.

And once everything is settled, update the diagram in another PR.

Or there will be a difficult to manage conflicts over the diagram file.

We now have one PR pending for diagrams changes
https://github.com/spdx/spdx-3-model/pull/1220

bact:
Looking forward, we may like to put one diagram per one file in the future, so it will be easier to manage.

Or rely on "text based" descriptive diagram (like PlantUML and Mermaid), which is more friendly to text version control system.

zvr:
+1 for having different `.drawio` files, one per profile. That's how they are being used, anyway: each one in a separate chapter in annex A.

FYI, spec-parser already produces a complete "text-based" diagram, in PlantUML format (command line switches `-P` or `--output-plantuml`, IIRC). But manually creating/manipulating in diagrams.net results in a more aesthetically appealing diagram.

umm0:
Just FYI, I needed to switch the base branch back and forth to update the file diff view to show the correct state after https://github.com/spdx/spdx-3-model/pull/1228 .
Github apparently does not do this automatically.

bact:
FYI #1220 (that modifies diagram) has just merged (thanks all the reviewers). So it should be good now to edit the diagram file -- please merge from develop again one more time.

mkurzman:
I started a new PR based on latest "develop" that only contains a dedicated drawio-file for profile operations as suggested: https://github.com/spdx/spdx-3-model/pull/1269
With that I think I can close this PR.

PR #1234, opened 2026-03-10
Relax patch version requirement for SPDX Versions

### LINKED ISSUE #1233: Version 3.1 generated model.ttl file creates invalid creation info spec versions
Individuals included in the SPDX 3.1 model.ttl file have a `ns1:specVersion` value of `"3.1"` which does not match the pattern prescribed for the property value.  The property value is of data type `SemVer` which has a `sh:pattern "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$" ]`

Note the required patch version.

There is a related issue where it isn't easy to tell the patch version of the spec for a given model.ttl file.

Note also that the patch version is required for the [SemVer spec](https://semver.org/).

goneall:
Fixes #1233

Note that there were no references to the SemVer datatype in the model.ttl file, so we may be safe in renaming data types.


model/Core/Datatypes/SemVer.md
```diff
@@ -18,4 +20,4 @@ A semantic version is a string that is following the specification of
 
 ## Format
 
-- pattern: ^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$
+- pattern: ^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(\\.(0|[1-9]\\d*))?(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$
```
bact: Should we do the escape in the spec or the escape should be the responsibility of the implementation?

Btw, this PR https://github.com/spdx/spdx-3-model/pull/1213 is proposing to use specific `[0-9]` instead of the `\d` (any Unicode digit)

goneall: Good catch - I copied the string from the model file which included the escapes.

I just pushed a commit to remove the escapes.

I can also include PR #1213 as well or we could do a separate PR.

goneall: @zvr @JPEWdev @kestewart - If I could get one more "thumbs up" on PR #1213, I'll include the change in this PR to avoid any merge conflicts.

zvr:
Having thought a little more about this, my current opinion is that, since we only use this Dataclass for the versions of the specification and of the license list, we should:

- probably rename it to something more generic, like (Version)
- say that it's the same as SemVer, to avoid having a backwards incompatibility (but deprecate the use of SemVer)
- simply the allowed regexp, only two or three numbers (i.e., do not allow things like "1.0.0-alpha.beta")

goneall:
PR updated to replace the pattern `\d` with `[0..9]` per PR #1213 

Note - this change was only made to the SemVer pattern.

goneall:
> Having thought a little more about this, my current opinion is that, since we only use this Dataclass for the versions of the specification and of the license list, we should:
> 
> * probably rename it to something more generic, like (Version)

Agree

> * say that it's the same as SemVer, to avoid having a backwards incompatibility (but deprecate the use of SemVer)

I think we can safely just rename it since it isn't referenced in any of the published model or context files - it looks like just the pattern is used directly.  Please double check me on this

> * simply the allowed regexp, only two or three numbers (i.e., do not allow things like "1.0.0-alpha.beta")

My initial reaction was that this would be incompatible - but since all version strings are under the control of SPDX (the spec and license list), I think this would be fine.

If there are no objections, I can update the PR to:
- rename SemVer to Version in core
- Change the regex to: `^(0|[1-9][0..9]*)\.(0|[1-9][0..9]*)(\.(0|[1-9][0..9]*))?$`

zvr:
> I think we can safely just rename it since it isn't referenced in any of the published model or context files - it looks like just the pattern is used directly. Please double check me on this

Yes, Datatypes are not referenced anywhere (besides these pages).

But the 3.0.1 spec that will result in an ISO does include it, so we should at least mention something in the text ("This was previously named SemVer" or similar).

On the format, I'd prefer to have the possibility to have three parts (the third optional), so `x\.y(\.z)?` with each of the `x`, `y`, and `z` the same `(0|[1-9][0-9]*)`. We went to a lot of trouble making sure License List versions are now in three parts, let's not undo it.

goneall:
Updated PR with renamed SemVer and removing the extensions.

model/Core/Datatypes/Version.md
```diff
@@ -0,0 +1,24 @@
...
+Note that this Datatype was named "SemVer" prior to the SPDX Specification release 3.1.
+
+## Metadata
+
+- name: Version
+- SubclassOf: xsd:string
+
+## Format
+
+- pattern: ^(0|[1-9][0..9]*)\.(0|[1-9][0..9]*)(\.(0|[1-9][0..9]*))?$
```
zvr:
```suggestion
- pattern: ^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(\.(0|[1-9][0-9]*))?$
```

zvr: Ranges in regex are marked with `-`, not `..`

zvr (changes requested):

model/Core/Datatypes/Version.md
```diff
@@ -0,0 +1,24 @@
...
+## Summary
+
+A string following the SemVer 2.0.0 specification with the exceptions of the patch version being optional and extensions are not supported.
+
+## Description
+
+A semantic version is a string that is following the specification of
+[Semantic Versioning 2.0.0](https://semver.org/).
+The SPDX SemVer type differs from Semantic Versioning 2.0.0 in that the patch version is optional and extensions are not supported.
+If no patch version is present, the latest stable patch version can be assumed.
```
zvr:
```suggestion
If no patch version is present, the latest released patch version can be assumed.
```

We do not have "stable" or "non-stable" versions; we have either released something or not.

zvr (changes requested):

bact (approved):

model/Core/Datatypes/Version.md
```diff
@@ -0,0 +1,24 @@
...
+
+A string following the SemVer 2.0.0 specification with the exceptions of the patch version being optional and extensions are not supported.
+
+## Description
+
+A semantic version is a string that is following the specification of
+[Semantic Versioning 2.0.0](https://semver.org/).
+The SPDX SemVer type differs from Semantic Versioning 2.0.0 in that the patch version is optional and extensions are not supported.
+If no patch version is present, the latest released patch version can be assumed.
+Note that this Datatype was named "SemVer" prior to the SPDX Specification release 3.1.
```
bact:
```suggestion
A version is a string of decimal numbers and dots assigned to a specific state
of software, a specification, a document, or an artifact.

The SPDX Version type follows [Semantic Versioning 2.0.0](https://semver.org/)
but differs in that the patch version is optional and
extensions are not supported.

If no patch version is present, the latest released patch version is assumed.

Note that this Datatype was named "SemVer" prior to the SPDX Specification release 3.1.
```

- Remove "semantic" from first line of description.
- Give generic description of version.
- Replace "SPDX SemVer" with "SDPX Version".
- "latest released patch version can be assumed" -> "latest released patch version is assumed", if it is to be assumed, we may like to use "is" to avoid the complication of "can" meaning in ISO ("can" in ISO is for possibility / capability)

bact (changes requested):

model/Core/Datatypes/Version.md
goneall: LGTM - I'll accept the change

bact (approved):

model/Core/Datatypes/Version.md
```diff
@@ -0,0 +1,29 @@
...
+A version is a string of decimal numbers and dots assigned to a specific state
+of software, a specification, a document, or an artifact.
+
+The SPDX Version type follows [Semantic Versioning 2.0.0](https://semver.org/)
+but differs in that the patch version is optional and
+extensions are not supported.
+
+If no patch version is present, the latest released patch version is assumed.
+
+Note that this Datatype was named "SemVer" prior to the SPDX Specification release 3.1.
```
bact:
```suggestion
Note that this Datatype was named "SemVer" prior to the SPDX specification release 3.1.
```

bact (changes requested):

bact (approved):

model/Core/Properties/specVersion.md
```diff
@@ -18,8 +18,9 @@ The major version number shall be incremented when incompatible changes between
 versions are made (one or more sections are created, modified or deleted).
 The minor version number shall be incremented when backwards compatible changes
 are made.
-The patch version number shall be incremented when backward compatible bug
+The optional patch version number shall be incremented when backward compatible bug
 fixes are made.
+If no patch version is present, the latest stable patch version can be assumed.
```
zvr:
```suggestion
If no patch version is present, the latest released patch version can be assumed.
```
We do not have "stable" or "non-stable" versions; we have either released something or not.

zvr (changes requested):

zvr (approved):

bact:
We may need to update the information about `specVersion` in the SPDX Lite annex in the spec:

https://github.com/spdx/spdx-spec/blob/develop/docs/annexes/spdx-lite.md#corecreationinfo

PR #1240, opened 2026-03-24
Clarify ElementMap key must be case insensitive

### LINKED ISSUE #1363: Mismatching guidance on LicenseRef- handling
In the current [text for customIdToLicense](https://github.com/spdx/spdx-3-model/blob/develop/model/SimpleLicensing/Properties/customIdToLicense.md), it says:

> Maps a "LicenseRef-" string for a custom license or a "AdditionRef-" string for a custom license addition to a CustomLicense, a CustomLicenseAddition, or a SimpleLicensingText.
...
> The key for the ElementMap is the string used in the license expression and the value is target Element, which must be a CustomLicense, CustomLicenseAddition, or SimpleLicensingText.

And in [the License Expression annex section on case sensitivity](https://spdx.github.io/spdx-spec/v3.1-dev/annexes/spdx-license-expressions/#case-sensitivity) it states:

> In SPDX 3, license expressions are completely case-insensitive.
> 
> That includes the operators ("AND", "OR", "WITH"), the special identifiers ("NONE" and "NOASSERTION"), as well as the license identifiers, including the user-defined ones.
> 
> For example, the expressions MIT AND NOASSERTION AND (BSD-3-Clause OR LicenseRef-Name) and mit aNd NoaSSerTion AnD (bSd-3-clausE OR licenseref-NAME) are equivalent.

It also refers to the `LicenseRef-...` (and `AdditionRef-`) using the form: `["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)`, which, I think was copied from 2.3 and seems to no longer be valid, since these are no longer URIs, instead just references to the map that links to a URI or an element. A recent discussion resulted in the conclusion that in 2.3 the `"LicenseRef-"(idstring)` should be case _sensitive_ because it's directly referring to a URI but that is not the case with a map indirection in 3.x. Additionally, this seems to be the first use of `idstring`, which in 2.3 is always a _case sensitive_ URI, but here the term is used without any other defining information, making it hard to understand the intent aside from typically IDs are case _sensitive_.

These seem to be conflicting statements and it's unclear what string should be used in the map lookup -- the `customIdToLicense` does not indicate that the lookup is case _insensitive_, in fact sounds the opposite and does not specify whether it should include the `LicenseRef-` portion at all (the text in the `customIdToLicense` makes me think it _should_ be included).

Clarification on the correct behavior here would be great.

goneall:
Fixes https://github.com/spdx/spdx-spec/issues/1363

### DIFF
--- a/model/SimpleLicensing/Properties/customIdToLicense.md
+++ b/model/SimpleLicensing/Properties/customIdToLicense.md
@@ -21,8 +21,11 @@ the official [SPDX License List](https://spdx.org/licenses/).
 The key for the ElementMap is the string used in the license expression
 and the elementValue is target Element, which must be a CustomLicense,
 CustomLicenseAddition, or SimpleLicensingText.
-The key should be matched to the license expression in a case insensitive manner 
-since the entire license expression string is case insensitive.
+
+The key for the ElementMap shall be matched against the license expression
+in a case-insensitive manner.
+This is required because the entire license expression string is itself
+case-insensitive.
 
 ## Metadata
 
--- a/model/SimpleLicensing/Properties/customIdToLicense.md
+++ b/model/SimpleLicensing/Properties/customIdToLicense.md
@@ -19,8 +19,10 @@ dictates any reference starting with a
 the official [SPDX License List](https://spdx.org/licenses/).
 
 The key for the ElementMap is the string used in the license expression
-and the value is target Element, which must be a CustomLicense,
+and the elementValue is target Element, which must be a CustomLicense,
 CustomLicenseAddition, or SimpleLicensingText.
+The key should be matched to the license expression in a case insensitive manner 
+since the entire license expression string is case insensitive.
 
 ## Metadata
 


model/SimpleLicensing/Properties/customIdToLicense.md
```diff
@@ -19,8 +19,10 @@ dictates any reference starting with a
 the official [SPDX License List](https://spdx.org/licenses/).
 
 The key for the ElementMap is the string used in the license expression
-and the value is target Element, which must be a CustomLicense,
+and the elementValue is target Element, which must be a CustomLicense,
 CustomLicenseAddition, or SimpleLicensingText.
+The key should be matched to the license expression in a case insensitive manner 
+since the entire license expression string is case insensitive.
```
bact:
```suggestion

The key for the ElementMap shall be matched against the license expression
in a case-insensitive manner.
This is required because the entire license expression string is itself
case-insensitive.
```

- Use "shall" (ISO requirement).
- Break the long sentence to two: one for requirement, another for rationale
- Add new line to put this requirement into another paragraph, separate it from the definition paragraph ("The key for the ElementMap is ...")

bact (approved):

kestewart (approved):
Matches discussion in Tech call.

PR #1245, opened 2026-04-07
Update DateTime pattern to follows ISO 8601:2019

bact:
Update the string pattern of [`DateTime`](https://github.com/bact/spdx-3-model/blob/develop/model/Core/Datatypes/DateTime.md). The new pattern will restrict:

- month to 01-12
- date to 01-31
- hour to 00-23
- minute to 00-59
- second to 00-60

These restrictions are aligned with ISO 8601:2019, which the `DateTime` is referenced to.

Note that the "60" second is reserved only for the positive leap second at 23:59:60. See ISO 8601:2019 Clause 3.1.1.24 https://www.iso.org/obp/ui/#iso:std:iso:8601:-1:ed-1:v1:en and RFC 3339 Appendix D https://www.rfc-editor.org/rfc/rfc3339.html#appendix-D

Based on changes in SPDX 2.3.1 draft  https://github.com/spdx/spdx-spec/pull/1377


model/Core/Datatypes/DateTime.md
```diff
@@ -21,4 +21,4 @@ The specific format is one of the most commonly used ISO-8601 formats.
 
 ## Format
 
-- pattern: ^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}Z$
+- pattern: ^[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|30|31)T([0-1][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60)Z$
```
zvr: For years, you can use `(19|20)[0-9][0-9]` since for all practical purposes we only deal with 20th and 21st century.
The day part can be expressed as `(0[1-9]|[12][0-9]|3[01])` (no need for range `1-2` and you can group `30` and `31`.
Similarly, hour is `([01][0-9]|2[0-3])`.
And there is no need for parentheses around minutes, there is no `|` in there.

```
    ((19|20)[0-9][0-9])
-
    (0[1-9]|1[0-2])
-
    (0[1-9]|[12][0-9]|3[01])
T
    ([01][0-9]|2[0-3])
:
    [0-5][0-9]
:
    ([0-5][0-9]|60)
```

We can go to even more lengths and make sure there is no April 31st, etc. or to even more and allow only the correct dates for February, but I don't think it's useful...

And the datatypes are not currently shown in the schema, anyway.

zvr (changes requested):

model/Core/Datatypes/DateTime.md
bact: Thank you. Explicit ranges and parentheses removed.

stevenc-stb (commented):
as a first look at it look fine. is SPDX going to support fractional seconds?

bact:
> as a first look at it look fine. is SPDX going to support fractional seconds?

Not that I know of. This PR is to allowed the recording of leap second, in case there is any, to aligned with ISO 8601. But the precision is still at the second level, not subseconds.

bact:
All change requests resolved.
Will merge with one approval after a month of no further request for changes.
Note that this changes mirror changes in the already merged https://github.com/spdx/spdx-spec/pull/1377

Thanks @zbleness again for the original PR.

Issue #1247, opened 2026-04-08
ai-prompt-proposal


ElyasRashno:
This issue is for the Prompt class under the AI profile. This version is a combination of the previous issue, community discussions, and Raymond's definition document.
PR #1249

ElyasRashno:
SPDX-License-Identifier: Community-Spec-1.0

# `add-ai-prompt-v2` Branch

## New Class: `AI/Prompt`



Represents a prompt — instructions and directly accompanying content (natural language text, documents, code, and/or media) provided to an AI system by an external entity (human or AI agent) to guide its output.

Prompts are categorized by their role:
- A **user prompt** is provided by a user (human or AI agent) and is associated with a single query.
- A **system prompt** is provided by the system's owner or administrator to inform the system's overall behavior for every query or instruction. In agentic contexts, system prompts are sometimes referred to as "instructions."

## Properties

- /Core/inLanguage
  - type: /Core/LanguageTag
  - minCount: 0
- /Core/contentModality
  - type: /Core/Modality
  - minCount: 0
- /Core/contentType
  - type: /Core/MediaType
  - minCount: 0
- /Dataset/dataCollectionProcess
  - type: xsd:string
  - minCount: 0
  - maxCount: 1
- /Dataset/dataPreprocessing
  - type: xsd:string
  - minCount: 0
  - maxCount: 1
- /Dataset/hasSensitivePersonalInformation
  - type: /Core/PresenceType
  - minCount: 0
  - maxCount: 1
- /Dataset/confidentialityLevel
  - type: /Dataset/ConfidentialityLevelType
  - minCount: 0
  - maxCount: 1
- isContextAugmented
  - type: xsd:boolean
  - minCount: 0
  - maxCount: 1
- promptPattern
  - type: xsd:string
  - minCount: 0
- promptRole
  - type: /AI/PromptRoleType
  - minCount: 0
  - maxCount: 1



---

## New Properties (AI Profile)

### `promptRole`
- **Nature:** ObjectProperty
- **Range:** `PromptRoleType`
- **Cardinality:** 0..1

Specifies the role of a prompt within an AI interaction. Prompts are categorized into two distinct roles:
- `user`: Provided by a user (human or AI agent), associated with a single query or instruction.
- `system`: Provided by the system's owner, maintainer, or administrator to inform the system's overall behavior for every query or instruction. Sometimes referred to as "instructions" in agentic contexts.

---

### `promptPattern`
- **Nature:** DataProperty
[...]

PR #1248, opened 2026-04-08
Add ai prompt v2

ElyasRashno:
This PR introduces newly proposed fields for the prompt class in the AI profile. Here is the related issue number, along with the discussions from the AI & Dataset group meeting.
https://github.com/spdx/spdx-3-model/issues/1247


bact:
Use https://github.com/spdx/spdx-3-model/pull/1249 instead

PR #1249, opened 2026-04-08
Add ai prompt v2

ElyasRashno:
This PR introduces newly proposed fields for the prompt class in the AI profile. Here is the related issue number, along with the discussions from the AI & Dataset group meeting.
https://github.com/spdx/spdx-3-model/issues/1247


model/AI/Properties/promptRole.md
```diff
@@ -0,0 +1,31 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# promptRole
```
bact, line 3: From 8 Apr 2026 SPDX AI WG call, the property name is update pending. The group is still to decide so Elyas can update this accordingly.

Possible alternatives are:

- promptKind
- promptSource
- promptLevel
- promptClass
- promptCategory

or it can be others.

model/AI/Vocabularies/PromptRoleType.md
```diff
@@ -0,0 +1,31 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# PromptRoleType
```
bact, line 3: When possible, avoid using "-Type" suffix for term that already indicate class or category.

In this case, "Role" already implies a class/category, so it is unnecessary to repeat the notion again with "Type".

No need to change anything here btw, since we will rename the `promptRole`.

.gitignore
```diff
@@ -2,3 +2,5 @@
 *.bkp
 *.dtmp
 *.swp
+.github/
+trash/
```
bact:
```suggestion

```
We need .github for repo workflows to work correctly.

Please undo the entire .gitignore changes.

bact (changes requested):

model/AI/Properties/isContextAugmented.md
```diff
@@ -0,0 +1,32 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# isContextAugmented
```
bact, line 3: isContextAugmented is bad name.

It can have two reads:

1) Is context augmented?
2) Is it augmented by context?

I think about intention is (2).

May need to rename

stevenc-stb, line 3:
```suggestion
# isPromptAugmented
```

stevenc-stb (changes requested):
the Property of isContextAugmented -> isPromptAugmented make it clearer, see meeting notes on 2026 08 19

PR #1254, opened 2026-04-14
Add symlink entry to FileKindType vocabulary

goneall:
Proposed solution for https://github.com/spdx/spdx-spec/issues/610

### DIFF
--- a/model/Software/Vocabularies/FileKindType.md
+++ b/model/Software/Vocabularies/FileKindType.md
@@ -8,7 +8,7 @@ Enumeration of the different kinds of SPDX file.
 
 ## Description
 
-An SPDX file may represent a file on disk or a directory of files.
+An SPDX File can represent a regular file, a directory of zero or more Files, or a symbolic link to a File.
 
 In the future, this may be extended to other kinds (e.g. network based files).
 
@@ -20,4 +20,4 @@ In the future, this may be extended to other kinds (e.g. network based files).
 
 - file: The file represents a single file (default).
 - directory: The file represents a directory and all content stored in that directory.
-- symlink: This file represents a symbolic link.
+- symlink: This file represents a symbolic link (soft link).
--- a/model/Software/Vocabularies/FileKindType.md
+++ b/model/Software/Vocabularies/FileKindType.md
@@ -10,7 +10,7 @@ Enumeration of the different kinds of SPDX file.
 
 An SPDX File can represent a regular file, a directory of zero or more Files, or a symbolic link to a File.
 
-In the future, this may be extended to other kinds (e.g. network based files).
+In the future, this can be extended to other kinds (e.g. network based files).
 
 ## Metadata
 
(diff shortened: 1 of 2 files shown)


model/Software/Vocabularies/FileKindType.md
```diff
@@ -20,3 +20,4 @@ In the future, this may be extended to other kinds (e.g. network based files).
 
 - file: The file represents a single file (default).
 - directory: The file represents a directory and all content stored in that directory.
+- symlink: This file represents a symbolic link.
```
bact:
```suggestion
- symlink: This file represents a symbolic link (soft link).
```

Not sure if necessary but may be make it more explicit (and emphasize that this is not a hard link, as mentioned here https://github.com/spdx/spdx-spec/issues/610#issuecomment-4247124532)

model/Software/Vocabularies/FileKindType.md
bact:
```suggestion
An SPDX file can represent a regular file on disk, a directory of files, or a symbolic link.
```

- Add the new symbolic link.
- Add "regular" quanlifier to "file on disk", since now we have the symlink, which is also "file on disk" (at least for Unix).
- Use "can" for capability. "May" is for permission. See: https://github.com/spdx/spdx-3-model/blob/develop/docs/format.md#verbal-forms-for-expressions-of-provisions

(btw, not sure if "directory of files" implies that the directory must contain at least one file or it can be an empty directory)

zvr: `An SPDX File can represent a regular file, a directory of zero or more Files, or a symbolic link to a File.`

- delete the "on disk" (there are memory-based filesystems)
- use capital F for File to refer to the SPDX notion

zvr (changes requested):

model/Software/Vocabularies/FileKindType.md
goneall: Updated per reccomendation

model/Software/Vocabularies/FileKindType.md
goneall: Updated with @zvr reccomendation

model/Software/Vocabularies/FileKindType.md
```diff
@@ -8,7 +8,7 @@ Enumeration of the different kinds of SPDX file.
 
 ## Description
 
-An SPDX file may represent a file on disk or a directory of files.
+An SPDX File can represent a regular file, a directory of zero or more Files, or a symbolic link to a File.
 
 In the future, this may be extended to other kinds (e.g. network based files).
```
bact: "In the future, this can be extended ..."


In this context, it is a capability/possibility, so use "can" per ISO verbal form.

goneall: Updated per suggestion

bact (approved):

zvr (approved):

bact (approved):

Issue #1256, opened 2026-04-17
ai-agent-proposal


ElyasRashno:
This issue is for the AIAgent class under the AI profile. This version is a combination of the previous issue and community discussions.
https://github.com/spdx/spdx-3-model/pull/1257/

ElyasRashno:
@gongli0929, this issue is for discussing newly proposed fields for AI Agent class in AI profile, which we have been continuously reviewing in the AI and Dataset Profile Working Group. I will link the correspond PR to this issue.

ElyasRashno:
SPDX-License-Identifier: Community-Spec-1.0

# `add-ai-agent-v2` Branch

## New Class: `AI/AIAgent`

**Subclass of:** `/Core/SoftwareAgent`

Represents an AI agent, a software entity that uses artificial intelligence techniques to perceive its environment, reason about its state, and take actions to achieve specified goals. It typically implements a perception-action cycle employing AI models, heuristics, or planning algorithms to guide autonomous or semi-autonomous behavior.

AI agents may be characterized by the modalities through which they perceive input and produce output, the capabilities they provide, the external tools or services they invoke, the memory mechanisms they employ to retain information across interactions, and their degree of automation relative to human oversight.

### Properties on `AIAgent`

| Field Name | Type | Required/Optional | Description |
|---|---|---|---|
| `/Core/isoAutomationLevel` | `/Core/IsoAutomationLevel` | Optional (0..1) | ISO level of automation relative to human oversight |
| `agentCapability` | `/Core/DictionaryEntry` | Optional (0..*) | Describes a specific domain task or goal the agent is designed to perform |
| `agentExternalTool` | `/Core/DictionaryEntry` | Optional (0..*) | Identifies an external tool or service that the AI agent can invoke to extend its capabilities |
| `agentMemoryMode` | `AgentMemoryMode` | Optional (0..*) | Indicates the memory category or storage mechanism used by the AI agent |
| `agentCommunicationProtocol` | `/Core/DictionaryEntry` | Optional (0..*) | Identifies a standardized communication protocol implemented by the AI agent for interacting with external tools, data sources, or other agents |
| `agentTrustLevel` | `AgentTrustLevel` | Optional (0..1) | Indicates the degree of external verification applied to the AI agent's identity, behavioral claims, and declared capabilities |
| `agentType` | `AgentType` | Optional (0..*) | Classifies the communication and interaction role of the AI agent within a multi-agent system or agentic workflow |
[...]

PR #1221, opened 2026-02-19
Adding Role to SPDX

stevenc-stb:
Adding Role to SPDX core. 
Role can be used define defines a specific position or function that an entity plays in the context of a build, system, distribution, or lifecycle.

PR #1257, opened 2026-04-17
Add ai agent v2

### LINKED ISSUE #1440: Add "usesModel" relationship type for AIAgent
The new `AIAgent` needs "usedModel"

See https://github.com/spdx/spdx-3-model/pull/1257

ElyasRashno:
ai agent proposal #1256


model/AI/Vocabularies/AutomationLevel.md
```diff
@@ -0,0 +1,41 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# AutomationLevel
```
bact: This is already merged to Core as `IsoAutomationLevel`. ("ISO" prefix is added per suggestion from Max during review in Tech call)

See https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Vocabularies/IsoAutomationLevel.md

ElyasRashno: Done.

model/AI/Classes/AIAgent.md
```diff
@@ -0,0 +1,53 @@
...
+  - type: xsd:string
+  - minCount: 0
+- agentProtocol
+  - type: /Core/DictionaryEntry
+  - minCount: 0
+- agentType
+  - type: AgentType
+  - minCount: 0
+- automationLevel
+  - type: AutomationLevel
```
bact:
```suggestion
- /Core/isoAutomationLevel
  - type: /Core/IsoAutomationLevel
```

We already have this in Core.

model/AI/Properties/agentProtocol.md
```diff
@@ -0,0 +1,39 @@
...
+
+```text
+AConP: https://spec.acp.agntcy.org/
+```
+
+## Metadata
+
+- name: agentProtocol
+- Nature: ObjectProperty
+- Range: /Core/DictionaryEntry
```
bact: Using `/Core/ExternalIdentiferType` may be more suitable here? 

Then we can specify all protocols in that type. SBOM producers will pick the standardized names.

The downside of doing that is, the list is fixed, if the protocol is not in the list, the SBOM producer has to pick "other".

model/AI/Classes/AIAgent.md
```diff
@@ -0,0 +1,53 @@
...
+  - minCount: 0
+  - maxCount: 1
+- /Software/inputModality
+  - type: /Core/Modality
+  - minCount: 0
+- /Software/outputModality
+  - type: /Core/Modality
+  - minCount: 0
+
```
bact:
```suggestion

```

.gitignore
```diff
@@ -2,3 +2,4 @@
 *.bkp
 *.dtmp
 *.swp
+.github/ai-prompt-agent.agent.md
```
bact:
```suggestion
```

model/AI/Properties/automationLevel.md
bact, line 1: Remove this in favor of the existing /Core/isoAutomationLevel

model/AI/Vocabularies/AutomationLevel.md
bact, line 1: Remove this in favor of the existing /Core/IsoAutomationLevel

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -99,6 +100,7 @@ name completes the sentence:
 - tracedToDetail: the `from` Requirement is refined and further elaborated by each `to` Requirement, which contains more detailed implementation information.
 - trainedOn: The `from` Element has been trained on the `to` Element(s).
 - underInvestigationFor: The `from` Vulnerability impact is being investigated for each `to` Element. The use of the `underInvestigationFor` type is constrained to `VexUnderInvestigationVulnAssessmentRelationship` classed relationships.
+- usesModel: The `from` Element uses each `to` /AI/AIPackage as its underlying AI model.
```
bact:
```suggestion
- usesModel: The `from` Element uses each `to` /AI/AIPackage as its AI model.
```

May need more description of how this is different from `usesTool`

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -69,7 +69,8 @@ name completes the sentence:
 - hasMetadata: Every `to` Element is metadata about the `from` Element (`from` hasMetadata `to`).
 - hasOptionalComponent: Every `to` Element is an optional component of the `from` Element (`from` hasOptionalComponent `to`).
 - hasOptionalDependency: The `from` Element optionally depends on each `to` Element, during a LifecycleScopeType period.
-- hasOutput: The `from` Build, DefinedProcess or Action element generates each `to` Element as an output.
+- hasOutput: The `from` Build element generates each `to` Element as an output, during a LifecycleScopeType period.
+- hasPersistentMemory: The `from` Element uses each `to` Element as a persistent or long-term memory store (e.g., a database, vector store, or file system that retains information across sessions or reasoning steps).
```
bact:
```suggestion
- hasPersistentMemory: The `from` Element uses each `to` Element as a persistent or long-term memory store (e.g., a database, vector store, or file system).
```

This will make the relationship type usable by any computational machine.

model/AI/Vocabularies/AgentTrustLevel.md
```diff
@@ -0,0 +1,32 @@
...
+## Description
+
+AgentTrustLevel is an enumeration that categorizes the confidence a consumer
+can place in an AI agent's declared identity and behavioral claims, based on
+the rigor of verification performed. Trust levels progress from no formal
+verification (Level 0) to formal third-party certification (Level 3), and
+determine the scope of resources an agent may be permitted to access and the
+types of operations it may be authorized to execute.
+
+This enumeration is informed by the NIST AI Agent Standards Initiative trust
```
bact: Do we have link for this?

See example https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Vocabularies/IsoAutomationLevel.md?plain=1

model/AI/Properties/externalResources.md
```diff
@@ -0,0 +1,35 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# agentExternalTool
```
bact: Wonder if we can use this opportunity to update [/Core/Tool](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Classes/Tool.md) class.

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -69,7 +69,8 @@ name completes the sentence:
 - hasMetadata: Every `to` Element is metadata about the `from` Element (`from` hasMetadata `to`).
 - hasOptionalComponent: Every `to` Element is an optional component of the `from` Element (`from` hasOptionalComponent `to`).
 - hasOptionalDependency: The `from` Element optionally depends on each `to` Element, during a LifecycleScopeType period.
-- hasOutput: The `from` Build, DefinedProcess or Action element generates each `to` Element as an output.
+- hasOutput: The `from` Build element generates each `to` Element as an output, during a LifecycleScopeType period.
```
stevenc-stb:
```suggestion
- hasOutput: The `from` Build, DefinedProcess or Action element generates each `to` Element as an output.
```
 Hardware,SupplyChain is using hasOutput. Are you meaning to change the definition of it? is this a bug from a outdate git branch?

model/AI/Properties/agentCommunicationProtocol.md
```diff
@@ -0,0 +1,42 @@
...
+
+```text
+Agent Connect Protocol (AConP): https://spec.acp.agntcy.org/
+```
+
+## Metadata
+
+- name: agentCommunicationProtocol
+- Nature: ObjectProperty
+- Range: /Core/DictionaryEntry
```
stevenc-stb: is a CommunicationProtocol a /Core/Specification see https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Classes/Specification/

stevenc-stb (changes requested):
hasOutput is used by the Hardware,SupplyChain.

model/Core/Vocabularies/RelationshipType.md
bact: @ElyasRashno to fix this, you can sync `ElyasRashno:add-ai-agent-v2` branch to the latest one in `spdx:develop` branch.

ElyasRashno: Thanks. I rebased the agent branch. It's fine now.

model/AI/Properties/agentCommunicationProtocol.md
ElyasRashno: Thanks @stevenc-stb for pointing that out. Specification has exactly four entries: formalStandard, regulation, specification, other. I think none of these classify "a versioned protocol an agent speaks." For example, formal Standard specifically means "ratified by a recognized standards-development organization", A2A, MCP, ANP, AConP are vendor/community specs (Google, Anthropic, etc.), not a standard like ISO.

stevenc-stb: it is a specification. like a node like 
{
  "spdxId": "urn:nanoid:1",
  "type": "Specification",
  "creationInfo": "_:b0",
  "description": "An open standard that provides an architecture for AI clients to safely discover and invoke data sources, tools, and prompts via a unified interface.",
  "intendedUse": "Standardizing context and API capabilities for AI applications, agents, and local development tools.",
  "name": "Model Context Protocol Specification",
  "specType": "specification",
 "externalIdentifier": "_:b1",
  "summary": "An open protocol enabling seamless integration between LLM applications and external context sources."
}
 {
      "type": "ExternalIdentifier",
      "externalIdentifierType": "webpage",
      "identifier": "mcp-spec-v2025-11-25",
      "identifierLocator": "https://modelcontextprotocol.io/specification/2025-11-25",
      "@id": "_:b1"
    },

model/AI/Properties/agentCommunicationProtocol.md
```diff
@@ -0,0 +1,31 @@
...
+Identifies a standardized communication protocol implemented by the AI agent
+for interacting with external tools, data sources, or other agents.
+
+## Description
+
+Specifies the standardized communication protocol(s) that the AI agent
+implements, enabling consumers, integrators, and automated tooling to determine
+how the agent can be invoked, composed, or connected within a larger system.
+
+Each `agentCommunicationProtocol` entry references a `Specification` Element
```
bact:
```suggestion
Each `agentCommunicationProtocol` entry references a `Specification`
```

"Element" (capitalised) is a class name, similar to "Specification".
Having only `Specification` is enough.

model/AI/Classes/AIAgent.md
```diff
@@ -0,0 +1,61 @@
...
+  - type: AgentMemoryMode
+  - minCount: 0
+- agentCommunicationProtocol
+  - type: /Core/Specification
+  - minCount: 0
+- agentTrustLevel
+  - type: AgentTrustLevel
+  - minCount: 0
+  - maxCount: 1
+- externalResources
```
bact:
```suggestion
- externalResource
```

Use singular form. ref https://github.com/spdx/spdx-3-model/blob/develop/docs/format.md#naming-convention

model/AI/Properties/externalResources.md
```diff
@@ -0,0 +1,33 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# externalResources
```
bact:
```suggestion
# externalResource
```

Use singular form.

ref: https://github.com/spdx/spdx-3-model/blob/develop/docs/format.md#naming-convention

Cardinality is up to the class that uses the property. Not at the property itself to decide.

model/AI/Properties/externalResources.md
```diff
@@ -0,0 +1,33 @@
...
+code-interpreter: pkg:pypi/jupyter-kernel-gateway@2.0.0
+```
+
+```text
+pubmed-query: https://eutils.ncbi.nlm.nih.gov/entrez/eutils/
+```
+
+## Metadata
+
+- name: externalResources
```
bact:
```suggestion
- name: externalResource
```

model/AI/Properties/externalResource.md
bact, line 1: Rename the file to use singular form

model/AI/Properties/externalResources.md -> model/AI/Properties/externalResource.md

bact (changes requested):

model/AI/Vocabularies/AgentTrustLevel.md
bact, line 1: This is potentially be shared by non-AI use cases.
So we can consider to bring this up to Core namespace.

bact:
For `AgentTrustLevel`, we can try to map it to some standard definitions.

- `selfDeclared` looks like related to ISO/IEC 17050 Conformity assessment — Supplier's declaration of conformity
- `thirdPartyReviewed` - ISO/IEC 17020 Conformity assessment — Requirements for bodies performing inspection / ISO/IEC 17021 Conformity assessment — Requirements for bodies providing audit and certification of management systems
- `certified` - ISO/IEC 17065 Conformity assessment — Requirements for bodies certifying products, processes and services

It is also potentially be useful for non-AI use cases as well.
So probably consider bring it up to `/Core/TrustLevel` and make the description more generic. We can discuss this in next call.

stevenc-stb (approved):
LGTM

bact:
Btw, on the name, should we go `AIAgent` or `AiAgent` ?

Other classes like `SpdxDocument`, `EpssVulnAssessmentRelationship`, go by first letter of the word goes uppercase, the rest goes lowercase.

There's an exception from that pattern though, which is `AIPackage`.

bact (approved):
LGTM.

Note that we should revisit this again when https://github.com/spdx/spdx-3-model/pull/1280 and https://github.com/spdx/spdx-3-model/pull/1249 landed.

The agent, the prompt, and the interaction template should work together in the consistent way.

model/AI/Classes/AIAgent.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Description
+
+An AI agent is a software entity that uses artificial intelligence techniques to perceive its environment, reason about its state, and take actions to achieve specified goals.
+
+The following relationship types are also used to describe an AI agent's functionalities and interactions:
+
+- invokedBy: Describes a relationship from any Element back to the Agent that initiated its execution (e.g., a Package is invokedBy an AIAgent; an AIAgent is invokedBy a Person).
+- usesTool: Describes a relationship from the AIAgent to an Element that the agent uses as a tool to extend its capabilities.
+- usesModel: Describes a relationship from the AIAgent to an /AI/AIPackage that represents the underlying AI model the agent relies on (e.g., its model weights, training data, and safety assessments), enabling BOM consumers to trace from an agent to its associated model artifacts.
```
bact: Need to add this "usesModel" to RelationshipType vocab.

- https://github.com/spdx/spdx-3-model/issues/1440

bact:
```suggestion
- usesModel: Describes a relationship from the AIAgent to an AIPackage that represents the underlying AI model the agent relies on (e.g., its model weights, training data, and safety assessments), enabling BOM consumers to trace from an agent to its associated model artifacts.
```

bact (changes requested):

model/AI/Classes/AIAgent.md
ElyasRashno: Done.

bact (approved):

model/AI/Properties/agentTrustLevel.md
bact, line 1: We can have non-AI use cases for this trust level.
Maybe worth to move it to Core, as /Core/trustLevel.

model/AI/Properties/externalResource.md
```diff
@@ -0,0 +1,33 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# externalResource
```
kestewart: I think this should be a relationship to a component that the AI agent - ie. dependsOn, or some other existing relationship to a named component, rather than an arbitrary string.

model/AI/Vocabularies/AgentTrustLevel.md
kestewart: Rename to TrustLevel if want to make it general

model/Core/Vocabularies/RelationshipType.md
```diff
@@ -100,6 +100,7 @@ name completes the sentence:
 - tracedToDetail: The `from` Requirement is refined and further elaborated by each `to` Requirement, which contains more detailed implementation information.
 - trainedOn: The `from` Element has been trained on the `to` Element(s).
 - underInvestigationFor: The `from` /Security/Vulnerability impact is being investigated for each `to` Element. The use of the `underInvestigationFor` type is constrained to `/Security/VexUnderInvestigationVulnAssessmentRelationship` classed relationships.
+- usesModel: The `from` /AI/AIAgent uses each `to` /AI/AIPackage as a model.
```
kestewart: Should this be a Lifecycle scoped relationship?

seungwooketi:
### Implementation feedback: per-tool authorization scope is still not expressible

We built an SPDX 3.1 instance for a **running** MCP-based agent (5 MCP servers, 53 tools) as part of a research PoC on agent composition transparency, and validated it against the official RC1 JSON Schema (`v3.1-RC1/rdf/schema.json`, shacl2code-generated, Draft 2020-12, `jsonschema` Draft202012Validator). Sharing what did and did not survive contact with a real instance, since it bears directly on this PR.

**What we could express** (instance validates clean, 0 errors)

- `AIPackage` for the backbone, with `packageUrl` / `packageVersion` / `verifiedUsing` → `Hash` (sha256 of the model manifest digest)
- `/Core/Tool` nodes for each tool, `name` + `description` + `verifiedUsing` → `Hash` (sha256 over the tool description text)
- `usesTool` (agent → tools), `delegatedTo` two hops (operator → agent → server), `invokedBy` (server → agent)
- `Relationship.startTime` / `endTime` on the delegation edge

**What we could not.** There is no way to bind a permission scope to a specific tool. Two rejection experiments against the RC1 schema, same validator, same schema:

| | mutation | result |
|---|---|---|
| A | add `authorization: "..."` to the `SoftwareAgent` element | **rejected** — `Unevaluated properties are not allowed ('authorization' was unexpected)` |
| B | add a `Role` node carrying `authorization` | **rejected** — no `Role` class in RC1 |
| control | unmutated instance | valid |

Since then #1221 has merged, so `Role` and `authorization` exist on the 3.1-rc2 line and experiment B would now pass. That closes "there is no slot at all", but not the part that matters here: `authorization` is `xsd:string` on a `Role`, with **no binding to a particular tool**. With 53 tools you can state that a role is authorized in prose, but not *which tool* at *which scope* — and nothing is machine-checkable.

**How this PR sits relative to that gap.** Reading the properties as they currently stand here:

- `externalResource` — maps an identifier to a URI/purl. Says *that* a resource is reachable, not which operations on it are permitted.
- `agentCommunicationProtocol` — references a `Specification` for MCP/A2A/etc. Protocol-level, not per-tool.
- `agentCapability` — free-form task ability, and the description allows noting whether a capability needs human approval. That is useful, but it is prose inside a string, so approval requirements cannot be checked or diffed.

So after this PR one can record that an agent reaches `web-search` and speaks MCP, and that some capability "requires approval" in free text — but not that *tool X on server Y is permitted at scope Z, until time T, revocable on condition C*.

**Suggestion.** The `toolsAvailable { toolType, toolScope }` shape @bennetkl sketched in #1256 (2026-04-23) still looks like the right place to put this, and it does not appear in the current diff. Options as we see them:

1. include `toolsAvailable` with `toolType` and `toolScope` in `AIAgent` here; or
2. keep it in Core and allow a qualified `Role` → `/Core/Tool` binding, so `authorization` can attach per tool rather than per role; or
3. leave it to a separate profile that binds `Role` to `Tool`, if `AIAgent` is meant to stay descriptive rather than normative about permissions.

We have no preference among these — our interest is that *some* first-class, per-tool slot exists, because without it a consumer cannot tell a permitted invocation from an unpermitted one by reading the document.

**Gaps with no counterpart at all**, for whatever it is worth as input: in mapping our descriptor to SPDX Core we found no counterpart for revocation triggers (`endTime` gives "when" but not "why"), dynamic conditions, approval-required actions, or per-tool constraints. We are not proposing all of these belong in this PR — only that per-tool scope is the one that blocks the others.

Happy to share the instance and the two rejection cases as a reproducible example if that would be useful to the discussion.

bact:
@seungwooketi Thanks for really valuable detailed feedback. We will bring this to the SPDX AI WG call this week on Wednesday.

Everybody can join. See details at https://github.com/spdx/meetings#ai-and-dataset-profiles

I think we can looking more at `/Core/Tool` and see what we can do with it.
Previously we tend to think about a shipped package (distribution), so a lot of properties were added to `/Software/Package` and it is quite rich. Compare to `/Core/Tool`, it is quite thin and it was fine because tools were previously more for build time that tend to be one-off. The case no longer true for MCP case or agentic case. Tools (with no package-like distribution) become essential part and it needs more metadata, like what you have suggested.

model/AI/Classes/AIAgent.md
```diff
@@ -0,0 +1,56 @@
...
+Specifies an AI agent and its associated metadata.
+
+## Description
+
+An AI agent is a software entity that uses artificial intelligence techniques to perceive its environment, reason about its state, use available tools, and take actions to achieve specified goals.
+
+The following relationship types are also used to describe an AI agent's functionalities and interactions:
+
+- invokedBy: Describes a relationship from any Element back to the Agent that initiated its execution (e.g., a Package is invokedBy an AIAgent; an AIAgent is invokedBy a Person).
+- usesTool: Describes a relationship from the AIAgent to an Element that the agent uses as a tool to extend its capabilities.
```
bact: From 23 Sep 2026 AI WG call:

1) Differentiate between actual use and the availability for use.
 
  - usesTool (actual use)
  - availableTool

  Need to have a clear description for users to know the difference.

2) How should we restrict classes the relationship can point from/to (any `Element` or more restricted)? Something similar to:

  > availableTool: The from Element can use each Element as a tool, during a LifecycleScopeType period.

  A different tool may available in different lifecycle scopes. May have to think more about this?

PR #1269, opened 2026-05-15
Profile operations mk202605

mkurzman:
This is a draft as input for our next increment discussion.
Restarted with latest fork from develop and only contains the draw.io extract for profile operations.


bact:
The `profile-operations` branch itself may need to get updated as well. (it is currently 52 commits behind the `develop`)

PR #1276, opened 2026-05-19
UUID -> UID

### LINKED ISSUE #1222: Generalize *UUID
Few new classes in 3.1 have a property for universally unique identifier (UUID):
 
- [`evidenceUUID`](https://github.com/spdx/spdx-3-model/blob/develop/model/FunctionalSafety/Properties/evidenceUUID.md)
- [`requirementUUID`](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/requirementUUID.md)
- [`verificationUUID`](https://github.com/spdx/spdx-3-model/blob/develop/model/FunctionalSafety/Properties/verificationUUID.md)

Does it possible to have one `UUID` property in Core and let these classes use this common UUID property?

Also, at least in the case of `requirementUUID`, the definition there may not aligned with the UUID as defined by IETF RFC 9562 https://datatracker.ietf.org/doc/rfc9562/ since its description says it is "usually a string that follows a project specific naming convention." May need to rename?

bact:
Rename *UUID properties to *UID.

Update entry and entry description in ExternalIdentifierType.

To close #1222, as agreed in https://github.com/spdx/spdx-3-model/issues/1222#issuecomment-4489963321

Questions:

1) `evidenceUID` definition mentions "requirement" but we already have `requirementUID` not sure if it is intentionally overlapping or a typo.

1) `verificationUID` definition mentions "Verification" but we don't have `Verification` class.
    > A verificationUID is a unique identifier assigned to a Verification item.

    Or can it be `RequirementVerification` ?

Need reviews from @nicpappler and @stevenc-stb


stevenc-stb:
Verification-> RequirementVerification  class yes. @nicpappler , @chuckwolber  How should we reword verificationUID definition?

stevenc-stb (approved):
LGTM

nicpappler (approved):
Implements the agreed changes - LGTM

bact:
Will merge with 2 approvals.

PR #1277, opened 2026-05-19
Drop the minor version in namespace URIs

### LINKED ISSUE #1146: namespace problems in spdx-model.ttl
https://spdx.org/rdf/3.0.1/spdx-model.ttl and https://spdx.github.io/spdx-spec/v3.1-dev/rdf/spdx-model.ttl
- Please DO NOT use versioned namespaces. I say this after 15 years of practical experience  building KGs.
  - Versioned/named namespaces are unstable and contravene TimBL's simple principle "COOLURIs don't change".
  - They lead to unnecessary trouble for data producers/stewards. Imagine you have 10B SPDX triples, now a new ontology version comes out and you need to migrate the whole database
  - They lead to confusion. Eg `v3.1-dev` refers to both 3.0.1 and 3.1 `Core` as you see below
  - Tools are slow to implement namespace changes, and need to be more complicated to track them. Eg as you can see in https://github.com/underpin-project/dpp#spdx-rdf-basic-structure (next version describes how the 3.0.1 ontology differs), `tools-java` can convert 2.3 to Turtle, but uses a flat unversioned namespace.
- Define and use meaningful prefixes rather than using auto-increment prefixes (the last few are used but no prefix is defined):
```ttl
@prefix ns1: <https://spdx.org/rdf/3.0.1/terms/Core/> .
@prefix ns2: <https://spdx.org/rdf/3.1/terms/Core/> .
@prefix ns3: <https://spdx.org/rdf/3.1/terms/Dataset/> .
@prefix ns4: <https://spdx.org/rdf/3.1/terms/Security/> .
@prefix ns5: <https://spdx.org/rdf/3.1/terms/ExpandedLicensing/> .
@prefix ns6: <https://spdx.org/rdf/3.1/terms/Service/> .
@prefix ns7: <https://spdx.org/rdf/3.1/terms/AI/> .
@prefix ns8: <https://spdx.org/rdf/3.1/terms/SupplyChain/> .
@prefix ns9: <https://spdx.org/rdf/3.1/terms/Software/> .
@prefix ns10: <https://spdx.org/rdf/3.1/terms/Hardware/> .
<https://spdx.org/rdf/3.1/terms/Extension/>
<https://spdx.org/rdf/3.1/terms/Build/>
```
I.e. do this:
```ttl
@prefix spdx:         <https://spdx.org/rdf/terms/Core/> .
@prefix spdx-ai:      <https://spdx.org/rdf/terms/AI/> .
@prefix spdx-build:   <https://spdx.org/rdf/terms/Build/> .
@prefix spdx-data:    <https://spdx.org/rdf/terms/Dataset/> . # or dataset?
@prefix spdx-expLic:  <https://spdx.org/rdf/terms/ExpandedLicensing/> .
@prefix spdx-ext:     <https://spdx.org/rdf/terms/Extension/> . 
@prefix spdx-hw:      <https://spdx.org/rdf/terms/Hardware/> .
@prefix spdx-sec:     <https://spdx.org/rdf/terms/Security/> .
@prefix spdx-supply:  <https://spdx.org/rdf/terms/SupplyChain/> .
@prefix spdx-svc:     <https://spdx.org/rdf/terms/Service/> .
@prefix spdx-sw:      <https://spdx.org/rdf/terms/Software/> .
```

I would also like to argue that it's better to use a single flat namespace rather than sub-namespaces per topic. SPDX is not such a huge ontology to require splitting into sub-namespaces. But I guess such change will be too disruptive.

### LINKED ISSUE #1378: Term URL changing between versions?
I noticed while browsing the documentation and the context files that term URLs are changing per version. For example:

The [v3.0.1 definition of Agent](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Classes/Agent/) states the term URL as
`https://spdx.org/rdf/3.0.1/terms/Core/Agent`

but the [v3.1-RC1 definition of Agent](https://spdx.github.io/spdx-spec/v3.1-RC1/model/Core/Classes/Agent/) states the term URL as
`https://spdx.org/rdf/3.1/terms/Core/Agent`

However, the two definitions (at least at the human-readable documentation level) look identical.

Changing the identifiers of the terms based on the aggregate work (vs. on the evolution of the discrete term) means that one cannot rely on stable term definitions during the course of the evolution of the aggregate ontology...which makes managing things overtime untenable and requires re-ingesting old data in new formats/mappings when a new version is released.

That said, I may be missing something obvious that's driving this choice. 😅

So, I'd love to know if the changing term URLs is intentional for some design reason I've missed or just a byproduct of documentation tooling (or other) that hasn't been tripped over yet. 😄 

Thanks for any clarity you can provide!

Cheers,
🎩

goneall:
Update namespace URIs to use the form `https://spdx.org/rdf/3/`... removing the minor version per discussion on the 2026 19 2026 tech team call.
Fixes #1146
Fixes https://github.com/spdx/spdx-spec/issues/1378

### DIFF
--- a/docs/format.md
+++ b/docs/format.md
@@ -191,16 +191,16 @@ will give this RDF graph
 (in [Turtle syntax](https://en.wikipedia.org/wiki/Turtle_(syntax))):
 
 ```ttl
-<https://spdx.org/rdf/3.1/terms/SimpleLicensing/SimpleLicensingText> a owl:Class,
+<https://spdx.org/rdf/3/terms/SimpleLicensing/SimpleLicensingText> a owl:Class,
         sh:NodeShape ;
     rdfs:comment "A license or addition that is not listed on the SPDX License List."@en ;
-    rdfs:subClassOf <https://spdx.org/rdf/3.1/terms/Core/Element> ;
+    rdfs:subClassOf <https://spdx.org/rdf/3/terms/Core/Element> ;
     sh:nodeKind sh:IRI ;
     sh:property [ sh:datatype xsd:string ;
             sh:maxCount 1 ;
             sh:minCount 1 ;
             sh:nodeKind sh:Literal ;
-            sh:path <https://spdx.org/rdf/3.1/terms/SimpleLicensing/licenseText> ] .
+            sh:path <https://spdx.org/rdf/3/terms/SimpleLicensing/licenseText> ] .
 ```
 
 ## Syntax
--- a/docs/translation.md
+++ b/docs/translation.md
@@ -106,7 +106,7 @@ The above model Markdown file will generate the following RDF
 (showing partial):
 
 ```ttl
-<https://spdx.org/rdf/3.1/terms/Core/AnnotationType> a owl:Class;
+<https://spdx.org/rdf/3/terms/Core/AnnotationType> a owl:Class;
     rdfs:comment "Specifies the type of an annotation."@en ;
     rdfs:comment "注釈の種類を指定します。"@ja ;
     rdfs:comment "指定注释的类型。"@zh-hans ;
--- a/model/AI/AI.md
+++ b/model/AI/AI.md
@@ -15,5 +15,5 @@ development process, such as software packages, models, and datasets.
 
 ## Metadata
 
-- id: https://spdx.org/rdf/3.1/terms/AI
+- id: https://spdx.org/rdf/3/terms/AI
 - name: AI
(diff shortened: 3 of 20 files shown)


bact (approved):

bact:
Btw, this PR will not fully fix #1146.

Issue #1146 also about "meaningful prefixes rather than using auto-increment prefixes" (spdx, spdx-ai, spdx-build vs ns1, ns2, ns3) in the model file.

stevenc-stb (approved):
it look good to me.  However i'm not deep in the spec parser.

bact:
> it look good to me. However i'm not deep in the spec parser.

@stevenc-stb  We need this too https://github.com/spdx/spec-parser/pull/212

PR #1292, opened 2026-06-22
Move downloadLocation to SoftwareArtifact

### LINKED ISSUE #1032: [Proposal] Move downloadLocation from Software/Package to Software/SoftwareArtifact to use it also in Software/File
Currently, it is not possible to simply specify the download URI of a file in SPDX.
Therefore, I suggest moving the existing property '[downloadLocation](model/Software/Properties/downloadLocation.md)' from '[Software/Package](model/Software/Classes/Package.md)' to '[Software/SoftwareArtifact](model/Software/Classes/SoftwareArtifact.md)'. This would enable '[downloadLocation](model/Software/Properties/downloadLocation.md)' to be used in '[Software/File](model/Software/Classes/File.md)' as well. This would simplify the specification of the download URI for files.

As an alternative, it would be useful to move the '[downloadLocation](model/Software/Properties/downloadLocation.md)' to the '[Core properties](model/Core/Properties)', in order to specify a download URI for each element.
This would be especially useful for linking [SBOMs](model/Software/Classes/Sbom.md) without an '[SPDX ID](model/Core/Properties/spdxId.md)', e.g. CycloneDX SBOMs.

bact:
Move `downloadLocation` property from `Package` to `SoftwareArtifact`.

This will make the property available to all subclasses of `SoftwareArtifact`, including `File`, `Package`, and `Snippet`.

To resolve #1032, proposed by @ThomSei, discussed and agreed in 26 May 2026 Tech call. See proposed use case in https://github.com/spdx/spdx-3-model/issues/1032#issuecomment-3008913383.

## Backward compatibility

- Note that this change will change the SHACL and the JSON schema of `Package` (apart from `SoftwareArtifact`), but the net effect should cause no backward compatibility issue.
- In SHACL (generated by spec-parser), `downloadLocation` will be moved away from `sh:property` of `Package` and placed in `SoftwareArtifact`
- In JSON schema (generated by shacl2code), `spdx_SoftwarePackage_props` will no longer have `spdx:Software/downloadLocation` in its `properties` list. But as the condition `allOf` refers to the updated `spdx_SoftwareSoftwareArtifact_props`, the `downloadLocation` will still effectively seen.

  ```json
        "spdx_SoftwarePackage_props": {
            "allOf": [
                { "$ref": "#/$defs/spdx_SoftwareSoftwareArtifact_props" },
                {
                    "type": "object",
                    "properties": {
                        "spdx:Core/version": {
                            "$ref": "#/$defs/prop_spdx_SoftwarePackage_spdxCoreversion"
                        },
                        ...
  ```

  ```json
        "spdx_SoftwareSoftwareArtifact_props": {
            "allOf": [
                { "$ref": "#/$defs/spdx_CoreArtifact_props" },
                {
                    "type": "object",
                    "properties": {
                        ...
                        "spdx:Software/downloadLocation": {
                            "$ref": "#/$defs/prop_spdx_SoftwareSoftwareArtifact_spdxSoftwaredownloadLocation"
                        },
  ```


goneall (approved):
Already agreed to on a previous tech call

goneall:
We need one more review

bact:
Will merge this PR, agreed during two Tech calls, with one approval and 10+ days (three weeks).

PR #1293, opened 2026-06-22
Use qualified name in relationship type

bact:
Use fully qualified name for non-Core classes.

Only cross-namespace references are required to use fully qualified name.
As `RelationshipType` is in Core, Core classes (same namespace) can use bare name.

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -25,38 +25,38 @@ name completes the sentence:
 
 ## Entries
 
-- affects: The `from` Vulnerability, Action or DefinedProcess affects each `to` Element.
+- affects: The `from` /Security/Vulnerability, Action or DefinedProcess affects each `to` Element.
 - amendedBy: The `from` Element is amended by each `to` Element.
 - ancestorOf: The `from` Element is an ancestor of each `to` Element.
-- assumes: The `from` Element assumes each `to` Assumption.
+- assumes: The `from` Element assumes each `to` /FunctionalSafety/Assumption.
 - availableFrom: The `from` Element is available from the additional supplier described by each `to` Element.
 - configures: The `from` Element is a configuration applied to each `to` Element, during a LifecycleScopeType period.
-- conformsTo: The `from` Element conforms to each `to` Assumption or Specification.
+- conformsTo: The `from` Element conforms to each `to` /FunctionalSafety/Assumption or Specification.
 - contains: The `from` Element contains each `to` Element.
-- coordinatedBy: The `from` Vulnerability is coordinatedBy the `to` Agent(s) (vendor, researcher, or consumer agent).
+- coordinatedBy: The `from` /Security/Vulnerability is coordinatedBy the `to` Agent(s) (vendor, researcher, or consumer agent).
 - copiedTo: The `from` Element has been copied to each `to` Element.
 - createdBy: The `from` Action or DefinedProcess is createdBy `to` Agent(s).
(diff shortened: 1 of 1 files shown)


bact:
The fully qualified name will make it possible for us to generate SHACL (see a generated [example](https://openregtech.github.io/spdx-spec/v3.1-dev/rdf/spdx-model.ttl)) and allow more complete validation.

goneall (approved):
Although this does somewhat negatively impact the human readability, I think it is a reasonable tradeoff to support generating stronger SHACL validation

stevenc-stb (commented):
this will need to be talked about in the tech meeting.  If we proceed, are there other places in the SPDX documentation that need to be updated to maintain consistency?

bact:
> this will need to be talked about in the tech meeting. If we proceed, are there other places in the SPDX documentation that need to be updated to maintain consistency?

I think, yes, but only for cross-namespace references.

I searched the spdx-3-model repo to find estimated number of instances that require fully qualified name (cross-namespace references). It is nearly 200.

But the search also bring interesting spots that we may like to make revisions.

- Almost 40 are in [/Software/SoftwarePurpose](https://github.com/spdx/spdx-3-model/blob/develop/model/Software/Vocabularies/SoftwarePurpose.md) which references a lot of [/Core/Element](https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Classes/Element.md). All of these are not necessary if we scoped the /Software/SoftwarePurpose to /Software/SoftwareArtifact (same namespace). Currently (3.1-dev), /Software/SoftwarePurpose is only used within Software namespace. If we replaced all "Element"s with "SoftwareArtifact", no need for qualifier.
- Some of them are licensing related and potentially wrong use of Capitalisation. For example, in the description of [/SimpleLicensing/SimpleLicensingText](https://github.com/spdx/spdx-3-model/blob/develop/model/SimpleLicensing/Classes/SimpleLicensingText.md):

  > A SimpleLicensingText represents a License or Addition that is not listed on

  at first glance, `License` and `Addition` looks like class names and may need a qualifier as they are not in SimpleLicensing namespace.

  `License` is from ExpandedLicensing namespace. So we have:

  > A SimpleLicensingText represents a /ExpandedLicensing/License or Addition that is not listed on

  `Addition` is actually a typo, it does very likely be `LicenseAddition` which is also in ExpandedLicensing. So it's getting more ugly:

  > A SimpleLicensingText represents a /ExpandedLicensing/License or /ExpandedLicensing/LicenseAddition that is not listed on

  But then we found that this does not make sense. Because the purpose of SimpleLicensingText is to have a very simple representation of a license or a license addition. It contains only one (direct) property, [/SimpleLicensing/licenseText](https://github.com/spdx/spdx-3-model/blob/develop/model/SimpleLicensing/Properties/licenseText.md), and it does not linked with License or LicenseAddition classes.

  It will be more precise to revise the description as:

  > A SimpleLicensingText represents a license or a license addition that is not listed on
  
  Since "license" and "license addition" (both lowercase) here are general concepts, not class names.

  Same for "Custom License" and "Custom License Addition" in [/SimpleLicensing/customIdToUri](https://github.com/spdx/spdx-3-model/blob/develop/model/SimpleLicensing/Properties/customIdToUri.md), they should be "custom license" and "custom license addition".

If we revise these kinds of proses, reduce unnecessary (or imprecise) Capitalisation, use generic references when possible, the references to Class names will be reduced, and the instances that need fully qualified name will be reduced as well.
 
We may also need a style guide on when to use Capitalisation and when to use `FixedWidth` typeface.

bact:
From 23 June 2026 Tech call:

- Consensus: worth adding this in, and checking that there are no other refs that should be qualified.
- @kestewart to do review, pass and merge if no issues are detected.

kestewart (approved):

PR #1296, opened 2026-06-25
projectOwner, productAgent: Consistent language + expand abbreviation

bact:
- Use "Agent responsible for ..." template to make languages in these two properties consistent:
  - `productAgent`
  - `projectOwner`
- Expand "OEM" abbreviation
- Fix "Person" -> "Agent" in `projectOwner` to match its range (`/Core/Agent`) (note: `Person` is subclass of `Agent`; If intended to use "Person", should update the range)


stevenc-stb (approved):
LGTM

goneall (approved):

PR #1297, opened 2026-06-29
Add hasInstall and hasUninstall relationships

JPEWdev:
Adds relationships that indicate the installation and uninstallation of artifacts during a lifecycle scope (e.g. build, runtime, etc.)


model/Core/Vocabularies/RelationshipType.md
```diff
@@ -66,6 +66,7 @@ name completes the sentence:
 - hasExample: Every `to` Element is an example for the `from` Element (`from` hasExample `to`).
 - hasHost: The `from` Build was run on the `to` Element during a LifecycleScopeType period (e.g. the host that the build runs on).
 - hasInput: The `from` Build, DefinedProcess or Action element has each `to` Element as an input.
+- hasInstall: The `from` Element has each `to` Artifact installed (added) into it during a LifecycleScopeType period. To express the Agent doing the install, link this relationship as the `from` in `performedBy`.
```
bact: Does this mean we have to update the allowed type of `from` in `performedBy` as well?

model/Core/Vocabularies/RelationshipType.md
bact:
```suggestion
- performedBy: Every `from` Action or Relationship is performedBy `to` Agent.
```

model/Core/Vocabularies/RelationshipType.md
JPEWdev: Ya, I missed that. Perhaps we can talk about it in the SPDX call tomorrow

kestewart:
Usecase coming from Yocto project.   Root file system image contains files,  and files are contained by packages came from.   Can only link root file systems to packages via contains & files.   Will be useful to be able to be explicitly - install and uninstalled.   Illustrate not used in root file system.   Runtime dependencies being uninstalled is not quite the right semantic.   What's on shipping vs. being running.  

Should the supply chain action be used?   Installed on a platform, not artifact?   Not quite, can do a cross compile.   Image being manipulated outside the context.   Host platform is capture in build profile. 

Can install things in build time vs run time.     Period type?   Definition of the word install?   apt install,  brew,  ...   boot loader installing a system image. 

Possibly lifecycle scope extension - manufacturing?  (consider later)

Don't want to conform to supply chain profile.   No conformance points though right now.   

Artifact installing, indicates method likely.   ie. debian package;  debian pkgs, apt. 

Installing is placing in position for use?

goneall (approved):
Aside from @kestewart questions in the comments which are related to Yocto, I can see this being a useful new relationship type.

bact (approved):

PR #1265, opened 2026-05-12
Use generic version property

bact:
To address #1225

Instead of having many `xxxVersion` properties to maintain and potentially complicating retrieval query, let classes share a common `version` (xsd:string) property.

- Use generic `version` property in `/Core/DefinedProcess`, `/Hardware/Hardware`, `/Software/Package`, and `/Hardware/ProductSpecification`
  - Changes in `/Core/DefinedProcess`, `/Hardware/Hardware`, and `/Hardware/ProductSpecification` do not introduce any compatibility issue as they are newly introduced in 3.1
  - Keep backward compatibility with 3.0 by just adding `/Core/version` and not removing anything from `/Software/Package`
  - Add deprecation message to `/Software/packageVersion`
- Rename `/Core/Version` datatype to `/Core/VersionNumber`
  - Avoid confusion, make it distinct from `/Core/version` property which is a string
  - VersionNumber does not support SemVer extensions, it is purely decimal numbers and dots, so the "Number" suffix is accurate
  - Does not introduce compatibility issue as SPDX 3 datatype names are internal and are not exposed to OWL ontology or JSON schema or JSON-LD context document.

PR #1304, opened 2026-07-11
Tool: Add version property

### LINKED ISSUE #1303: Should `Tool` class has a `version` property ?
`/CreationInfo.createdUsing` has a range of `/Core/Tool`.

Tool can be either hardware or software:

> An element of hardware and/or software utilized to carry out a particular function.

But not exactly the `/Hardware/Hardware/` or `/Software/Package` according to current model, so it doesn't have `/Core/version` property. (In fact, `/Core/Tool` doesn't have any additional property, it just has what bare `/Core/Element` would have)

Should we allow `/Core/Tool` to have the `/Core/version` property?

This can be used to record the version of SBOM generator as well.

bact:
Add /Core/version property to /Core/Tool class

To close #1303

### DIFF
--- a/model/Core/Classes/Tool.md
+++ b/model/Core/Classes/Tool.md
@@ -15,3 +15,10 @@ A Tool is an element of hardware and/or software utilized to carry out a particu
 - name: Tool
 - SubclassOf: Element
 - Instantiability: Concrete
+
+## Properties
+
+- version
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1


bact:
@AlexanderDenkBMW please kindly review. Thank you.

AlexanderDenkBMW (approved):

goneall:
Do we need to add a `version` property in the core properties directory?  There is a `packageVersion` in the `Software` profile and a `specVersion` in core, but no `version` property.

Also - for consistency - I would suggest naming it `toolVersion`.

bact:
> Do we need to add a `version` property in the core properties directory? There is a `packageVersion` in the `Software` profile and a `specVersion` in core, but no `version` property.

We already have /Core/version - https://github.com/spdx/spdx-3-model/blob/develop/model/Core/Properties/version.md - as part of the generalization attempt of many *`version`s.
- see #1265

> Also - for consistency - I would suggest naming it `toolVersion`.

To follows #1265, we can utilize `/Core/version` for `/Core/Tool`

goneall (approved):
LGTM

PR #1413, opened 2026-07-20
Add in requirementStatus field Requirement.md

kestewart:
Need to be able to express if status is draft, reviewready, active, retired, backlog - per discussions in SPDX safety profile and ELISA architecture meetings


model/Core/Vocabularies/RequirementStatusType.md
```diff
@@ -0,0 +1,23 @@
...
+
+# RequirementStatusType
+
+## Summary
+
+Provide an enumerated set of requirement statuses that can provide state to a requirement.
+
+## Description
+
+This enumeration summarizes common states that a requirement may pass through in it's lifecycle.  The same requirement wording may have different states depending on the context of the product it is being included in.
```
kestewart:
```suggestion
This enumeration summarizes common states that a requirement may pass through in it's lifecycle.  The same requirement wording may have different states depending on the context of the product it is being included in.
```

model/Core/Classes/Requirement.md
```diff
@@ -38,3 +38,7 @@ requirement statement itself.
   - type: ExternalIdentifier
   - minCount: 0
   - maxCount: 1
+- reqirementStatus
+  - type: RequirementStatusType
+  - mincount: 0
+  - maxcount: 1
```
bact:
```suggestion
  - minCount: 0
  - maxCount: 1
```

stevenc-stb (changes requested):
In the FUSA meeting, the solution discussed was to make requirementStatus an part relationship between Requirement and Project (some Element), rather than a property of Requirement. Making it part of Requirement would cause problems, because a change in requirementStatus would generate a new SPDX ID for the node. 

It also makes things harder for users: when the requirementStatus changes that makes a new Requirement, people would have to figure out whether the EvaluationResult for the old Requirement - obtained through a RequirementVerification - would still be valid for the new Requirement. That creates so much more work. The better approach is simply to make requirementStatus part of the relationship between Requirement and Project( or some Context Element) .

bact:
> Making it part of Requirement would cause problems, because a change in requirementStatus would generate a new SPDX ID for the node.

What if the requirementStatus is a real characteristic of the requirement and not the review workflow status?

--

I think we have two kinds of status here:

1) Requirement document status -- The status reflects completeness and maturity of the document itself. 
2) Requirement review/workflow status -- The status reflects acceptance/opinion towards the document, from a specific point of view.

A document can be complete (status in (1) category) and rejected (status in (2) category).

--

Related examples,

In [**Asset Description Metadata Schema (ADMS)**](https://semiceu.github.io/ADMS/releases/2.00/#status), we have these status vocabs:

- http://purl.org/adms/status/Completed
- http://purl.org/adms/status/UnderDevelopment
- http://purl.org/adms/status/Deprecated
- http://purl.org/adms/status/Withdrawn
- http://publications.europa.eu/resource/authority/dataset-status/COMPLETED
- http://publications.europa.eu/resource/authority/dataset-status/DEVELOP
- http://publications.europa.eu/resource/authority/dataset-status/DEPRECATED
- http://publications.europa.eu/resource/authority/dataset-status/WITHDRAWN
- http://publications.europa.eu/resource/authority/concept-status/CURRENT

(the purl.org ones are terms from early ADMS draft, the europa.eu ones are terms since the ADMS is maintained by the EU SEMIC program)

The "COMPLETED" and "DEVELOP"/"UnderDevelopment" are in (1) category, it is the status of the document content.

The "DEPRECATED", "WITHDRAWN", "CURRENT" are in (2) category, it is the status towards the document and temporal.

In [**Bibliographic Ontology (BIBO)**](https://www.dublincore.org/specifications/bibo/bibo/), we have:

a) http://purl.org/ontology/bibo/status/draft 
b) http://purl.org/ontology/bibo/status/nonPeerReviewed 
c) http://purl.org/ontology/bibo/status/peerReviewed 
d) http://purl.org/ontology/bibo/status/accepted
e) http://purl.org/ontology/bibo/status/rejected 
f) http://purl.org/ontology/bibo/status/forthcoming 
g) http://purl.org/ontology/bibo/status/published 
h) http://purl.org/ontology/bibo/status/unpublished 
i) http://purl.org/ontology/bibo/status/legal

(a) and (i) are status of the document itself. (a) reflects the content maturity. They are in (1) category.

The rest are review status or publication status, and are temporal. They are in (2) category.

From the current proposal

I think these are in (1) category:

- draft
- reviewable
- backlog

And these are in (2) category:

- active
- retired

----

Btw, I found this document that link to OMG [**Requirements Interchange Format (ReqIF)**](https://www.omg.org/reqif/) -- the ReqIF Workflow Forum (ReqIF WF) and ReqIF Implementor Forum (ReqIF IF) jointly publish [ReqIF Recommendation Comprehensive Collection of Industrial ReqIF Use Cases Version 1.0](https://www.prostep.org/fileadmin/prod-pay-download-8c1d/PSI18_prostep-ivip_Recommendation_ReqIF_V1-0.pdf). If you look at Table 2 (p.13), Table 3 (p.14), Figures 7 & 8 (p.16), you can see `ReqIF-WF.CustomerStatus` and `ReqIF-WF.SupplierStatus` are their interaction. Maybe of your interest.

model/Core/Classes/Requirement.md
```diff
@@ -38,3 +38,7 @@ requirement statement itself.
   - type: ExternalIdentifier
   - minCount: 0
   - maxCount: 1
+- reqirementStatus
```
bact:
```suggestion
- requirementStatus
```

typo

model/Core/Properties/requirementStatus.md
```diff
@@ -0,0 +1,17 @@
...
+
+## Description
+
+The status specifies the state of a requirement in its lifecycle.
+
+## Metadata
+
+- name: requirementStatus
+- Nature: DataProperty
+- Range: requirementStatusType
```
bact:
```suggestion
- Range: RequirementStatusType
```

typo

kestewart:
@stevenc-stb - After thinking through this further, putting the status on the relationship is going to cause different problems.   Status is more naturally tied to the statement that's been reviewed, so keeping the text and status together seems like the path of clarity.

stevenc-stb:
> @stevenc-stb - After thinking through this further, putting the status on the relationship is going to cause different problems. Status is more naturally tied to the statement that's been reviewed, so keeping the text and status together seems like the path of clarity.

There is no single, neutral arbiter of completeness or maturity. Every assessment comes from an agent with a specific role, context, and purpose. 

RequirementStatement -> item shall be blue.  From the perspective of a marketing team or a product owner, the first statement ("shall be blue") might be perfectly complete. It’s unambiguous enough for them to make decisions about branding, packaging, or customer-facing materials.
From the perspective of an optical engineer or quality assurance tester, that same statement is incomplete. They need the wavelength range to verify compliance, design test fixtures, or select materials.

chuckwolber:
TL;DR - I should be free to point to a specific requirement at a specific commit, and not have to explain to the world why the word "draft" appears in the definition of a requirement that my design relies upon.

The longer answer is probably worth reading though...

There are two problems here:

1. Once published to a project, a numbered requirement is forever immutable and cannot be changed. If a requirement is wrong, you write a new one.
2. A requirement can be reused in many different projects. There is no such thing as an outbound requirement link; projects know about requirements, requirements cannot know about projects.

On that basis, you have two forms of requirement status:

1. The requirement wording itself: draft or published
2. The requirement's relationship to the committed design: accepted or not-accepted

Referring to the comment @bact made [here](https://github.com/spdx/spdx-3-model/pull/1413#issuecomment-5023919825), there is some nuance involved because validation involves cycles of customer review and acceptance. But at a high level the problem boils down to draft/published and accepted/not-accepted.

A requirement in draft state obviously cannot be considered "accepted" by a project, although it probably implies a form of "tentative" acceptance. An accepted requirement cannot be in draft state, because acceptance is a project level design commitment.

Once published, a requirement cannot be modified because you cannot assume that you are the only user of that requirement. And even if you could prove this, you have code/test/docs/nomenclature/culture/etc accreted around the old requirement that needs a clearly demarcated path from old to new.

So to the extent that the requirement itself needs to telegraph to the world - "do not be surprised if you wake up tomorrow and my wording has changed", then a draft/published status on the requirement itself might be justifiable. But there are still problems with that approach because the design intent is what "shapes" the individual requirement. 

**In other words: requirements do not arise out of a vacuum, requirements arise out of a relationship.**

An aerospace example might help drive home the real problem with putting a status on the requirement itself. Nature dictates the design of the wing, not the other way around. Engineers who disagree with nature tend to find themselves considering other forms of employment. Therefore we iterate a lot until we have an optimal wing, and along the way we throw out a lot of designs. But discarded designs are still legitimate designs and on occasion we can make use of those ideas elsewhere (e.g. maybe water was a better medium, etc).

I believe this somewhat speaks to the point @stevenc-stb made [here](https://github.com/spdx/spdx-3-model/pull/1413#issuecomment-5026988463). "Item shall be blue" might be a good thing until an optical engineer says, "blue is fine if you want block all of the wavelengths you actually care about".

That does not mean "Item shall be blue" is invalid on its face; refer to the point made in problem 2 - requirements have no outbound links. It just means "Item shall be blue" is an invalid requirement for **that particular project.** Therefore the relationship with "Item shall be blue" needs to be eliminated from the committed design and a new relationship added with a "Item shall be red" requirement.

So fundamentally, this probably boils down to a practical concern about what we need to telegraph while the requirement is going through its wordsmithing phase prior to any form of design acceptance. After that, the problem is exclusively the domain of the requirement's relationship with a particular committed design.

And as tempting as the practical concerns may be, there is probably a near 100% chance of a requirement status field being used incorrectly and becoming an ingrained problem area. And even if it were used correctly, I should still be free to point to a specific requirement at a specific commit, and not have to explain to the world why the word "draft" appears in the definition.

bact:
For high-level draft/published and accepted/not-accepted, the [BIBO](https://www.dublincore.org/specifications/bibo/bibo/) ontology is sufficient then:

- http://purl.org/ontology/bibo/status/draft
- http://purl.org/ontology/bibo/status/published
- http://purl.org/ontology/bibo/status/accepted
- http://purl.org/ontology/bibo/status/rejected

kestewart:
Actively discussed this in the Safety meeting, and explored use cases and how it would be leveraged.
Focused discussion on SPDXid vs. RequirementUID.   RequirementID is delegated to a particular agent.   Specification of what it means stays with the agent.    If you make a copy of the requirement, you need to use a new requirement ID.

SPDXid will change underneath them as the status changes.   This is a way of expressing version. 

Highlighted that we should treat requirements like code vs. bringing in a binary.   Example of Yocto source vs. Debian distribution of binary.    Products including requiements bring them under their own control and bring them into their context.   There are components that may need to be reused untouched.    

Use cases to contemplate:
- Requirement Tooling
- Product lines - how they would work. 
- Change Impact Analysis - any change should be assessed draft --> approved.  Approved to new revision of it. 

Discussion of teasing should just be Draft/Approved?    Tooling needs the other states. 

Participants:   Chuck, Nicole, Greg, Stephen, Alfred, Kate

stevenc-stb (approved):
Note: RequirementStatusType will need more work.  the change can be done in different PR

model/Core/Properties/requirementStatus.md
```diff
@@ -0,0 +1,17 @@
+SPDX-License-Identifier: Community-Spec-1.0
+
+# requirementStatus
+
+## Summary
+
+The assessment of the state of a requirement.
```
bact, line 7: This summary does not quite match the description

chuckwolber:
> Actively discussed this in the Safety meeting, and explored use cases and how it would be leveraged. Focused discussion on SPDXid vs. RequirementUID. RequirementID is delegated to a particular agent. Specification of what it means stays with the agent. If you make a copy of the requirement, you need to use a new requirement ID.

In practice it should be reasonable to reuse the requirement ID assuming it does not conflict with local (i.e. project specific) policies. This is a good reason for all projects to consider adopting a hash generated form of requirement IDs that have a high likelihood of global uniqueness (i.e. please do not use [RFC 9562](https://www.rfc-editor.org/info/rfc9562/) UUIDs for requirement IDs).

> 
> SPDXid will change underneath them as the status changes. This is a way of expressing version.

Concur.

I also recommend that we consider where the distinction lies with respect to the use of a hash key to tie together relevant implementation and requirement. The hash key concept is necessary for projects (e.g. open-source) that are not specifically intended to address safety (or security) critical needs. It allows for traceability when the implementation evolves without considering the impact on requirements.

> 
> Highlighted that we should treat requirements like code vs. bringing in a binary. Example of Yocto source vs. Debian distribution of binary. Products including requiements bring them under their own control and bring them into their context. There are components that may need to be reused untouched.

Where requirements can be referenced by a project (e.g. via `SRC_URI`) and only the deviations need to be retained by the project. This concept is foundational to the way bitbake recipes work in the Yocto/OpenEmbedded ecosystem. 

> 
> Use cases to contemplate:
> 
>     * Requirement Tooling
> 
>     * Product lines - how they would work.
> 
>     * Change Impact Analysis - any change should be assessed draft --> approved.  Approved to new revision of it.

Deviation retention would drive an SCIA (Software Change Impact Analysis).
 
> 
> Discussion of teasing should just be Draft/Approved? Tooling needs the other states.

Concur. I recommend we limit the published enumeration to only what is required to telegraph the intent of the actual content of the requirement (e.g. the wording), and avoid any form of status that implies the way the requirement is being **used**. Draft/Published should be sufficient to cover the absolute minimum intent.

Draft == "Only the requirement ID is immutable. The wording will change without notice."
Published == "Both the requirement ID and the requirement wording are permanently immutable."

> 
> Participants: Chuck, Nicole, Greg, Stephen, Alfred, Kate

@kestewart If you add me as a reviewer to this PR I can add my approval. I will also generate issues to address follow-on discussions (use of hash-key, and draft/published state enumeration).

gregshue:
> In practice it should be reasonable to reuse the requirement ID assuming it does not conflict with local (i.e. project specific) policies. This is a good reason for all projects to consider adopting a hash generated form of requirement IDs that have a high likelihood of global uniqueness (i.e. please do not use [RFC 9562](https://www.rfc-editor.org/info/rfc9562/) UUIDs for requirement IDs).

Reusing the requirement ID without reusing the statement (or state) is an example of clone-and-own paradigm. This is the antithesis of the large-scale-reuse paradigm that is the foundation of PLE. I don't think we can prevent clone-and-own usage, but we must document the behavior restrictions needed to support large-scale-reuse, especially for development of components/elements/assets that are intended to be reused by other organizations (e.g., substantial functionality of Zephyr is intended to be reused/extended by chip and IoT device manufacturers). 

> > Highlighted that we should treat requirements like code vs. bringing in a binary. Example of Yocto source vs. Debian distribution of binary. Products including requiements bring them under their own control and bring them into their context. There are components that may need to be reused untouched.

An important use case is when a product integrator wants to reuse the (published) requirement without bringing them under their own control. This is appropriate when the product context is a subset of the context of the component/element/asset being brought in. Zephyr's stated context (i.e., PLE "Domain") is outlined in the Zephyr Project Charter as "deliver the best-in-class RTOS for connected resource-constrained devices, built to be secure and safe". As the Zephyr Project has evolved the domain has grown beyond RTOS to include non-RTOS images (e.g., Zephyr minimal build). Once the MCUboot functionality was integrated into the Zephyr ecosystem we have a configurable feature set that Product Manufacturers would like to completely reuse (including requirement statements and verification processes). In fact, the Zephyr Project Charter explicitly intends the Security Artifacts to be reusable by the Platinum Members - which requires reuse of requirements as well as code.

> > * Product lines - how they would work.

For PLE the context of the requirement statements is the entire Product Line Domain, which includes existing, intended, and reasonably foreseeable products tangential to the existing/intended scope. Since the requirement statements must know nothing about the specific product(s), the standalone/complete/unambiguous requirement statements must include a condition clause (e.g., EARS "Where" clause referencing a domain-specific configuration parameter). Each product will have a bill-of-features that is used to filter the aggregate requirement set down to those relevant to said product.

> > Discussion of teasing should just be Draft/Approved? Tooling needs the other states.
> 
> Concur. I recommend we limit the published enumeration to only what is required to telegraph the intent of the actual content of the requirement (e.g. the wording), and avoid any form of status that implies the way the requirement is being **used**. Draft/Published should be sufficient to cover the absolute minimum intent.
> 
> Draft == "Only the requirement ID is immutable. The wording will change without notice." Published == "Both the requirement ID and the requirement wording are permanently immutable."

I agree with only standardizing Draft/Published (or Draft/Published/Deprecated). Any additional states should be provided by the tool through a tool-, process-, or vendor-specific Extension.

model/Core/Properties/requirementStatus.md
kestewart: I didn't want to say for Summary "The status of a requirement" which feels redundant.  I'm going to go ahead and merge this.   If you've got a better suggestion for the wording, please open a new PR.  Chuck is planning to already for the states.   So this isn't done, but at least will move forward.

kestewart:
@chuckwolber - I can't seem to add you.   You should be able to add a review though without being added explicitly.   Given your comments though and Steven's approval.  I'm going to go ahead and merge this.

Issue #1367, opened 2026-07-14
[3.1-RC1] Add missing ExternalIdentifierType values and update its RFC citation


henrylyons:
**Affected Sections**: model/Core/Vocabularies/ExternalIdentifierType.md<br><br>**Suggested Change:**<br>Consider adding iban, vatNumber, eori, and bicSwift (common in supplier traceability), or document why they're excluded. Update the webpage entry, which cites the obsoleted RFC 7230, to RFC 9112/RFC 3986. (model/Core/Vocabularies/ExternalIdentifierType.md)<br><br>**Notes:**<br>This feedback was provided by Microsoft at the SPDX working group meeting on June 30, 2026.

bact:
Common supplier traceability IDs:
- `bicSwift` - already included as `bic` (ISO 9362)
- `iban` - missing - was initially part of this PR https://github.com/spdx/spdx-3-model/pull/1187, but later [removed](https://github.com/bact/spdx-3-model/commit/695f3ee084580beff11819a0e5e029cf379bafbd) due to concerns of the sensitive nature of data.
- `vatNumber` - missing - VAT
- `eori` - missing - Economic Operators Registration and Identification number - for EU custom/export/import/transit purpose

This PR https://github.com/spdx/spdx-3-model/pull/1420 will add `iban`, `vatNumber`, `eori`.

webpage:
- RFC 7230 - see PR https://github.com/spdx/spdx-3-model/pull/1417 (it is about "http" and "https", RFC 9112/RFC 3986 don't have that section, need RFC 9110)

PR #1417, opened 2026-07-21
Update RFC 7230 (obsoleted) -> RFC 9110

bact:
Also update section numbers to the new RFC

To address one of issues reported in https://github.com/spdx/spdx-3-model/issues/1367

(Do we need to fix 3.0 too?)


stevenc-stb (approved):
LGTM. ( Note there might be a add definition need )
in rfc9110
  http-URI = "http" "://" authority path-abempty [ "?" query ]
in RFC 7230  
http-URI = "http:" "//" authority path-abempty [ "?" query ]
                 [ "#" fragment ]
                 
is fragment sill allowed in rfc9110

stevenc-stb (commented):
Further Reading this might be a breaking change. The fragment in [4.2.5. ](https://datatracker.ietf.org/doc/html/rfc9110#section-4.2.5)[http(s) References with Fragment Identifiers](https://datatracker.ietf.org/doc/html/rfc9110#name-https-references-with-fragm)
is not part of the [4.2.1](https://datatracker.ietf.org/doc/html/rfc9110#section-4.2.1) and in [2.7.1](https://datatracker.ietf.org/doc/html/rfc7230#section-2.7.1) it is. This appears to be a type shrinkage.   Do we need to a add with note adding in  [4.2.5.](https://datatracker.ietf.org/doc/html/rfc9110#section-4.2.5)

bact:
from 4.2.5. http(s) References with Fragment Identifiers:

> Fragment identifiers allow for indirect identification of a secondary resource, independent of the URI scheme, as defined in [Section 3.5](https://www.rfc-editor.org/rfc/rfc3986#section-3.5) of [[URI](https://datatracker.ietf.org/doc/html/rfc3986)].

Does this mean the fragment will be part of any URI anyway, without the need to explicitly mention it?

model/Core/Vocabularies/ExternalIdentifierType.md
```diff
@@ -58,4 +58,4 @@ ExternalIdentifierType specifies the type of an external identifier.
 - swid: Concise Software Identification (CoSWID) tag, as defined in [RFC 9393](https://datatracker.ietf.org/doc/rfc9393/) Section 2.3.
 - urlScheme: [Uniform Resource Identifier (URI) Schemes](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml). The scheme used in order to locate a resource.
 - verificationUID: The unique identifier used by a verification management system or any other lifecycle management tool to uniquely identify a verification item.
-- webpage: Absolute URL that can be used to locate a resource, as defined in [RFC 7230](https://datatracker.ietf.org/doc/rfc7230/) Section 2.7.1 or Section 2.7.2.
+- webpage: Absolute Uniform Resource Locator (URL) that can be used to locate a resource, as defined in [RFC 9110](https://datatracker.ietf.org/doc/rfc9110/) Section 4.2.1 or Section 4.2.2.
```
stevenc-stb:
```suggestion
- webpage: A Uniform Resource Locator (URL) that can be used to locate a resource, as defined in [RFC 9110](https://datatracker.ietf.org/doc/rfc9110/) Section 4.2.1 or Section 4.2.2, which may include an optional fragment identifier as specified in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) Section 3.5.
```

Old (RFC 7230): Allowed the fragment directly within the http-URI definition text string itself.
New (RFC 9110): Stripped the fragment from the http-URI definition. If a # fragment exists, it changes the string's classification from an absolute URI into a URI Reference.

stevenc-stb (changes requested):

stevenc-stb (approved):

goneall (approved):

Issue #1166, opened 2025-11-17
SPDX enumeration value URLs could use `.` or `_` not `/` as separator


VladimirAlexiev:
(A sub-issue of https://github.com/spdx/spdx-3-model/issues/1165)

URLs of enumeration values:
- In 2.3 used to be like this `<https://spdx.org/rdf/terms/IsoAutomationLevel_conditionalAutomation>`, which can be written SPARQL like `spdx:IsoAutomationLevel_conditionalAutomation`
- Currently they are like this `<https://spdx.org/rdf/3.1/terms/Core/IsoAutomationLevel/conditionalAutomation>`. Even after including per-module prefixes (eg `spdx-core:`) this would need to use a backslash in SPARQL: `spdx_core:IsoAutomationLevel\/conditionalAutomation`

Example:
- Electrical CIM enumeration values use `.` as separator, eg
```ttl
cim:PhaseCode.N a cim:PhaseCode
cim:UnitSymbol.W a cim:UnitSymbol
eu:LimitKind.tatl a eu:LimitKind 
nc:GateLogicKind.and a nc:GateLogicKind
```

PR #1425, opened 2026-07-30
Add SPARQL validation

JPEWdev:
An example SPARQL query in the model. See: https://github.com/spdx/spec-parser/pull/213

### DIFF
--- a/model/Core/Vocabularies/RelationshipType.md
+++ b/model/Core/Vocabularies/RelationshipType.md
@@ -103,3 +103,674 @@ name completes the sentence:
 - usesTool: The `from` Element uses each `to` Element as a tool, during a LifecycleScopeType period.
 - validatedOn: The `from` Element has been validated on the `to` Element(s).
 - verifiedBy: The `from` Requirement that has verification (test, review, analysis etc.) details defined in the `to` /FunctionalSafety/RequirementVerification.
+
+## SPARQL
+
+- affects_from
+  - message: When relationship is affects, from must be a Vulnerability, Action, or DefinedProcess
+  - query: <<<
+
+        SELECT $this WHERE {
+            $this <Core/relationshipType> <Core/RelationshipType/affects> .
+            $this <Core/from> ?from .
+            FILTER NOT EXISTS { ?from rdf:type/rdfs:subClassOf* <Security/Vulnerability> }
+            FILTER NOT EXISTS { ?from rdf:type/rdfs:subClassOf* <Core/Action> }
+            FILTER NOT EXISTS { ?from rdf:type/rdfs:subClassOf* <Core/DefinedProcess> }
+        }
+
+- assumes_to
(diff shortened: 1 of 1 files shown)


bact:
@JPEWdev may be related issue
- https://github.com/spdx/spdx-3-model/issues/1166

JPEWdev:
Generating these SPARQL rules from the text descriptions seemed like a really good fit for AI generation, and it did a really good job.

JPEWdev:
> @JPEWdev may be related issue
> 
>     * [SPDX enumeration value URLs could use `.` or `_` not `/` as separator #1166](https://github.com/spdx/spdx-3-model/issues/1166)

Yes, I did notice that. It does work with correctly with BASE though, which is why I chose to use that instead. I think that should be good enough for the SPARQL queries

bact:
This will work as an example or for a validation?

JPEWdev:
It turned out to be pretty easy to get the AI to spit out the SPARQL queries, so I would probably say this isn't an example anymore. I've been testing the rules by generating a model with https://github.com/spdx/spec-parser/pull/213 applied to the spec parser and it does appear to work as intended

bact:
Looks great. Will be really really useful if we can have this integrated into the published RDF/Turtule file:

```turtle
sh:sparql [
    sh:message "..." ;
    sh:select """
        SELECT ...
    """ ;
]
```

which https://github.com/spdx/spec-parser/pull/213 is exactly doing.

In terms of human editor creation and review, this requires knowledge of SPARQL.
As long as this field is optional, it should be fine (as the author who add constraints in prose doesn't have to immediately have the SPARQL; the SPARQL can be added later in time, by the same or the different author).

(I have a different proposal, based on [a YAML-based rule](https://github.com/bact/specmd/blob/main/docs/format.md#structured-entry-format) - which may be more author-friendly - but it is still another domain-specific language to learn - and not a standardized one like the SPARQL used here)

bact (approved):

model/AI/Classes/AIPackage.md
```diff
@@ -80,3 +80,13 @@ artificial intelligence capabilities.
...
+
+## SPARQL
+
+- primary_purpose
+  - message: A value for /Software/SoftwareArtifact/primaryPurpose is required for an AIPackage
+  - query:
+
+        SELECT $this WHERE {
+            FILTER NOT EXISTS { $this <Software/primaryPurpose> ?primaryPurpose }
+        }
```
zvr:
```suggestion

- primary_purpose
  - message: A value for /Software/SoftwareArtifact/primaryPurpose is required for an AIPackage
  - query: <<<
        SELECT $this WHERE {
            FILTER NOT EXISTS { $this <Software/primaryPurpose> ?primaryPurpose }
        }
```

bact (approved):

zvr (approved):

PR #1427, opened 2026-08-04
Add version field to BOM

### LINKED ISSUE #1288: Add SBOM version
Among other elements, the ["Software Bill of Materials (SBOM) for Artificial Intelligence - Minimum Elements"](https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/KI/SBOM-for-AI_minimum-elements.html ) from G7 Cybersecurity Working Group listed **"SBOM version"** as one of the minimum elements.

Current SPDX 3 model does not have any field to record the SBOM version.
There is `specVersion` in `CreationInfo`, but that is for SBOM spec version (SPDX 3.0, SPDX 3.1), not the version of SBOM document.

The G7 SBOM for AI's description of "SBOM version":

> Identifier designated by the SBOM author to specify a change  in the SBOM document from a previously identified version or to indicate that it is the first version.
>
> The SBOM version element indicates a relationship with earlier iterations of an SBOM and signals that the SBOM  author made changes to the previous iteration. The SBOM version tracks the SBOM component data for a componentname/component-version pair. The SBOM version element may use Semantic Versioning; if it uses Semantic Versioning, the major version of a published SBOM following these minimum elements should be 1. If using an identifier such as a serial number to differentiate between versions of an SBOM, use of the identifier should conform to relevant standards (for example, RFC 9562 for serial numbers). SBOM authors should use minor version and patch version to indicate changes to the content of the SBOM elements as appropriate. SBOM authors should update the version for an SBOM for a given component name-version pair when editing data about the target component.

ENISA [**SBOM Landscape Analysis: Towards an implementation guide**](https://www.enisa.europa.eu/sites/default/files/2025-12/SBOM%20Analysis%20-%20Towards%20an%20Implementation%20Guide_v1.20-Published.pdf) (Public Draft 1.2 December 2025) also discussed about SBOM versioning and retention requirements (page 27, last column in "Implementation Risk Management" row) and put "(SBOM) version history retention" as part of Table 9 - Minimum Requirements for SBOM management; page 30, look at first row of Table 9 "Storage & Retrieval").

ENISA requirements also discussed during [SPDX Tech Team Meeting 2026-06-22 ASIA](https://docs.google.com/document/d/1NdHYU_VZtLacD4bEmf2GiUVRTbrcev1beaJpq8s8-pU/edit?tab=t.ewp5o8u06tfu)

- Should we have the SBOM version information in SPDX 3 SBOM?
- How should it be named ?
- Where should this information be hosted/attached to? (more like to `Bom` or than to `SpdxDocument`?)

goneall:
Fixes #1288

### DIFF
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -17,8 +17,9 @@ provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
 The `version` property is an optional human readable hint for different versions of a BOM
-representing the same basic `Bundle` of elements provided by the same BOM
-`creationInfo` `createdBy` `Agent` with a BOM of the same `name`.
+representing the same basic `Bundle` of elements
+provided by the same creator (same `Agent` in BOM's `creationInfo.createdBy`)
+with a BOM of the same `name`.
 To accurately represent changes or updates to a BOM, a `Relationship`
 should be created from the updated BOM to the original BOM with the `amendedBy`
 relationship type.
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -20,9 +20,11 @@ The `version` property is an optional human readable hint for different versions
 representing the same basic `Bundle` of elements
 provided by the same creator (same `Agent` in BOM's `creationInfo.createdBy`)
 with a BOM of the same `name`.
+
 To accurately represent changes or updates to a BOM, a `Relationship`
 should be created from the updated BOM to the original BOM with the `amendedBy`
 relationship type.
+
 Since the `spdxId` must be unique for each version of a given BOM, it can be considered
 a unique version string if the version field is not used.
 
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -22,7 +22,7 @@ representing the same basic `Bundle` of elements provided by the same BOM
 To accurately represent changes or updates to a BOM, a `Relationship`
 should be created from the updated BOM to the original BOM with the `amendedBy`
 relationship type.
-Since the `spdxId` must be unique for each version of a given SBOM, it can be considered
+Since the `spdxId` must be unique for each version of a given BOM, it can be considered
 a unique version string if the version field is not used.
 
 ## Metadata
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -16,8 +16,24 @@ This could include details of the content and composition of the product,
 provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
+The `version` property is an optional human readable hint for different versions of a BOM
+representing the same basic `Bundle` of elements provided by the same BOM
+`creationInfo` `createdBy` `Agent` with a BOM of the same `name`.
+To accurately represent changes or updates to a BOM, a `Relationship`
+should be created from the updated BOM to the original BOM with the `amendedBy`
+relationship type.
+Since the `spdxId` must be unique for each version of a given SBOM, it can be considered
+a unique version string if the version field is not used.
+
 ## Metadata
 
 - name: Bom
 - SubclassOf: Bundle
 - Instantiability: Concrete
+
+## Properties
+
+- /Core/version
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -38,7 +38,7 @@ a unique version string if the version field is not used.
   - minCount: 0
   - maxCount: 1
 
-## SPARQL 
+## SPARQL
 
 - name_when_version
   - message: A value for Core/name is required when a value for Core/version is present
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -17,7 +17,7 @@ provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
 The `version` property is an optional human readable hint for different versions of a BOM
-representing the same basic `Bundle` of elements
+representing the same overall system
 provided by the same creator (same `Agent` in BOM's `creationInfo.createdBy`)
 with a BOM of the same `name`.
 
--- a/model/Core/Classes/Bom.md
+++ b/model/Core/Classes/Bom.md
@@ -37,3 +37,15 @@ a unique version string if the version field is not used.
   - type: xsd:string
   - minCount: 0
   - maxCount: 1
+
+## SPARQL 
+
+- name_when_version
+  - message: A value for Core/name is required when a value for Core/version is present
+  - query: <<<
+        SELECT $this WHERE {
+            $this <Software/version> ?version .
+            FILTER NOT EXISTS {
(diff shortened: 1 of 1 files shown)


model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,24 @@ This could include details of the content and composition of the product,
 provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
+The `version` property is an optional human readable hint for different versions of a BOM
+representing the same basic `Bundle` of elements provided by the same BOM
+`creationInfo` `createdBy` `Agent` with a BOM of the same `name`.
```
bact, line 21: This is correct but we may need some convention on notation to express this consistently.

I see some people use `/Core/Bom.creationInfo.createdBy` (`Agent` is a range, we may not need to put it there; same as we will not say `version` `xsd:string`).

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,24 @@ This could include details of the content and composition of the product,
...
+To accurately represent changes or updates to a BOM, a `Relationship`
+should be created from the updated BOM to the original BOM with the `amendedBy`
+relationship type.
+Since the `spdxId` must be unique for each version of a given SBOM, it can be considered
```
bact: Need to be a BOM here ? Not sure

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,24 @@ This could include details of the content and composition of the product,
...
 - name: Bom
 - SubclassOf: Bundle
 - Instantiability: Concrete
+
+## Properties
+
+- /Core/version
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
```
zvr, line 39:
```suggestion
  - maxCount: 1

## SPARQL 

- name_when_version
  - message: A value for Core/name is required when a value for Core/version is present
  - query: <<<
        SELECT $this WHERE {
            $this <Software/version> ?version .
            FILTER NOT EXISTS {
                $this <Software/name> ?name .
            }
        }
```

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,24 @@ This could include details of the content and composition of the product,
 provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
+The `version` property is an optional human readable hint for different versions of a BOM
+representing the same basic `Bundle` of elements provided by the same BOM
```
zvr, line 20: "same `Bundle` of elements" ?
But on v2 I want to fix my omissions and add new elements that were mistakenly forgotten...

zvr (changes requested):

model/Core/Classes/Bom.md
bact: Similar notation issues in the SBOM minimum elements table that currently under review.

model/Core/Classes/Bom.md
goneall: Hence the term "basic" in the sentence.  We could use the word "similar" or be more specific and state that the root of the SBOM should be the same or an updated version of the same Element.

What I was trying to get across is that the version should represent a version of the same SBOM content.

Perhaps we should state just that - ... "for different version of a BOM representing the same content for the "...

If it is confusing, we could just remove it since I think it should be pretty obvious the version applies to the same "basic"content.

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,36 @@ This could include details of the content and composition of the product,
...
 - Instantiability: Concrete
+
+## Properties
+
+- /Core/version
+  - type: xsd:string
+  - minCount: 0
+  - maxCount: 1
+
+## SPARQL
```
goneall:
```suggestion
## SPARQL
```

model/Core/Classes/Bom.md
zvr: Grrr... I wrote the correct `Core/name` and `Core/version` in the message, but somehow put `Software/` in the query...

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,36 @@ This could include details of the content and composition of the product,
 provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
+The `version` property is an optional human readable hint for different versions of a BOM
+representing the same basic `Bundle` of elements provided by the same BOM
+`creationInfo` `createdBy` `Agent` with a BOM of the same `name`.
+To accurately represent changes or updates to a BOM, a `Relationship`
+should be created from the updated BOM to the original BOM with the `amendedBy`
+relationship type.
+Since the `spdxId` must be unique for each version of a given SBOM, it can be considered
```
bact:
```suggestion
Since the `spdxId` must be unique for each version of a given BOM, it can be considered
```

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,36 @@ This could include details of the content and composition of the product,
 provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
+The `version` property is an optional human readable hint for different versions of a BOM
+representing the same basic `Bundle` of elements provided by the same BOM
+`creationInfo` `createdBy` `Agent` with a BOM of the same `name`.
```
bact, line 21:
```suggestion
representing the same basic `Bundle` of elements
provided by the same creator (same `Agent` in BOM's `creationInfo.createdBy`)
with a BOM of the same `name`.
```

Will this help readability?

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,36 @@ This could include details of the content and composition of the product,
...
+To accurately represent changes or updates to a BOM, a `Relationship`
+should be created from the updated BOM to the original BOM with the `amendedBy`
+relationship type.
+Since the `spdxId` must be unique for each version of a given BOM, it can be considered
+a unique version string if the version field is not used.
```
bact, line 26:
```suggestion

To accurately represent changes or updates to a BOM, a `Relationship`
should be created from the updated BOM to the original BOM with the `amendedBy`
relationship type.

Since the `spdxId` must be unique for each version of a given BOM, it can be considered
a unique version string if the version field is not used.
```

Try to make the paragraphs smaller and each one has one core idea.

goneall:
@bact @zvr - If you could re-review - if it's OK, we can merge

bact (approved):

kestewart (approved):
I still think version field is going to cause confusion on how it's used, so appreciate that "ammendedby" relationship is called out as the way to signal change,  as is the spdxID is the version.

model/Core/Classes/Bom.md
```diff
@@ -16,8 +16,39 @@ This could include details of the content and composition of the product,
 provenance details of the product and/or
 its composition, licensing information, known quality or security issues, etc.
 
+The `version` property is an optional human readable hint for different versions of a BOM
+representing the same basic `Bundle` of elements
```
kestewart:
```suggestion
representing the same overall system
```

kestewart (commented):
Discussion on SPDX Tech call,  removing specificity from bundle to system.   Per discussion with ZVR.

zvr (approved):

PR #1429, opened 2026-08-06
Update CountryCodeAlpha3.md for clarity and accuracy

### LINKED ISSUE #1369: [3.1-RC1] Note ISO 3166-1 alpha-3 validity for CountryCodeAlpha3
**Affected Sections**: model/Core/Datatypes/CountryCodeAlpha3.md<br><br>**Feedback**<br>The pattern accepts any three uppercase letters.<br><br>**Suggested Change:**<br>Add a non-normative note that conforming codes must be in the current ISO 3166-1 alpha-3 list and recommend supplementing SHACL with a code-list lookup. (model/Core/Datatypes/CountryCodeAlpha3.md)<br><br>**Notes:**<br>This feedback was provided by Microsoft at the SPDX working group meeting on June 30, 2026.

goneall:
Clarified the definition and validation requirements for the ISO 3166-1 alpha-3 country code.

Fixes #1369 

Includes the text proposed in [this comment.](https://github.com/spdx/spdx-3-model/issues/1369#issuecomment-5198586149)

### DIFF
--- a/model/Core/Datatypes/CountryCodeAlpha3.md
+++ b/model/Core/Datatypes/CountryCodeAlpha3.md
@@ -8,7 +8,7 @@ A string constrained to the ISO 3166-1 alpha-3 three-letter uppercase format.
 
 ## Description
 
-The value shall be an uppercase three-letter country code assigned by [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) at the time the SPDX element is created.
+The value shall be an uppercase three-letter country code assigned by [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) at the time the SPDX element containing this data type is created.
 
 The pattern constraint validates only the lexical form; it does not establish that the value is an assigned ISO 3166-1 alpha-3 code. Implementations should validate values against the applicable ISO code list.
 
--- a/model/Core/Datatypes/CountryCodeAlpha3.md
+++ b/model/Core/Datatypes/CountryCodeAlpha3.md
@@ -4,7 +4,7 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-A string containing an assigned ISO 3166-1 alpha-3 country code.
+A string constrained to the ISO 3166-1 alpha-3 three-letter uppercase format.
 
 ## Description
 
--- a/model/Core/Datatypes/CountryCodeAlpha3.md
+++ b/model/Core/Datatypes/CountryCodeAlpha3.md
@@ -4,13 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-A string constrained to the ISO 3166-1 alpha-3 three-letter format.
+A string containing an assigned ISO 3166-1 alpha-3 country code.
 
 ## Description
 
-The string shall be in the [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) three-letter format.
+The value shall be an uppercase three-letter country code assigned by [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) at the time the SPDX element is created.
 
-See the [ISO 3166-1 alpha-3 Wikipedia page](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) for more information.
+The pattern constraint validates only the lexical form; it does not establish that the value is an assigned ISO 3166-1 alpha-3 code. Implementations should validate values against the applicable ISO code list.
 
 ## Metadata
 
--- a/model/Core/Properties/country.md
+++ b/model/Core/Properties/country.md
@@ -9,6 +9,7 @@ Specifies a country code of the location.
 ## Description
 
 Specifies a country code of the location.
+The value shall be an uppercase three-letter country code assigned by [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) at the time the SPDX element containing this property is created.
 
 ## Metadata
 
--- a/model/Operations/Properties/exportingCountry.md
+++ b/model/Operations/Properties/exportingCountry.md
@@ -9,6 +9,7 @@ Country for which export controls must be taken into account.
 ## Description
 
 Country for which export controls must be taken into account.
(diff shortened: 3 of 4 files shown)


bact (approved):

model/Core/Datatypes/CountryCodeAlpha3.md
```diff
@@ -4,13 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
 
 ## Summary
 
-A string constrained to the ISO 3166-1 alpha-3 three-letter format.
+A string containing an assigned ISO 3166-1 alpha-3 country code.
```
stevenc-stb: “A string containing an assigned ISO 3166-1 alpha-3 country code.”
A datatype does not “contain” anything; it constrains the form of a string. Using “containing” suggests the value is a container of a couuntrycode, which is a property . 

See LanguageTag -> A string constrained to the IETF BCP 47 language tag format,
See MediaType -> is a string constrained to the [RFC 2046 MIME Part Two: Media Types](https://datatracker.ietf.org/doc/rfc2046/).

model/Core/Datatypes/CountryCodeAlpha3.md
```diff
@@ -4,13 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
 ## Description
 
-The string shall be in the [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) three-letter format.
+The value shall be an uppercase three-letter country code assigned by [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) at the time the SPDX element is created.
```
stevenc-stb: is this mixing the datatype’s lexical constraint with a temporal semantic requirement?

stevenc-stb (changes requested):
This read like a mix of a datatype and an Property

model/Core/Datatypes/CountryCodeAlpha3.md
stevenc-stb:
```suggestion
The value shall be an uppercase three-letter country code assigned by [ISO 3166-1 alpha-3](https://www.iso.org/obp/ui/#iso:std:iso:3166:-1) 
```

model/Core/Datatypes/CountryCodeAlpha3.md
stevenc-stb:
```suggestion
A string constrained to the ISO 3166-1 alpha-3 three-letter uppercase format.
```

model/Core/Datatypes/CountryCodeAlpha3.md
```diff
@@ -4,13 +4,13 @@ SPDX-License-Identifier: Community-Spec-1.0
...
 
-See the [ISO 3166-1 alpha-3 Wikipedia page](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) for more information.
+The pattern constraint validates only the lexical form; it does not establish that the value is an assigned ISO 3166-1 alpha-3 code. Implementations should validate values against the applicable ISO code list.
```
stevenc-stb, line 13:
```suggestion
Note: The pattern constraint validates only the lexical form; it does not establish that the value is an assigned ISO 3166-1 alpha-3 code. Implementations should validate the value against the applicable ISO 3166-1 list (e.g., the version current at the time the SPDX element is created) when semantic correctness is required.
```
That might better as a note.

stevenc-stb (changes requested):
This might be a better wording. @bact? @goneall

model/Core/Datatypes/CountryCodeAlpha3.md
goneall: I do think this is important in that the ISO alpha-3 country codes do change.  We solve this in the SPDX listed license reference by including a license list version as a field.  The ISO 3166 standard, on the other hand, uses a year designation - so a creation date would be more appropriate.

stevenc-stb: i thought in OWL datatypes are static, timeless constraints. 
From https://www.w3.org/TR/owl2-syntax/#Datatypes
a datatype has three properties:
    A [·value space·](https://www.w3.org/TR/xmlschema11-2/#dt-value-space), which is a set of values.
    A [·lexical space·](https://www.w3.org/TR/xmlschema11-2/#dt-lexical-space), which is a set of [·literals·](https://www.w3.org/TR/xmlschema11-2/#dt-literal) used to denote the values.
    A small collection of functions, relations, and procedures associated with the datatype.  Included are equality and (for some datatypes) order relations on the [·value space·](https://www.w3.org/TR/xmlschema11-2/#dt-value-space), and a [·lexical mapping·](https://www.w3.org/TR/xmlschema11-2/#dt-lexical-mapping), which is a mapping from the [·lexical space·](https://www.w3.org/TR/xmlschema11-2/#dt-lexical-space) into the [·value space·](https://www.w3.org/TR/xmlschema11-2/#dt-value-space).

None of these properties references when an element is created. They are sets and functions defined once, independent of any instance.

It introduces ambiguity. What does "created" mean? When the SPDX document is serialized? When the element is added to a graph?  This is not a clearly defined boundary, and it makes the specification harder to implement consistently. Not all uses of [/Core/CountryCodeAlpha3](https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Datatypes/CountryCodeAlpha3/) are subclass of Element see https://spdx.github.io/spdx-spec/v3.1-dev/model/Operations/Classes/ExportControlClassification/ with the [exportingCountry](https://spdx.github.io/spdx-spec/v3.1-dev/model/Operations/Properties/exportingCountry/) 

while in the ExportControlClassification would use the ExportControlClassificationAssessment [creationInfo](https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Properties/creationInfo/) [created](https://spdx.github.io/spdx-spec/v3.1-dev/model/Core/Properties/created/)

The Time constant should be put in the Properties or as a Note that is nonNomitive.

goneall:
Discussed on the 18 Aug 2026 tech call:
- The "datatype" is not an actual `owl:datatype` and doesn't carry the same level of restrictions
- The "datatype" is not an object, it is just information about the property so it doesn't carry the same level of metadata
- The creationDate used would the be creation date on the element that contains this property
- Need to update the description to reference the creation date for the element that contains this property - Gary to update

goneall:
@stevenc-stb - following up on the 18 Aug 2026 tech call, I updated the data type description and copied the same description to all of the properties that use the data type.

I didn't think we needed to description at the class level.

Please review.

stevenc-stb (approved):

PR #1435, opened 2026-08-14
serialization: Add documentation about json signing and attestation

JPEWdev:


bact:
@JPEWdev @goneall @zvr should we have these information here or in the spdx-spec repo ?

goneall:
> @JPEWdev @goneall @zvr should we have these information here or in the spdx-spec repo ?

If I recall a discussion a few months back, I think we wanted most of this information in the spdx-spec repo.

Reference https://github.com/spdx/spdx-spec/pull/1381

zvr:
I think information explaining signing and attestation, like in this PR, belongs here as submitted (or even in the `using` repo).

In the spec, in the Serialization chapter, we should only mention a couple of sentences about support of In-Toto and JSF, adding them to the list of references. No examples or code needed there.

serialization/jsonld/signing.md
```diff
@@ -0,0 +1,33 @@
...
+
+SPDX does not currently allow sub-objects of the document to be signed (e.g.
+individual objects in the `@graph` list). Only the complete document can be
+signed.
+
+## Example
+
+```jsonc
+{
+    "@context": "https://spdx.org/rdf/3.0.1/spdx-context.jsonld",
```
bact:
```suggestion
    "@context": "https://spdx.org/rdf/3.0/spdx-context.jsonld",
```

The version part of the URL is inconsistent with another example in attestation.md

serialization/jsonld/signing.md
```diff
@@ -0,0 +1,30 @@
+# Signing SPDX 3 JSON Documents
+
+SPDX JSON Documents may be signed using [ITU-T X.590 JSON Signature Schema (JSS)][JSS].
+The signature must be placed in a `signatures` property at the top level of the
+document. For the structure of the signature, see the [JSS][JSS] specification.
+
+SPDX does not currently allow sub-objects of the document to be signed (e.g.
```
goneall, line 7:
```suggestion
For SPDX, this method of signing does not allow sub-objects of the document to be signed (e.g.
```

Suggested on the 18 Aug 2026 tech call

goneall (changes requested):
One change requested from review on tech call

goneall:
Discussed on the tech call:
- We shouldn't have this in the model repo - hard to find
- Split into 2 PR's
  - The spec repo describes signing in general
  - The actual examples and specifics on how to implement will be in the Using repo
  - To implement - this PR as is could be moved to the Using repo, and we can extract a few sentences to copy to the spec repo
  - In the spec repo - we should mention that the JSS signature needs to be at the top level - consistent with the schema

serialization/jsonld/signing.md
JPEWdev: I think there is a confusion here. There's not a technical limitation imposed by JSS as to why you couldn't sign a subelement of the document; I just don't think it's useful and I don't want to say it's allowed until we think about it. In that context, the proposed change makes it seem like a limitation of JSS

JPEWdev:
See https://github.com/spdx/spdx-spec/pull/1468

goneall (approved):
LGTM

goneall:
Discussed on the tech call - 15 Sept 2026 - still some disagreement on the exact wording on the subelement signing, but we agreed to go ahead and merge.  We can open a separate PR later after discussing with @JPEWdev

PR #1456, opened 2026-09-16
Deprecate standardCompliance; use conformsTo instead

bact:
From 16 Sep 2026 AI team call,

- Deprecate `/AI/standardCompliance` property in favor of "conformsTo" relationship type.
  - Relationship allows us to point to a `Specification` (standard) (which will partially address "standard name" and "standard URL" as raised in #387)
- Fix ISO verb (may -> can)


stevenc-stb (approved):

goneall (approved):
Reviewed and approved on the 22 Sept 2026 tech call

